PR #303, opened 2026-06-04
fix(tests): repair flaky wiki-sync release-gate smoke harness

stevensacks:
## Summary

Triage and fix of the flaky `wiki-sync` release-gate smoke scenarios (`.gaia/tests/smoke/wiki-sync/01..05`). They failed on clean `main` (pre-existing, not a regression) with a pass/fail split that varied run-to-run.

**Root cause is a harness provisioning defect, not (primarily) LLM variance.** The sync playbook (`.claude/skills/gaia/references/wiki/sync.md`, Steps 1-9) shells out to `.gaia/cli/gaia wiki ...` for every state read, commit classification, log write, state bump, and land. **The fixtures never copied that binary into the synthetic temp repo** (they only `cp` hooks + skill markdown). With no CLI present the sync subagent had no deterministic oracle and bailed at Step 1; when an earlier improvising run hand-wrote artifacts instead of bailing, the loose assertions passed or failed by chance, producing the observed run-to-run flips.

Confirmed empirically by removing the `> /dev/null 2>&1` blackhole and reading the real sessions:

> **Unable to execute playbook.** The `.gaia/cli/gaia` CLI tool is not available in this environment ... Without this binary, the playbook cannot proceed past Step 1.

After provisioning the CLI, **all five scenarios pass live (01-05 green).**

## Root cause per scenario

The deterministic classifier output (`gaia wiki commit-classify`) was verified directly against each fixture's commit history:

| # | What it asserts | Deterministic layer (CLI) | Why it flaked before |
|---|---|---|---|
| 01 | `feat:` w/ `Invariant:` body → `wiki/services/Gemini.md` + WORTHY | CLI: **WORTHY** (`app/** non-test`) | CLI absent → subagent bailed → no page/log |
| 02 | typo `fix:` → SKIP, no page | CLI: **WORTHY** (`fix:` touching only README falls to the "defer to human review" default) | CLI absent → bailed; *and* even with the CLI this needs the LLM to demote WORTHY→SKIP per Step 4 (pure judgment) |
| 03 | 6 commits → >=5 log entries, >=1 page, `Serena handles inventory` marker | CLI: 2x SKIP "...(Serena handles inventory)", 1 WORTHY fix, docs/deps SKIP | CLI absent → bailed; marker only appears if the LLM logs the CLI `suggestion_reason` verbatim |
| 04 | shell commit → first prompt surfaces drift; sync catches up | CLI: Auth **WORTHY** | first assertion greps Claude's *prose* for "drift" (hook-injected, LLM-echoed); sync half needs the CLI |
| 05 | vanilla service add → SKIP, no page, `Serena handles inventory` | CLI: **SKIP** "...(Serena handles inventory)" | CLI absent → bailed; marker depends on verbatim reason logging |

**Conclusion:** there is a real, deterministic harness defect (missing CLI) that guaranteed failure, fixed here. But even with the CLI provisioned, every scenario's assertions ride on free-form output of nested non-deterministic sessions (outer Sonnet `claude -p` + the Haiku sync subagent): which page got created, the exact reason string logged, whether a prose answer says "drift", and the WORTHY→SKIP demotion in 02. That layer cannot be made deterministic without rewriting the assertions to call the CLI directly, which would stop testing the LLM playbook (the whole point of these E2E smokes). So the right move is **both**: fix the real defect *and* de-gate the irreducibly-nondeterministic lane.

## Fixes (all under `.gaia/tests/smoke/`; no CLI src or workflow touched)

1. **Provision the bundled `gaia` CLI** into each fixture (`cp "$GAIA_REPO/.gaia/cli/gaia" .gaia/cli/gaia`). The binary is a self-contained esbuild bundle and runs standalone. This is the real fix; it turns a 100%-fail harness green.
2. **Relative-path `cd`-before-resolve fix.** `GAIA_REPO` is now resolved from the script's own location (`git -C "$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)" ...`) **before** `cd "$TMP"`. The old order resolved a relative `BASH_SOURCE` against the temp dir and died with `fatal: cannot change to '.gaia/tests/smoke/wiki-sync'` before any assertion ran, masking the real failure. Verified: a relative `bash NN.sh` invocation now works.
3. **Stop the blackhole.** Each scenario captures its session to `$TMP/claude-sync.log` and dumps it via the `EXIT` trap on any non-zero exit, so failures are diagnosable.
4. **De-gate to an advisory lane.** New `wiki-sync/run.sh` runs the scenarios with bounded retries (`WIKI_SYNC_MAX_ATTEMPTS`, default 2) and surfaces captured output on a final failure. `run-all.sh` now invokes it as a non-gating **ADVISORY** lane; the deterministic structural smokes (`wiki-promote`, `uat-write`) stay blocking. The deterministic CLI core (`commit-classify`, `log-prepend`, `state`, `sync land`) is already unit-tested under `.gaia/cli/src/wiki/` and gated by `cli-tests.yml`.

## Verification

- Live: all 5 scenarios pass on the fixed harness (01, 02, 03, 04, 05).
- Relative-path fix proven with a throwaway probe (NEW form resolves the repo root; OLD form dies with the exact masking `fatal` error). Probe removed.
- `run.sh` retry/capture/summary plumbing validated deterministically with pass / flaky / always-fail stubs (no billable sessions).
- `bash -n` clean on all touched scripts; em-dashes confined to the two sanctioned fixture lines (`03`, `04`); observability/serena note in `run-all.sh` preserved verbatim.

## Notes for review

- Left open for review per request; not merged.
- No `.gaia/cli/src` or `.github/workflows/*` files touched, so the bundle rebuild and tmpl-byte-identity constraints do not apply; the audit-marker requirement's in-scope trigger does not fire.
- Quality Gate: no staged `.ts/tsx/js/jsx/mjs/cjs/css` or gate-affecting config, so the gate has nothing to check (pre-commit confirmed "No changed files -- skipping lint-staged").
- The `run-all.sh` observability note still carries a pre-existing `SPEC-005` narrative reference; preserved verbatim rather than scrubbed (out of scope for this fix).

### DIFF
--- a/.gaia/tests/smoke/README.md
+++ b/.gaia/tests/smoke/README.md
@@ -14,17 +14,18 @@ Artifacts that should NOT be here are markdown walk-through runbooks tied to a s
 
 ## Running
 
-### Wiki-sync E2E (billable)
-
 ```bash
 bash .gaia/tests/smoke/run-all.sh
 ```
 
-Walks every `wiki-sync/*.sh` scenario, prints PASS/FAIL, exits non-zero on any failure.
+Two lanes:
+
+- **Blocking** (sets the exit code): the deterministic structural harnesses `wiki-promote/run.sh` and `uat-write/run.sh`.
+- **Advisory** (never gates): `wiki-sync/run.sh` (billable). It drives real `claude -p` sessions, so its assertions depend on free-form LLM output and cannot block a release on a coin-flip. It retries each scenario to absorb one-off variance and surfaces the captured session output on a final failure. Run it standalone with `bash .gaia/tests/smoke/wiki-sync/run.sh`. See `wiki-sync/README.md` for the rationale.
 
 ## When to run
 
-- **Before cutting a GAIA release**; wiki-sync E2E. Verifies the wiki-sync system works under real Claude judgment, including the post-Serena WORTHY narrowing.
+- **Before cutting a GAIA release**; wiki-sync E2E (advisory). Verifies the wiki-sync system works under real Claude judgment, including the post-Serena WORTHY narrowing. Read the advisory summary; an `ADVISORY` line is a signal to investigate, not a hard block.
 - **Before merging a PR that touches the SPEC-003 or SPEC-004 hook surface**; `uat-write/` and `wiki-promote/` structural smokes.
 
 ## See also
--- a/.gaia/tests/smoke/run-all.sh
+++ b/.gaia/tests/smoke/run-all.sh
@@ -1,9 +1,15 @@
 #!/usr/bin/env bash
-# Run all release-gate smoke scenarios, report pass/fail.
-# Walks .gaia/tests/smoke/wiki-sync/*.sh in lexicographic order, then
-# invokes wiki-promote/run.sh and uat-write/run.sh; both are structural
-# release-gate harnesses with PASS/FAIL semantics, so they belong in this
-# driver.
+# Run the release-gate smoke harnesses, report pass/fail.
+#
+# Blocking lane (gates the release; sets the exit code): wiki-promote/run.sh and
+# uat-write/run.sh. Both are deterministic structural harnesses with PASS/FAIL
+# semantics, so they belong in the gate.
+#
+# Advisory lane (does NOT gate): wiki-sync/run.sh drives real `claude -p`
+# sessions whose assertions ride on free-form LLM output, so they are inherently
+# non-deterministic and cannot block a release on a coin-flip. It runs here for
+# visibility and retries each scenario to absorb one-off variance, but its
+# result never sets this driver's exit code. See wiki-sync/README.md.
 #
 # The observability/serena/ subtree (relocated from .gaia/tests/smoke/serena/
 # per SPEC-005 §Resolutions Q8) is intentionally NOT run from here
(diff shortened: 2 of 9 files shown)


stevensacks:
## Full advisory lane run (live confirmation)

```
PASS  01-meaningful-change.sh    (attempt 1/2)
PASS  02-typo-only-skip.sh       (attempt 1/2)
PASS  03-multi-commit-catchup.sh (attempt 1/2)
PASS  04-non-claude-merge.sh     (attempt 2/2)
PASS  05-serena-inventory-skip.sh(attempt 1/2)
RUNNER_EXIT=0
```

Scenario **04 flaked on attempt 1 and passed on retry** -- the de-gating + retry + capture design caught a real, irreducible LLM nondeterminism in the wild. The captured session output (now surfaced instead of blackholed) showed the cause:

> `Wiki already in sync at a047be4.`

04 first sends "What's the status of this repo?"; the drift-check hook injects *"Run /gaia-wiki sync to evaluate"*, and on that attempt the model proactively ran the sync itself, advancing state before the test captured `pre_claude_head` and before the explicit sync. The explicit sync then correctly reported "already in sync," so the `state == pre_claude_head` assertion failed. Attempt 2 did not pre-sync and passed.

This is exactly the coin-flip that would have failed a **blocking** gate on unchanged code. As an advisory lane: the retry absorbed it, the capture made it diagnosable, and the gate exit code stayed 0. Confirms all three hardening moves end-to-end (CLI provisioned so every scenario reaches its real assertions; retries absorb variance; capture diagnoses failures).

PR #332, opened 2026-06-10
chore(deps): refresh 9 packages for 1.6.0 / pnpm 11.5 readiness

stevensacks:
## Migration Report

Routine dependency refresh for the 1.6.0 release. All bumps are minor/patch (Wave A); no major migrations. Targets the **pnpm 11.5** upgrade: every package picked is past the 7-day `minimumReleaseAge` window, so the install stays clean under the stricter quarantine.

### Updated packages
| Group | Package | From | To | Type |
| --- | --- | --- | --- | --- |
| react | react | 19.2.6 | 19.2.7 | patch |
| react | react-dom | 19.2.6 | 19.2.7 | patch |
| react | @types/react | 19.2.15 | 19.2.16 | patch |
| storybook | @storybook/addon-docs | 10.4.1 | 10.4.2 | patch |
| storybook | @storybook/addon-links | 10.4.1 | 10.4.2 | patch |
| storybook | @storybook/react-vite | 10.4.1 | 10.4.2 | patch |
| storybook | storybook | 10.4.1 | 10.4.2 | patch |
| axe-core | axe-core | 4.11.4 | 4.12.0 | minor |
| chromatic | chromatic | 17.1.0 | 17.2.0 | minor |

### Breaking changes applied
None. Wave A only; no major version bumps.

### Overrides audited
- **Retained** (`pnpm.overrides`): `brace-expansion@>=5.0.0 <5.0.6`, `qs`, `ws`. These are CVE/security version-floors, not peer-dependency fixes, so the standard "remove and check for peer-dep errors" obsolescence test does not apply (removing a security pin produces no peer-dep error). `qs` resolves to exactly its floor `6.15.2`, confirming it is load-bearing. Kept as forward supply-chain defense.
- **Added** (`pnpm-workspace.yaml` -> `trustPolicyExclude`): `semver@6.3.1`. See note.

### Skipped packages
None reverted mid-run.

### Quality gate
| Step | Result |
| --- | --- |
| typecheck | pass |
| lint | pass |
| test | pass (120/120, 26 files) |
| pw | pass (2/2; first run hit a cold-start hydration timeout, confirmed flaky, passes warm) |
| build | pass |

---

### Notes for the reviewer

**1. `trustPolicyExclude: semver@6.3.1` was required to unblock the install.**
pnpm's live trust DB flags `semver@6.3.1` as a provenance downgrade: the 6.x line ended at 6.3.1 (before npm provenance was standard) and carries no attestation, while the 7.x line does. `trustPolicy: no-downgrade` then fails on every re-resolving install. The version is legitimate and required: `eslint-plugin-import` (pulled transitively via the first-party `@gaia-react/lint`) pins `semver ^6.3.1`. The exclusion is scoped to that exact version; `no-downgrade` stays enforced for everything else. **This also unblocks the pnpm 11.5 upgrade**, which enforces the same check harder.

**2. ESLint held at 9.39.4.** `eslint@10.4.1` is available but the 9.x cap defers the major. ESLint 10 is a separate, deliberate migration, out of scope here.

**3. Cooldown working as intended.** Newer releases exist for several of these (`storybook@10.4.3`, `@types/react@19.2.17`, `chromatic@17.4.0`) but are still inside the 7-day `minimumReleaseAge` window, so pnpm correctly held them back. They will age in and bump on a future run.


stevensacks:
Added `@gaia-react/lint` 1.5.0 → 1.5.1 (now released from gaia-react/lint#26) so the lint config refresh ships with 1.6.0. This brings `eslint-plugin-storybook` 10.4.2 into gaia's tree via lint, aligned with the storybook 10.4.2 bump already in this PR.

Re-resolving lint's subtree surfaced one more latent trust-downgrade — `eslint-import-resolver-typescript@3.10.1` (final 3.x, required by `eslint-plugin-canonical ^3.7.0`; unattested while the 4.x line has provenance). Same benign pattern as `semver@6.3.1`, pre-existing via lint 1.5.0, added to `trustPolicyExclude`. The full set for 1.5.1 is exactly those two — verified by re-running the install to a clean pass.

Gate re-run green: lint, typecheck, test (120/120), build, e2e (2/2 warm).

PR #403, opened 2026-06-22
Refactor YearMonthDay/utils.ts — Remove import-time clock read + add unit tests

### LINKED ISSUE #399: Refactor YearMonthDay/utils.ts: remove import-time clock read + add direct unit tests for getSafeValue/getValues

PP-Namias:
Closes #399

## Summary

Two low-risk improvements to `app/components/Form/YearMonthDay/utils.ts`:

1. **Removed the module-level `new Date()`** — the helper module no longer has an import-time clock side effect.
2. **Added direct Vitest unit tests** for `getSafeValue` and `getValues`, covering leap-year, month-overflow, and day-clamping edge cases.

## What changed

| File | Action |
|------|--------|
| `app/components/Form/YearMonthDay/utils.ts` | Removed `TODAY`/`THIS_YEAR` constants; added exported `getCurrentYear()` function; `YEARS` and `DEFAULT_DATE` now use inline `new Date()` instead of a shared frozen timestamp |
| `app/components/Form/YearMonthDay/tests/utils.test.ts` | **New** — 11 unit tests for `getSafeValue` (8 cases) and `getValues` (3 cases) |

## Why

- An otherwise-pure helper module carried an import-time clock read, hurting test determinism.
- The bug-prone logic in `getSafeValue` (leap-year / month-overflow / day-clamping) had no direct test coverage — only indirect coverage through component renders.

## Behavior unchanged

- `YEARS`, `MONTHS`, `DEFAULT_DATE`, `DEFAULT_VALUE` — identical output
- Existing component tests in `tests/index.test.tsx` — untouched and still passing

## Test coverage added

**`getSafeValue`** (8 cases):
- Jan 31 → Feb in leap year (clamps to 29)
- Jan 31 → Feb in non-leap year (clamps to 28)
- Mar 31 → Apr (clamps to 30)
- Aug 31 → Sep (no clamp, September has 30 days)
- Feb 29 year change: leap → non-leap (clamps to 28)
- Feb 29 year change: leap → leap (preserves 29)
- Year change preserving day (Aug 31 → Aug 31)
- Month change within bounds (May 15 → Jun 15, no clamp)

**`getValues`** (3 cases):
- Valid ISO-8601 date → `[year, month, date]` split
- Invalid string → `DEFAULT_VALUE` split (`['2000', '01', '01']`)
- Empty string → `DEFAULT_VALUE` split

## Verification

- [x] All 11 new unit tests pass
- [x] Existing component tests unaffected
- [x] No new lint or type errors introduced


stevensacks:
Thanks for jumping on this. I'm closing it. Issue #399 is entangled with internal work that isn't visible from the issue: the file is deliberately being kept in its current shape as a fixture for tooling still in progress, so this refactor can't land as a standalone change. The issue was mislabeled "good first issue" in error and I've removed the label. Apologies for the wasted effort.

PR #852, opened 2026-07-19
fix(hooks): inspect every line of multi-line commands in block-manifest-write (#838)

### LINKED ISSUE #838: tech-debt: block-manifest-write.sh inspects only the first line of a multi-line command
## `block-manifest-write.sh` tokenizer inspects only the first line of a multi-line command

**Location:** `.claude/hooks/block-manifest-write.sh:77` (`read -r -a toks <<<"$cmd"`).

**Failure mode / bad outcome:** the guard tokenizes the command with `read -r -a toks <<<"$cmd"`, which reads only the **first line** of `$cmd`. A manifest write on line 2+ of a multi-line Bash command is therefore never inspected and passes the guard. The same first-line/whitespace tokenization also misses the no-space redirect shape `>.gaia/manifest.json` and command-substitution shapes `$(...)`. Example: a heredoc or `&&`-joined multi-line command whose second line does `printf ... > .gaia/manifest.json` slips through.

This is a **pre-existing best-effort gap, not a regression** — surfaced during the #683 audit. It is consistent with the hook's documented contract ("best-effort defense-in-depth, not airtight, biased toward allowing on ambiguity"), so it is a hardening candidate, not a correctness break. The airtight control remains the OS-level `sandbox.filesystem` deny; this hook is heuristic depth.

**Suggested fix:** iterate over every line of `$cmd` (not just the first `read`), and extend the redirect-target detection to the no-space `>path` and `$(...)` shapes, keeping the existing allow-on-ambiguity bias. Add fixtures covering the multi-line and no-space cases. If the maintainer decides the OS sandbox is sufficient and this depth layer should stay first-line-only, close as `wontfix` with that rationale.

Handler: prompt

stevensacks:
Closes #838

## What

The manifest-write guard hook (`.claude/hooks/block-manifest-write.sh`) tokenized only the **first line** of a command (`read -r -a toks <<<"$cmd"`), so a write to `.gaia/manifest.json` on line 2+ of a multi-line command was never inspected and slipped past the guard. It also missed the no-space redirect shapes (`>.gaia/manifest.json`, `>>.gaia/manifest.json`) where the operator and target land in a single token.

## Fix

- Tokenize **every** line of `$cmd`, appending a synthetic `;` at each line boundary so a newline resets segment state exactly like an explicit statement separator.
- Add a `'>'*` case branch (after the exact `'>' | '>>'` branch) to catch the glued redirect form, stripping the operator and running the same `is_guarded_path` check on the target.

The documented **allow-on-ambiguity bias** is preserved — the guard still blocks only on a confident match. Spaced `$(...)` targets were already covered by the existing glob suffix match; the only gap was the no-space/glued case, now covered.

## Tests

8 new fixtures in `.gaia/tests/hooks/block-manifest-write.bats`: multi-line write on line 2 (denied), a line-1 marker not exempting an unmarked line-2 write (denied), a marked line-2 write (allowed), no-space `>path` / `>>path` / `>$(...)path` (denied), and a manifest-adjacent no-space target (still allowed, no over-blocking).

## Verify

- Confirmed red first (TDD), then green.
- `shellcheck .claude/hooks/block-manifest-write.sh` → clean.
- `bats .gaia/tests/hooks/block-manifest-write.bats` under bash 5 → 44/44 passing.
- `.gaia/tests/` is release-excluded, so the new tests ship to no adopter.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.e9sHwenhIA

Issue #679, opened 2026-07-13
tech-debt: three parsers of .gaia/release-exclude disagree; a glob entry omits a file from the manifest while still shipping it


stevensacks:
## Three independent parsers of `.gaia/release-exclude` disagree on the same input

**Location:** `.gaia/cli/src/release/manifest.ts:112`

### Failure mode

`.gaia/release-exclude` is the distribution boundary: it alone decides which tracked files reach an adopter. It is parsed by **three independent implementations that do not agree**:

| Parser | Location | Treats `*` as |
|---|---|---|
| TypeScript regex compiler | `.gaia/cli/src/release/manifest.ts:112-137` (`escapeRegExp` + `parseExcludePatterns`) | a **glob**: `.replaceAll('*', '[^/]*')` |
| Shell staging pipeline | `.github/workflows/release.yml:94-96` (awk → sed → awk) | a **literal**: `sed 's|[][\\.*^$()+?{}|]|\\&|g'` escapes it to `\*` |
| Distribution suite | `.gaia/tests/distribution/01-files-present.sh:42-49` | a **literal**: bare `[ -e "$STAGING/$pat" ]` |

They also disagree on **leading whitespace**. `parseExcludeLines` (`manifest.ts:125-132`) calls `line.trim()`; the release-workflow awk prints `$0` verbatim, so an indented entry compiles to `^  wiki/meta(/|$)` and matches nothing in `git ls-files`.

**The divergence runs in the dangerous direction.** For a glob-shaped entry (or an indented entry), the TS compiler matches and **omits the file from `.gaia/manifest.json`**, while `rsync --files-from` in the staging step (`release.yml:88-106`) does **not** match and **stages the file into the tarball anyway**. A file the maintainer believes is withheld ships to every adopter, and because the manifest was generated with the same TS parser that excluded it, `release manifest --check` reports **clean**. Nothing in the pipeline notices.

Two further consequences of the same split:

- `escapeRegExp` escapes only `.`, `+`, `?` and rewrites `*`. It leaves `[`, `]`, `(`, `)`, `^`, `$`, `{`, `}`, `|` unescaped, so an entry containing `[` compiles into a character class, or throws a `SyntaxError` out of `new RegExp` and crashes `release manifest`. The shell escapes all of them literally.
[...]

PR #854, opened 2026-07-19
test(release): prove exclude-parser parity across CLI, staging, and distribution suite (#839)

### LINKED ISSUE #839: tech-debt: unify exclude-pattern parsers across CLI + staging pipeline (#679 remainder)
## #679 remainder: unify the exclude-pattern parsers across CLI + staging pipeline, add a complement-set regression test

**Location:** `.gaia/tests/distribution/01-files-present.sh:39` (the glob-shaped-entries handling), spanning the release CLI exclude parser (`.gaia/cli/src/release/manifest-answers.ts`, `escapeRegExp` / `parseExcludePatterns`) and the shell staging pipeline (`.github/workflows/release.yml`).

**Failure mode / bad outcome:** #679 made the CLI's exclude entries strictly literal (drop the `*`→`[^/]*` glob rewrite; `validateExcludeText` throws on any metacharacter). Three surfaces independently parse `.gaia/release-exclude`: the release CLI, the shell staging pipeline in `release.yml`, and this distribution bats suite. They **currently agree** (all treat entries as literal), so there is no active bug — but the agreement is by three separate implementations with no shared parser and no test proving they stay in sync. A future edit to one parser (e.g. re-introducing glob handling in the staging pipeline) would silently diverge from the CLI's now-enforced literal-only semantics, and nothing would catch it. #679 flagged the cross-parser unification + a complement-set regression test as its remainder, out of L1's `.gaia/cli/src/release/` scope because it touches `.github/` and `.gaia/tests/`.

**Suggested fix:** factor the exclude-pattern semantics into one shared source the three surfaces consume (or, minimally, add a complement-set regression test that asserts the CLI, the staging pipeline, and this suite classify the same `.gaia/release-exclude` fixture identically — ship vs withhold on every entry). Refresh the stale glob-shaped-entries comment in this suite to match the literal-only reality.

Handler: plan

stevensacks:
Closes #839

## What

The #679 remainder. Three surfaces independently parse `.gaia/release-exclude` and currently agree (all literal-only, post-#679), but nothing proved they stay in sync — a future edit to one parser could silently diverge with no test to catch it. This adds a complement-set regression proof and refreshes the one stale comment, taking the issue's **preferred minimal path** (not full cross-language unification).

## Changes

- **`.gaia/cli/src/release/exclude-parser-parity.test.ts`** (new CLI vitest, release-excluded):
  - Asserts the `awk | sed | awk` exclude-compile pipeline is **byte-identical** between `.github/workflows/release.yml` and `.gaia/tests/distribution/lib/build-staging.sh` — extracted live from both files at test time (not a hardcoded copy), with presence guards so it can't pass vacuously. This catches the exact #679 failure mode (one copy re-diverging).
  - Asserts the CLI compiler's excluded-path set (`parseExcludePatterns`) equals the real shell pipeline's `grep -vE -f` excluded set for a representative fixture (dir-prefix, dot/`+` escaping, near-miss decoys), pinned to the expected set of 4.
- **`.gaia/tests/distribution/09-exclude-parser-parity.sh`** (new, release-excluded): builds a fixture tracked tree + exclude file, runs the real 3-stage compile + staging, then the literal `[ -e ]` check (mirroring `01-files-present.sh`) and asserts agreement, with a positive control that deliberately reintroduces a leak and confirms the check catches it.
- **`.gaia/tests/distribution/01-files-present.sh`**: refreshed the stale comment that still described glob-shaped entries checked via `find … -path`; the code has been literal-only `[ -e ]` since #679.

## Not done (by design)

Full cross-language parser unification (one shared source across TS + shell + workflow YAML) is not clean — `release.yml` cannot depend on a `.gaia/tests/` script. Recommended as a separate follow-up rather than sprawled in here.

## Verify

- No bundled CLI source touched → binaries unchanged (confirmed via `git status`), no rebuild.
- CLI vitest full suite: 1507 passed / 2 pre-existing skips; `typecheck` + `lint --max-warnings=0` clean.
- `09-exclude-parser-parity.sh` under bash 5 → PASS; `shellcheck -S warning` clean; full `shell-lint.sh` gate clean.

All touched surfaces are maintainer-only / release-excluded → no CHANGELOG entry owed.

### DIFF
--- a/.gaia/cli/src/release/exclude-parser-parity.test.ts
+++ b/.gaia/cli/src/release/exclude-parser-parity.test.ts
@@ -60,6 +60,35 @@ const extractStageCommand = (text: string, marker: string): string => {
 };
 
 /**
+ * `grep -vE` exits 1 (not 0) when every candidate line matches the filter
+ * (i.e. zero survivors), which `execFileSync` treats as a thrown error even
+ * though it's a normal "no matches" outcome, not a failure. The fixture
+ * below always leaves a survivor so this can't fire today, but a future
+ * all-excluded fixture would otherwise surface an opaque "Command failed"
+ * instead of a clean empty `kept` set. Only a real error (exit >= 2, e.g. a
+ * malformed regex file) should propagate.
+ */
+const grepKeptLines = (regexPath: string, candidateText: string): string[] => {
+  try {
+    return execFileSync('grep', ['-vE', '-f', regexPath], {
+      encoding: 'utf8',
+      input: candidateText,
+    })
+      .split('\n')
+      .filter((line) => line.length > 0);
+  } catch (error) {
+    const status =
+      error instanceof Error && 'status' in error ?
+        (error as Error & {status?: number}).status
+      : undefined;
+
+    if (status === 1) return [];
+
+    throw error;
+  }
+};
+
+/**
  * Stage 1 (comment/blank strip) also carries a trailing file-path argument
  * that differs between the two callers (`.gaia/release-exclude` vs
  * `"$PROJECT_ROOT/.gaia/release-exclude"`); dropping it makes the awk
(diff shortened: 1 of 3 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.RewewUa51J

PR #882, opened 2026-07-19
fix(audit-ci): stamp GAIA-Audit status on the chore(deps) skip path

stevensacks:
## Problem

A `chore(deps):` / `chore(deps-dev):` PR skips the expensive frontend audit (the `/update-deps` wrapper runs the full quality gate locally before pushing). But the required `GAIA-Audit` check is a **POSTed commit status**, not a job conclusion, so unlike Chromatic / Vitest / bats it is *not* satisfied by a job-level skip. The chore-deps skip path posted only an advisory PR comment and never the status, so where `GAIA-Audit` is a required check a dep-bump PR passed every other check yet stayed unmergeable (`mergeStateStatus: BLOCKED`) because the required context was never posted. No chore(deps) PR had ever merged through the gate, so the gap sat latent since the skip was introduced.

## Fix

The chore-deps skip path now stamps `GAIA-Audit`, **member-aware, exactly as the out-of-scope skip step already does**:

- **success** when the bump touches only the frontend member's surface (the common case, and every adopter case, where `gate-pending-members.sh` reports no co-dispatched specialized member).
- **pending** (non-clobbering) when a specialized member CI cannot run is co-dispatched — matching the local merge hook, which clears the frontend member for chore(deps) via `check_chore_deps_pr` but still requires every specialized member's own marker.

The terminal comment now reports the actual stamp outcome (green / pending / not-stamped) instead of a bare "skipped" line that would read green on the blocked paths.

## Tests

- New regression guard (`chore-deps-audit-stamp.test.ts`) parses the template YAML and pins the stamp step to the skip path (`state=success` + `state=pending` + `gate-pending-members.sh`), so a future edit can't silently drop it and reopen the gap.
- Byte-identity drift guards (`audit-template-dogfood.test.ts` + artifact guard) pass: live == source template == build artifact. Full `.gaia/cli` suite green; CLI typecheck + lint clean.

Edited the live workflow and the source template together, regenerated the build artifact via `pnpm -C .gaia/cli bundle:adopter` (binary rebuilt byte-identical — template-only change). Portable: no maintainer-only path or member named in the shipped template.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.bAX6GTtGaQ

Issue #940, opened 2026-07-20
block-worktree-path-mismatch.sh: main_root derivation ignores payload cwd, so the guard goes inert when the hook's own process cwd leaves the repo


stevensacks:
`.claude/hooks/block-worktree-path-mismatch.sh:65`

## Failure mode

`main_root` derives from `git rev-parse --git-common-dir`, run with no `-C` flag against the hook's own process cwd:

```bash
common_dir="$(git rev-parse --git-common-dir 2>/dev/null)" || exit 0
```

When the hook's process cwd sits outside any git repository, this command fails and the hook exits 0 at line 65, before it reaches any of the payload-aware logic that resolves `current_root`. The guard is inert for that call: a target `file_path` resolving to the main checkout goes unguarded, even when the payload's `cwd` field names a perfectly valid checkout of this repo. Only `current_root` consults the payload (`jq -r '.cwd // empty'`); `main_root`'s own derivation never does.

This is pre-existing, fail-open behavior, consistent with the hook's stated posture ("any ambiguity … allows the call rather than blocking a legitimate edit"). It is worth tracking because it is the same silent-inertness class as the payload-cwd work already done for `current_root`: the guard has one remaining dependency on ambient process state that a payload value could answer instead.

The bats suite pins the current behavior directly: `a session whose cwd is not inside any git repository fails open (allowed)` (`.gaia/tests/hooks/block-worktree-path-mismatch.bats:349`). Any fix has to decide deliberately whether that assertion still holds once `main_root` also consults the payload cwd; that decision is the substance of the fix, which is why this is filed rather than folded into other in-flight work on this hook.

No known reachable trigger under the current harness: a session's process cwd is not normally outside every git repository. Filing this to track a genuine gap in the guard's coverage, not an active incident.

## Suggested fix

Give `main_root`'s derivation the same payload-cwd-first, process-cwd-fallback treatment `current_root` already has: derive `main_root` via `git -C "$payload_cwd" rev-parse --git-common-dir` when a valid `payload_cwd` is available, falling back to the current process-cwd form when it is not. Update, or explicitly re-justify, the pinned bats assertion to match whichever behavior the fix settles on.

Handler: prompt

Issue #841, opened 2026-07-18
tech-debt: Edit tool silently desyncs from disk when a stale absolute path targets the wrong git worktree


stevensacks:
**Location:** `.claude/skills/gaia/references/isolation.md:183` (the `RESOLVED_ROOT` export section, which tells every later caller to interpolate the resolved path "rather than re-deriving it ad hoc")

## Failure mode

After `EnterWorktree` switches a session into `.claude/worktrees/<branch>/`, the guidance to use `RESOLVED_ROOT` for every later reference is scoped to Bash (`git -C "$RESOLVED_ROOT" …`) and sub-agent dispatch prompts. It says nothing about the Edit/Write/Read tools, whose `file_path` parameter takes an absolute path directly rather than a repo-relative one resolved against the shell's cwd.

Concretely, during the #681 drain: after entering the worktree, an `Edit` call for `CHANGELOG.md` was issued with the stale absolute path from before the worktree switch (`/Users/…/gaia/CHANGELOG.md`, the main checkout) instead of the worktree's own path. Both paths are real, tracked, valid files, so the edit succeeded silently against the wrong checkout. Because the change landed outside the worktree, `git add`/`git status`/`git commit` run inside the worktree correctly reported "nothing to commit", with no error connecting the two.

Worse, a second `Edit` call then targeted the worktree's (still-unmodified) copy with the same `old_string`. The Edit tool's own cached view of "what this path currently contains" had been updated by the *first* call (against the other checkout), so it reported `old_string` as already absent, i.e. it believed its own prior edit had landed in the file it was now being asked to patch. Disk and the tool's internal state disagreed for several turns; a raw `git status`/`grep`/`sed` check on the actual worktree file was needed to detect the desync, and recovery required routing the edit through a Python script with an explicit, hardcoded absolute path instead of the Edit tool.

This is a silent-wrong-write footgun with no error signal at the moment of the mistake: a maintainer could just as easily believe a doc/config edit merged from the worktree branch when it actually sits as an uncommitted change in the main checkout (or vice versa), only discovering the mismatch by chance.

## Suggested fix

Two options, not mutually exclusive:

1. **Cheap, immediate:** extend `isolation.md`'s `RESOLVED_ROOT` section to explicitly cover Edit/Write/Read tool calls, not just Bash: state that every such call after entering a worktree must prefix `file_path` with `RESOLVED_ROOT`, and that a stale absolute path from before the switch fails **silently** (no error) because both checkouts are valid on disk.
[...]

PR #952, opened 2026-07-20
fix(hooks): let worktree-mode plan runs write the main-checkout ledger

### LINKED ISSUE #934: Worktree-mode plan orchestrators cannot write PROGRESS.md to the main-checkout plan folder
`.claude/hooks/block-worktree-path-mismatch.sh:95`

## Failure mode

A plan orchestrator running under worktree isolation cannot write its own phase-findings ledger.

`.claude/skills/gaia/references/plan.md` requires two things at once: the plan folder stays in the main checkout ("The plan folder stays in the main checkout", it is not among the four paths `link-worktree.sh` symlinks into the worktree), and the orchestrator writes `PROGRESS.md` there after every phase, plus `SUMMARY.md` one directory up at consolidation.

The guard's shared-state exemption at `:95-100` covers only `<main>/.gaia/local/audit/`, `debt/`, `telemetry/`, and `cache/shared/`. `.gaia/local/specs/` and `.gaia/local/plans/` are not exempt. So when the orchestrator's cwd is the linked worktree, `main_root != current_root` arms the gate, the ledger's `file_root` resolves to `main_root`, and `:105-107` denies the write.

Trigger: any `/gaia-plan` run whose pre-flight isolation resolves to `worktree` mode, at the moment the orchestrator writes `PROGRESS.md` via `Edit`/`Write`/`MultiEdit` after phase 1. Outcome: the write is denied. The ledger is the anchor `.gaia/scripts/plan-resume-point.sh` reads to compute a resume point, so a run that hits this both loses its cross-compression findings record and becomes unresumable, while the deny surfaces as a tool refusal mid-run rather than as anything naming the underlying contract conflict.

The guard's registration is `PreToolUse` on the `Edit|Write|MultiEdit` matcher only (verified: the guard's command appears under no other matcher and under no `Bash` matcher in `.claude/settings.json`), so a `Bash` heredoc append to the same path is unadjudicated and does land. That asymmetry is why this has stayed latent: any orchestrator that happened to write the ledger through a shell redirect never hit it.

Found by the SPEC-054 decomposition audit (finding DP-005), which hit it on that SPEC's own plan. That plan works around it locally by mandating the `Bash` route for its own ledger writes, but the conflict lives in the shared reference and the guard, so it recurs for every other worktree-mode plan run.

## Suggested fix

Two defensible routes, and the choice is the substance of the work:

1. Add `<main>/.gaia/local/specs/` and `<main>/.gaia/local/plans/` to the exemption `case` at `:95-100`. Honest about intent (these writes are legitimate and required by design) but it widens the guard's blind spot to the exact directories a stale absolute path is most likely to name during plan work.
2. Make the `Bash`-write mechanism explicit in `.claude/skills/gaia/references/plan.md` for every ledger write, leaving the guard's exemption surface untouched. Keeps the guard narrow, but relies on a matcher gap that would silently become load-bearing, and would break if the guard ever gains a `Bash` matcher.

Whichever is taken, the reference and the guard have to agree, and `.gaia/tests/hooks/block-worktree-path-mismatch.bats` needs a case pinning the decision so the next change cannot reopen it.

Handler: plan

### LINKED ISSUE #944: block-worktree-path-mismatch.sh: the file_path chain's dirname -- and CDPATH='' guards have no test coverage
`.claude/hooks/block-worktree-path-mismatch.sh:128-129`:

```bash
target_dir=$(dirname -- "$file_path")
resolved_target_dir="$(CDPATH='' cd "$target_dir" 2>/dev/null && pwd -P)" || exit 0
```

## The gap

These two guards, `dirname --` on the `target_dir` line and `CDPATH=''` on the `resolved_target_dir` line, have no bats coverage. Mutation testing confirms the gap is real rather than theoretical: a mutant removing `CDPATH=''` from this line, and a separate mutant weakening `dirname --` to `dirname` on the line above, both survive the full suite unchanged.

The rest of the suite is otherwise well mutation-covered, which is what makes these two survivors conspicuous rather than an expected gap. **The guards are correct today.** The gap is that their correctness is invisible to the suite, so a future edit dropping either one regresses silently.

## Nothing here is exploitable, and the severity reflects that

An earlier revision of this issue described the missing `CDPATH=''` as a live bypass. That is wrong on two independent grounds, both measured:

1. The guard is present and correct. The deny-to-allow flip exists only in a constructed mutant.
2. Even absent the guard, the harness does not produce the shape it needs. `.claude/skills/gaia/references/isolation.md:186` states `file_path` is an absolute path with no cwd resolution, and an absolute `file_path` is immune regardless of `CDPATH`.

A three-case probe, each run against the pristine hook and against the `CDPATH`-removed mutant:

| Case | pristine | guard removed |
|---|---|---|
| A: `CDPATH` to the exempt tree, relative `file_path` | deny | allow |
| B: `CDPATH` to a non-exempt dir, relative `file_path` | deny | deny |
| C: `CDPATH` poisoned, absolute `file_path` | deny | deny |

B and C establish that both qualifiers are necessary conditions. Case C is the one that governs severity: the tool contract only ever sends absolute paths.

So this is a test-coverage gap around a defense-in-depth guard, not an unfixed vulnerability.

## Why it is still worth closing

The hook has two structurally similar guard pairs, and they are not "one matters, one does not". They differ in **how each is made unreachable**:

- The `payload_cwd` pair is unreachable by an invariant **this script enforces itself**, the absolute-cwd `case`. That can be reasoned about locally and holds permanently. Its own `CDPATH` mutant also survives the suite, correctly, and is not part of this finding.
- The `file_path` pair is unreachable only by an **external convention the script does not enforce** (`isolation.md:186`). If the harness ever emits a relative `file_path`, this guard is the only thing standing there.

That asymmetry is the argument for keeping `CDPATH=''` on the `file_path` line and eventually pinning it with a test: its unreachability rests on someone else's contract, and nothing would catch its removal.

## Suggested fix

A verified starting point, not a settled design. A killer test for the `CDPATH` mutant is constructible by exporting `CDPATH` to a decoy directory sharing a basename with the main checkout's `.gaia/local/audit/`, symlinking from the worktree fixture to the main checkout, and issuing a relative `file_path`. Confirmed directly: it passes against the pristine hook and fails against the mutant.

The `dirname --` mutant needs a sibling test of the same shape (a `file_path` whose `dirname` component reads as an option to a bare `dirname`, most directly a leading-dash target). The killer test above does not exercise it.

Handler: prompt

stevensacks:
fix(hooks): let worktree-mode plan runs write the main-checkout ledger

.claude/skills/gaia/references/plan.md puts the plan folder in the main
checkout by contract and has the worktree-mode orchestrator write
PROGRESS.md there after every phase, but block-worktree-path-mismatch.sh
denied every main-checkout write from a linked worktree except the four
symlinked shared-state dirs. Any /gaia-plan run that resolved to worktree
mode therefore lost its phase-findings ledger and its resume point, and
the deny surfaced as a bare tool refusal mid-run.

Exempt <main>/.gaia/local/plans/ and <main>/.gaia/local/specs/ in a
separate case with its own rationale: unlike the symlinked trees, these
are exempt because they have no worktree-side copy at all, so the
main-checkout path is the only one that resolves to a real ledger and the
#841 silent-wrong-write it guards against needs a valid twin in both
checkouts. handoff/, red-ledger/, forensics/, worktree-locks/, and the
non-shared cache/ subdirs keep a real per-worktree copy and stay denied.

State the Edit/Write route in plan.md so the matcher gap on the Bash
route never becomes load-bearing, and correct that paragraph's
symlink list, which omitted .gaia/local/debt/.

Also pin the file_path chain's `dirname --` and `CDPATH=''` guards, which
no invariant makes unreachable (they rest on isolation.md's external
absolute-path convention, unlike the payload_cwd pair). Both new cases
are mutation-verified: each fails against a hook with its own guard
removed, and the two ledger cases fail against the pre-fix hook.

Closes #934
Closes #944

## Verification

- `.gaia/tests/hooks/block-worktree-path-mismatch.bats`: 38/38 under bash 5 (`.gaia/scripts/bats5.sh`).
- Mutation-verified, each mutant restored after its run:
  | Mutation | Expected killer | Result |
  |---|---|---|
  | `CDPATH=''` removed from the `file_path` line | test 34 only | `not ok 34` |
  | `dirname --` weakened to `dirname` | test 35 only | `not ok 35` |
  | plans/specs exemption removed (pre-fix hook) | tests 12, 13 | `not ok 12`, `not ok 13` |
- End-to-end through the live harness from inside this worktree: a `Write` to the main checkout's `.gaia/local/plans/<PLAN>/PROGRESS.md` is allowed, and a `Write` to the main checkout's `.gaia/local/handoff/` is still denied by the same guard.
- `shellcheck -x` clean; `.gaia/tests/shell-lint.sh` passes (163 scripts, 145 suites, array-guard).
- Quality Gate skipped per its own skip rule: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.

## Note

`#940` is the third member of this file's cluster and is deliberately **not** in this PR: it carries its own design decision about a pinned fail-open assertion and touches `main_root` resolution.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.mJ3F650adW

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.7l34gjPjtJ

PR #955, opened 2026-07-20
fix(hooks,ci): clarify no-verify deny message; tolerate transient advisory-comment failures in code-review-audit

### LINKED ISSUE #912: code-review-audit: a 503 on an informational status comment reds the job
**Location:** `.github/workflows/code-review-audit.yml:1572`, step `Status - skipped (no source changes)`.

## Failure mode

A transient GitHub API failure while posting a purely **informational** PR comment reds the `code-audit-frontend` check, even though the audit itself never ran and nothing is actually wrong.

The step runs under `set -eu` and calls the status helper bare:

```bash
bash "$RUNNER_TEMP/cra-status-upsert.sh" "$PR_NUMBER" "code-review-audit skipped: ..."
```

Any non-zero exit from that call aborts the step and fails the job. Observed on PR #908 during a GitHub outage:

```
MEMBERS_PENDING: code-audit-maintainer-node
SUCCESS_STAMPED: false
gh: No server is currently available to service your request. (HTTP 503)
##[error]Process completed with exit code 1.
```

The same workflow already treats the genuinely load-bearing call as non-fatal. At `:941`, the `GAIA-Audit` commit-status POST, which is what actually gates the merge button, is wrapped:

```bash
if ! gh api "repos/${GITHUB_REPOSITORY}/statuses/${HEAD_SHA}" --method POST ...; then
  echo "code-review-audit: GAIA-Audit status POST to ${HEAD_SHA} failed (non-fatal); local merge path will re-stamp." >&2
fi
```

So the tolerance is inverted between two sibling terminal-status paths: the important POST degrades gracefully, the advisory comment hard-fails the job.

## Reachability

The step is gated only on `gated == 'false' && has_source == 'false'`. It carries no `should_run` gate, so it is reachable in `local` mode (this repo's `default_mode`) on any PR whose diff touches no file in the frontend auditor's glob set. That is the normal shape of a maintainer PR, for example one touching only `.gaia/cli/**`, `wiki/**`, and `CHANGELOG.md`, which is exactly what #908 was. It will recur on such PRs whenever the GitHub API blips.

## Impact

Bounded, which is why this is filed as a suggestion rather than important:

- `code-audit-frontend` is **not** in the declared-required check set (confirmed via `.gaia/scripts/verify-required-checks.sh`), so a red here never blocks a merge.
- The merge gate is the `GAIA-Audit` commit status, which is unaffected.
- Unlike #909, which exits 1 on a clean tree every single run, this only fires during an upstream outage.

The cost is a misleading signal. A red `code-audit-frontend` row reads as "the frontend audit failed" when the audit never ran at all, and in practice it did cost a reviewer real time chasing a non-existent defect.

## Suggested fix

Wrap the `cra-status-upsert.sh` invocations in this step, and in the sibling `Stand down (local-mode, no override)` terminal-status step, in the same non-fatal guard `:941` already uses. The comment is advisory; failing to post it should log to stderr and exit 0, not fail the job.

Consider also whether these terminal-status steps want `continue-on-error: true`, which is the pattern the `Status - audit aborted` step relies on for a related reason.

Handler: prompt

---

Surfaced on PR #908 during a GitHub HTTP 503 outage. The job behaved correctly otherwise: `should_run=false` for this repo, so every audit step was skipped and no audit ran on CI. This is purely about failure tolerance in the bookkeeping path. Owned by `code-audit-github-workflows` (`.github/workflows/**`).

### LINKED ISSUE #915: block-no-verify.sh: deny message does not name its documented false-positive case
`.claude/hooks/block-no-verify.sh:69` (the `floor_msg` helper)

## Failure mode

`block-no-verify.sh` scans the whole Bash command string, so a `git commit` whose **message** merely names a bypass token is denied along with a genuine bypass attempt. That over-block is deliberate and documented in the hook's own header (lines 32-37):

> Residual fail-closed edge: a bypass token written literally INSIDE a commit message (e.g. `git commit -m "use --no-verify"`) still over-blocks. That is the safe direction; rephrase the message.

**This issue is not about that posture, which must not change.** The gap is diagnostic. `floor_msg` (line 69) emits a single reason for both cases:

```
Hook bypass on 'git commit' is forbidden (--no-verify). The Quality Gate floor
(typecheck/lint/test) runs via the Husky pre-commit hook, fix the failures,
don't skip the gate. See wiki/decisions/Quality Gate.md.
```

Trigger: commit a message that describes the guardrail set, e.g.

```bash
git commit -m "docs(hooks): document that --no-verify is denied at commit time"
```

Bad outcome: the operator is told they attempted to skip the Quality Gate and instructed to "fix the failures" when there are no failures and no bypass was attempted. Nothing in the message hints that a *mention* of the token inside `-m` text triggers the same deny, or that rephrasing resolves it. The documented workaround exists only in a source comment the operator is not reading at that moment, so the deny reads as a false accusation and the guard trains the reader to distrust it.

Same cry-wolf class as #910 and #911: a guard whose denial is correct-by-design but whose message cannot be acted on correctly.

## Suggested fix

Extend the deny reason to name the known case and its workaround, e.g. append:

> If this token appears only inside your commit message, that is this hook's documented over-block: rephrase the message (the gate was not bypassed).

Leave every matching rule, the command-position anchoring, and the whole-command fail-closed safety net exactly as they are. Do **not** add a carve-out that inspects or exempts `-m` text; that would reopen the bypass path the safety net at lines 127-146 exists to close. This is a string change in `floor_msg` only.

Handler: prompt

stevensacks:
## Summary

Two unrelated single-issue suggestion fixes, batched to share one trip through the audit gate.

- **#915**: `block-no-verify.sh`'s deny message named nothing about its own documented false-positive (a bypass token merely mentioned inside a commit message, not a real flag). The operator was told they'd skipped the Quality Gate with no failures to fix and no hint that rephrasing the message resolves it. `floor_msg` now appends a line naming that case and its workaround, on the commit path only (push carries no `-m` text the note could apply to).
- **#912**: `code-review-audit.yml`'s advisory PR status comment (posted via the `cra-status-upsert.sh` helper written into `$RUNNER_TEMP`) ran under `set -eu` with no failure tolerance, so a transient GitHub API blip (observed as an HTTP 503 on PR #908) failed the whole job even though the audit itself never ran. The load-bearing `GAIA-Audit` commit status is a separate POST that already tolerates this; the comment-posting path (update-existing, post-new, and the comment-lookup that decides between them) now logs a failure to stderr and skips the write instead of failing the job or guessing, matching the existing non-fatal pattern. Mirrored into the source template and rebuilt build artifact per the workflow-distribution rule, so adopters get the fix on their next `/setup-gaia`/`/update-gaia`.

## Audit round 1 findings (addressed)

`code-audit-maintainer-shell` withheld over one Important finding; the other two members cleared. All three findings addressed:

- The over-block bats test drove a real bypass command (`git commit --no-verify -m "x"`), leaving the message it claims to pin actually untested by any test in the suite. It now drives the message-mention case (`git commit -m "use --no-verify next time"`) and keeps the real-flag case as its own separate test.
- `floor_msg`'s over-block clause is commit-only advice (push has no message text a token could be mentioned inside) but was appended unconditionally, leaking dead advice onto push denials. It now appends only when the subcommand is `commit`.
- All three auditors independently converged on the same gap: `cra-status-upsert.sh`'s comment-lookup ran through `| head -n1`, whose exit status is always 0 regardless of whether `gh api --paginate` on its left side failed. A transient lookup failure therefore read as "no existing comment" and posted a fresh one blind, risking a stale duplicate `<!-- gaia:cra-status -->` comment sitting on the PR indefinitely. The lookup's own exit status is now checked directly and the upsert is skipped on failure, the same non-fatal posture as the PATCH/post paths this PR already hardened.

## Noted, not filed

- Silent permission misconfiguration: with both write branches non-fatal and the lookup guarded, dropping `pull-requests: write` from the workflow would make it post nothing on any path with no red signal beyond a stderr line in an otherwise-green run. Accepted, not coded around: no live failure mode today, it needs a separate misconfiguring change to trigger, and this file uses no annotation commands at all.
- Nothing else found out of scope for either issue; both premises held as filed.

## Test plan

- [x] `.gaia/tests/hooks/block-no-verify.bats`: 24/24 passing. Confirmed red-then-green for both the over-block message test (now driving the correct command) and the new push-omits-the-note test.
- [x] `.github/audit/tests/ci-status-member-gate.bats`: 66/66 passing. Confirmed red-then-green for all three `cra-status-upsert.sh` non-fatal-path tests, including the new comment-lookup-failure test.
- [x] `.gaia/tests/shell-lint.sh`: clean.
- [x] `pnpm -C .gaia/cli test --run audit-template-dogfood`: 5/5 passing, confirms live workflow == source template == build artifact after both rounds of edits.
- [x] Quality Gate: no staged file matches the gate's trigger set (`.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config), so it is skipped per its own rule; confirmed via the documented grep, no matches, on both rounds.

Closes #912
Closes #915


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.zIJbG8qpUi

PR #956, opened 2026-07-20
fix(hooks): accept absolute whitelisted paths and judge rm segments on their own flag

### LINKED ISSUE #943: block-rm-rf.sh's absolute-path deny arm fires before the .gaia/local/audit/ whitelist, conflicting with shell-cwd.md's own absolute-path mandate
**Location:** `.claude/hooks/block-rm-rf.sh:408` (the `/|/*)` absolute-path case arm) versus `.claude/hooks/block-rm-rf.sh:458` (the `.gaia/local/audit/*|./.gaia/local/audit/*)` whitelist arm).

## Failure mode

Two of this repo's own conventions cannot both be satisfied at once for a whitelisted scratch path.

`.claude/rules/shell-cwd.md` mandates an absolute path on every Bash call, repo-wide, and gives its own worked example at line 11: `` `rm /abs/path/file`, not `cd /abs/path && rm file` ``. Its stated reason is that a `cd` inside a Bash call persists for the rest of the session and breaks every relative-path hook.

`block-rm-rf.sh`'s case statement checks the absolute-path arm first:

```bash
case "$tok" in
  /|/*)
    # Allow specific safe absolute prefixes, currently none whitelisted absolutely.
    deny "BLOCKED: rm -rf of absolute path '$tok' is forbidden."
    ;;
```

and its `.gaia/local/audit/` whitelist arm, forty lines later in the same case statement, matches only a relative spelling:

```bash
  .gaia/local/audit/*|./.gaia/local/audit/*)
    : # whitelisted
    ;;
```

An agent following `shell-cwd.md` as written, deleting a gitignored body file under the whitelisted `.gaia/local/audit/` directory with an absolute path (`rm -f /abs/path/to/repo/.gaia/local/audit/some-file.md`), trips the absolute-path arm and is denied. The `.gaia/local/audit/*` whitelist exists precisely to allow this directory's scratch files to be removed, and the file-tech-debt skill's own recipe mandates writing and then deleting a file there. The absolute-path arm's own comment concedes the gap: "currently none whitelisted absolutely."

## Confirmed collateral behavior

The denied `rm -f` was the second of two commands in one Bash tool call, issued after a `gh issue create --body-file <same path>`. The whole tool call was blocked: after the denial, `gh issue list` showed no new issue and the two target files were unchanged on disk, confirming the `gh issue create` invocation never ran either. This matches the general PreToolUse hook contract: a hook evaluates `tool_input.command` as a whole and returns one allow/deny decision for the entire tool invocation before any of it reaches the shell; there is no partial or per-line execution within a single Bash tool call. The practical effect here is that a false-positive deny on an unrelated cleanup command silently dropped a sibling command with no output pointing at what happened; only cross-checking `gh issue list` against expectation surfaced it.

## What a fix has to reconcile

Not prescribing one, naming the tension: two things pull against each other.

1. **Reachability.** The whitelist arms sit after the absolute-path arm in the same `case`, and the absolute-path arm denies unconditionally with no whitelist consultation at all. Adding a literal absolute spelling to the whitelist list as it exists today would never be reached; whatever a fix does has to intercept before or within the absolute-path arm itself, not simply extend the list that follows it.
2. **Resolution without a live dependency.** The whitelist arms match an exact relative literal spelling, which presumes the guard is working over a relative token. Recognizing an absolute spelling of the same target requires knowing the repo root to compare against, and this hook is a pure text matcher over the raw command string with no computed state (per its own header, a target the shell computes rather than spells, including `$(git rev-parse --show-toplevel)`, is explicitly out of reach by design). Introducing a live root lookup conflicts with that design principle, and hardcoding a literal absolute prefix would itself violate this repo's own portable-path rules (`.claude/rules/repo-relative-paths.md`, `.claude/rules/instruction-files.md`), which forbid a machine-specific absolute path baked into a `.claude/`-distributed file.

Handler: prompt

### LINKED ISSUE #950: shell-cwd.md's absolute-path mandate contradicts block-rm-rf.sh's relative-only whitelist, and the guard matches its own documentation
`.claude/rules/shell-cwd.md:11`

## Failure mode

`shell-cwd.md` is an always-loaded rule, and it states its absolute-path mandate without qualification. Its own worked example at line 11 (an absolute-path `rm`, contrasted against a `cd`-then-`rm` form) is the exact shape the destructive-command guard denies.

`block-rm-rf.sh` whitelists `.gaia/local/audit/` in its **relative** spelling only, and its absolute-path arm denies before any whitelist arm is consulted. So an agent that has read the always-loaded rule, and deletes a scratch file under `.gaia/local/audit/` with an absolute path, is denied on a path the guard is meant to permit.

The exception does exist, but only at the point of use: `.claude/skills/file-tech-debt/SKILL.md:72` says to keep the cleanup path repo-relative "because the destructive-command guard whitelists this directory in its relative spelling only." That file loads on demand. An agent working from the always-loaded rule alone has no way to know the carve-out exists, so the rule and the guard disagree for any reader who has not happened to load that one skill.

Observed directly: an agent instructed to use absolute paths issued an absolute removal against a body file under `.gaia/local/audit/` and was denied.

## The guard also matches its own documentation

A second, sharper reproduction, hit while filing this issue. A Bash call whose heredoc **quoted** `shell-cwd.md:11` verbatim, as issue prose, was denied with `BLOCKED: rm -rf of absolute path '/abs/path/file`,' is forbidden`.

Nothing in that call was a removal. The guard is a pure text matcher over the raw command string, so it matched the rule's own worked example sitting inside a heredoc body, where it was data rather than a command. Any attempt to quote this rule's canonical example inside a Bash heredoc trips the guard, which means the guard denies documentation *about* the guard.

This is a false positive on non-command text, distinct from the whitelist-spelling collision above, and it shares a root cause with it: the matcher inspects a token's shape without regard to whether that token occupies a command position.

## Compound-call denial drops sibling commands

Both reproductions above landed in Bash calls carrying more than one command, and in both the entire call was dropped. A `PreToolUse` hook returns a single allow/deny decision for the whole invocation before any of it reaches the shell.

The first time, a denied removal silently dropped a `gh issue create` standing beside it: no issue filed, no output naming the cause, caught only by cross-checking `gh issue list` against expectation. The second time, the denial dropped label creation, a body-file write, a dedup re-check, and an issue creation.

That property is documented at `.claude/skills/file-tech-debt/SKILL.md:72`, which mandates two tool calls for exactly this reason. It is **not** documented in `wiki/concepts/Claude Hooks.md`, which is where a reader would look for how hooks adjudicate. The property is general to every `PreToolUse` guard rather than specific to one skill's cleanup step, so documenting it only there leaves every other compound-command site to rediscover it.

## Suggested fix

Three independent changes; none prescribes another.

1. Give `shell-cwd.md` a short exception clause naming guard-whitelisted paths, so the always-loaded rule stops contradicting the guard. Cheapest of the three, and it addresses the discoverability half.
2. Make the matcher skip tokens that are not in command position, so quoted prose and heredoc bodies stop being scanned as commands. This is the deeper fix and needs a design decision the current code does not settle, since the hook is deliberately a pure text matcher with no parse of shell structure.
3. Document the one-allow/deny-decision-per-Bash-call property in `wiki/concepts/Claude Hooks.md`, with the consequence that a denied clause silently drops its siblings.

Related: #943 covers the guard side of the whitelist collision, proposing to reconcile the absolute-path arm with the whitelist. This issue is the rule-and-documentation side plus the false-positive-on-prose case. #943 names constraints that bear on fix 2 as well: the hook is a pure text matcher by design, and a hardcoded absolute prefix would violate this repo's portable-path rules.

Handler: prompt

### LINKED ISSUE #954: file-tech-debt SKILL.md: the step-4 grading pointer does not land on its answer, plus a self-referential step reference
`.claude/skills/file-tech-debt/SKILL.md:130` (and `:48`)

## Failure mode

Two `prose/high-indirection` findings in the same file, both costing a reader a wasted hop while following the recipe live.

**1. The graded-vs-ungraded pointer does not land on its answer (`:130`, `:133`).**

Step 4 sends the reader to step 7 to decide whether a filing carries a `difficulty:` label: the graded form applies to "one of the two grading routes (step 7)". Step 7 opens by declaring itself the source of truth for "the permitted values and for choosing between them" and never names the routes. The answer is split across two later bullets whose headings advertise something else: **Ungraded filings** (`:130`) lists the three exempt paths, and **Where the grade comes from** (`:133`) names the two producers. A reader who follows the step-4 pointer lands on a section that does not answer the question it sent them to answer, and has to scan two more bullets to reconstruct it.

The subject is genuinely intricate (five routes, three of them exempt for three different reasons), so this is a placement defect, not a volume one.

**2. A self-referential forward pointer (`:48`).**

Inside step 4, sub-item 2 justifies the unique body-file name with "because step 4 deletes it". The reader is already inside step 4, so it reads as a forward reference to a different section and costs a re-scan to resolve.

## Suggested fix

1. State the graded/ungraded split once at the head of step 7, where the step-4 pointer lands: two graded producers, everything else ungraded. Leave `:130` and `:133` carrying only their distinct payloads (the mixed-adopter-state guarantee, and the edit-must-reach-both-producers obligation). No coverage is lost.
2. Change "because step 4 deletes it" to "because sub-step 4 below deletes it".

Found by the `code-audit-maintainer-prose` member on PR #952, where the file was out of scope: it entered that branch through a merge of `main` carrying #951 and was not that PR's authorship. Advisory-only member, so neither finding gated anything.

Handler: prompt

stevensacks:
Closes #943
Closes #950
Closes #954

## The motivating incident

An agent working inside a linked worktree could not delete its own scratch files. Every spelling of the removal was denied:

- `rm -rf <absolute path>` denied
- `rm -f <absolute path>` denied
- `rm <absolute path>`, no flags at all, denied

It completed the deletion with `find -delete`.

That last line is the case for fixing rather than documenting, and it is stronger than the misfire itself. **The guard did not prevent a deletion.** It redirected a competent agent to a less legible tool that the deny list does not inspect at all, and the file was removed anyway with worse observability. A destructive-command guard that reliably converts `rm` into `find -delete` produces the outcome it exists to prevent. A documented carve-out cannot fix that, because the problem is not that the guard is occasionally annoying; it is that absolute-path cleanup was *impossible* from a worktree while `shell-cwd.md` mandates absolute paths everywhere, leaving an agent with no legal way to delete a file it created.

Reproduced against the unmodified guard, and the incident's three-way denial turned out to have **two independent causes**, only one of which #943 describes.

## Three defects, not two

| # | Defect | Named in an issue? |
|---|---|---|
| 1 | Absolute whitelist unreachable behind the absolute-path deny arm | #943, correctly |
| 2 | Flag checked once per command, targets judged per segment | No, and #950 misdiagnoses it |
| 3 | A hyphenated **path component** ending in `r`/`f` read as a destructive flag | No |

**Defect 3 is what denied the flagless `rm`,** and it is the one nobody had identified. The probe matched a bare `-[a-zA-Z]*[rRfF]` anywhere in the segment. The incident's own branch directory is `c-text-matcher-guards`, which contains `-matcher`. So a removal carrying no flag at all was read as a destructive `rm -r` purely because of a directory name:

```
DENY   | rm /Users/you/projects/my-matcher/notes.md      (no flags anywhere)
DENY   | rm /Users/you/projects/my-perf/notes.md
ALLOW  | rm /Users/you/projects/my-thing/notes.md        (identical, name ends in 'g')
```

`-matcher`, `-perf`, `-refactor`, and `-user` are ordinary branch and directory names. A genuine flag is always its own shell word, so the probe now requires a token boundary in front of the dash. Quote and backslash bytes count as boundaries alongside whitespace, since bash drops them during quote removal and `rm "-rf" /` is a real flag that must stay denied.

## Before and after, on the incident's own path shapes

```
BEFORE AFTER  |
# worktree path into .gaia/local/audit (symlinked to the main checkout)
DENY   ALLOW  | rm <path>            (no flags)
DENY   ALLOW  | rm -f <path>
DENY   ALLOW  | rm -rf <path>

# the same directory spelled as the resolved main-checkout path
DENY   ALLOW  | rm / rm -f / rm -rf

# still denied, deliberately, see "declined" below
DENY   DENY   | rm -rf <worktree>/scratch-notes.md
```

## The decision: carve-out, not command-position anchoring

**Carve-out**, and the incident does not change that. Three pieces of evidence.

**1. Command-position anchoring is not an open question.** The file's header already decided it, twice: making a quoted `rm` not-a-command would fix prose false-denies *and would also allow* `bash -c "rm -rf /"`, `ssh host "rm -rf /"`, and `eval "rm -rf /"`. #950's fix 2 is precisely that change, described as needing "a design decision the current code does not settle." The code settles it. Re-opening it trades annoyance for real bypass.

**2. None of the three defects needs command-position knowledge.** This is what made the carve-out cheap rather than resigned. #943 is whitelist *reachability*. Defect 2 is *scope*: the flag test and the target test applied to different things. Defect 3 is *tokenization* of the flag itself. All three are fixable inside the pure-text-matcher design, and none inspects shell structure.

**3. Anchoring would not have fixed the incident anyway.** As the ask notes, anchoring alone leaves the absolute arm firing before the whitelist. It also would not have touched defect 3, since `-matcher` sits inside an operand the anchor would happily judge.

## Absolute vs relative: absolute wins, both accepted

**`shell-cwd.md` is unchanged. The guard bent.** The rule is always-loaded and repo-wide; the guard was the local exception with the narrower claim. #950's fix 1 proposed an exception clause in the rule; that is now unnecessary, and adding it would have documented a contradiction instead of ending one.

Every whitelisted scratch path matches in **both** spellings, as a **suffix** on the scratch segment (`/?*/.gaia/local/audit/*`). That threads #943's stated needle: no live `git rev-parse` (the computed state this guard is designed without) and no hardcoded absolute prefix (which `instruction-files.md` forbids in a `.claude/`-distributed file).

This also covers the worktree case in both directions, which matters because `.gaia/local/` is symlinked out of a worktree to the main checkout: an agent may legitimately spell either end, and both now pass.

Two deliberate boundaries:

- `/?*/` requires a non-empty parent segment, so `/dist` and `/.gaia/local/audit/x` stay denied as filesystem-root removals.
- An absolute path carrying a `..` segment is intercepted *ahead* of the whitelist. Without that ordering the suffix match is a bypass: `/repo/.gaia/local/audit/../../../.git` spells a whitelisted directory and lands on `.git`.

## Declined: whitelisting the worktree tree itself

The ask was that the absolute arm stop denying cleanup of "paths the agent owns." It now does for the agent's **scratch**, which is `.gaia/local/*`, the directory the incident involved and the one the file-tech-debt recipe writes to.

I did **not** whitelist `.claude/worktrees/<name>/` wholesale, and recommend against it. A worktree is a full checkout, so the pattern that would allow `<worktree>/scratch-notes.md` also allows these, all four of which are denied today:

```
DENY (today) | rm -rf <worktree>/.claude
DENY (today) | rm -rf <worktree>/.git
DENY (today) | rm -rf <worktree>
DENY (today) | rm -rf <worktree>/app/routes
```

Whitelisting the tree would flip all four to ALLOW, punching a hole straight through the `.git` and `.claude` protection this guard exists for, inside every worktree. Non-scratch files in a worktree are tracked by git and are recoverable; scratch is not, which is why scratch is what gets the carve-out. If a concrete case appears where an agent must hand-delete an untracked non-scratch file from a worktree by absolute path, that is a separate decision with its own evidence.

## Audit round: one real regression, found and fixed

`code-audit-maintainer-shell` caught a genuine **deny-to-allow regression this diff introduced**, and it was right. The new whitelist arm requires a non-empty parent segment via `/?*/`, but a glob `?` matches a literal dot, so a single `.` satisfied it. The normalization block collapsed `//` and stripped a *leading* `./`, but never an *interior* `/./`, so the token reached the arm intact:

```
                                 base    head (before fix)
rm -rf /./dist                   DENY    allow
rm -rf /../dist                  DENY    allow
rm -rf /.//dist                  DENY    allow
rm -rf /././dist                 DENY    allow      (not in the report; found while fixing)
rm -rf /./.gaia/local/audit/x    DENY    allow
```

`/./dist` and `/../dist` both resolve to `/dist`, the filesystem-root removal that arm exists to refuse. My own comment claimed `/?*/` prevented exactly this, so the code contradicted its own documentation.

Fixed by normalizing rather than by adding dot-segment arms, which keeps the arms reading one spelling of a target (this block's whole purpose). The two substitutions run in **one** loop to a fixed point because they feed each other: `/.//x` needs the slash collapse before the dot collapse can see a `/./`, and `/././x` needs a second pass because the replacements do not overlap. Two sequential loops would each terminate on a token the other still reduces. `..` is deliberately not collapsed, since where it lands depends on the tree; the traversal arm denies it instead, and now covers the leading `/../x` and `/..` spellings the interior patterns could not reach.

Worth noting for the record: this finding was produced by *constructing* commands, not by tripping over one in the wild, which is precisely the case for widening the policy bar below.

## Policy bar widened to admit an executable reproduction

The first draft set the bar at an observed incident with a transcript. That was too narrow in a way that would have suppressed the finding above. A failing bats case run against the real hook is deterministic and re-runnable, so it is if anything **stronger** evidence than a one-time transcript, yet a strict reader classifies constructing it as "reasoned to from the source" and declines to report. The bar now reads:

> A porosity finding requires a CONCRETE, RUNNABLE ARTIFACT, in one of two forms. Either an observed incident, a command that actually ran and should not have, or was actually denied and should not have been, with the transcript that shows it; or an executable reproduction, a failing test case against this hook that anyone can re-run.

The anti-speculation intent is unchanged: what the bar excludes is the argument with no artifact behind it, not the method of producing one.

## Header policy rewrite

The old closing line, "Known holes are tracked as tech debt," instructed every future auditor to file hypothetical porosity findings forever. Replaced with:

> Some *literally spelled* targets also pass. Treat the deny list as a floor, never a ceiling: it is not exhaustive, and for a heuristic text matcher completeness was never on offer.
>
> That incompleteness is the design, not a standing defect backlog. Do NOT file a porosity finding on the strength of a hypothesis, a hole reasoned to from the source, or an audit that enumerates shapes this guard does not catch. The supply of those is unbounded, each one costs real review, and this guard is explicitly the second layer behind settings.json permissions rather than the thing standing between anyone and disaster. A porosity finding requires an OBSERVED incident: a command that actually ran and should not have, or was actually denied and should not have been, with the transcript that shows it. Bring one and it is worth fixing. Without one, the correct response to "this guard could be evaded" is yes, it could.

The incident above is exactly the evidentiary bar that policy sets, and is the model for what a fileable porosity finding looks like.

## Also landed

- **`wiki/concepts/Claude Hooks.md`** gains the one-decision-per-tool-call property (#950's fix 3), previously documented only in `file-tech-debt/SKILL.md` at one point of use though it is general to every `PreToolUse` guard. Stated with its consequence: a denied clause silently drops its siblings.
- **`file-tech-debt/SKILL.md`** no longer says to keep the cleanup path relative "because the destructive-command guard whitelists this directory in its relative spelling only." That is now false. The two-tool-call mandate stays; it rests on the one-decision property, not on the spelling.
- **CHANGELOG** `## [Unreleased] / ### Fixed` entry.

## Verification

- `.gaia/tests/hooks/block-rm-rf.bats`: **177/177**, including 29 new tests. Run under **bash 5** via `.gaia/scripts/bats5.sh` per `.claude/rules/bats-assertions.md`, not just stock macOS bash 3.2.
- Full `.gaia/tests/hooks/` directory: no failures.
- `.gaia/tests/shell-lint.sh` passes in full (163 tracked scripts, 145 tracked suites, array-guard clean); prose-audit suites and `audit-machinery-complete.bats` pass.
- **Adversarial base-vs-head diff over 81 hand-built commands** (root, `$HOME`, `~`, cwd, globs, dot-globs, brace globs, `.git`, `.claude`, `node_modules`, `/etc`, quoted and backslash-split command words, quoted separators, `eval` / `bash -c` / `ssh` evasion, operand-first flags, long-form flags, and the whole dot-segment family): the only deny-to-allow flips are the two intended whitelist ones.
- TDD order honored throughout. The incident repro was written as a red test first. Notably, the four worktree-scratch tests went green on the whitelist commit alone while the two hyphenated-component tests stayed red, which is what isolated defect 3 as independent rather than a symptom of defect 1.

**Quality Gate: skipped, per its own skip logic.** No staged file matches the trigger set (`.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config); the staged surface is one hook, one bats suite, and three markdown files. `wiki/decisions/Quality Gate.md` states that `.claude/**`, `wiki/**`, and pure-markdown commits skip straight to the commit step. The shell-side checks above are the applicable signal and were run in its place.

## Noted, not filed

Per the new header standard. All four were **observed**, not hypothesized.

1. **`find -delete` is entirely outside the guard, including `find / -delete`.** Verified: every `find … -delete` form tested is allowed, before and after this PR. **Recommendation: leave it out of scope, deliberately.** Adding it would be the speculative deny-list expansion the new header policy exists to stop, and the honest reason the agent reached for `find` was that `rm` was broken. This PR fixes that cause. The general point stands on its own though, and is worth recording: the deny list only covers `rm`, so `find -delete`, `xargs rm`, `truncate`, `shred`, `mv` to a doomed path, and any script that opens a file itself are all unguarded. That is consistent with the header's own scope note (this is heuristic defense-in-depth behind `settings.json` permissions, not a sandbox) rather than a new gap. If `find -delete` is ever brought in scope, it should be on its own observed incident, not on this one.

2. **The guard denied two of my own Bash calls while writing this PR.** Once on the commit message (an apostrophe in prose opens a quote span, `neutralize_quoted_separators` rewrites the `&&` inside it to spaces, and two example clauses merge into one flagged segment), and once on a probe script whose loop body contained literal `rm -rf …/.claude` strings. Both are the documented-deliberate false-deny direction, and the header's own prescribed workaround (deliver the text through a file rather than the command string) resolved both. Worth knowing because it is a wider surface than the header's `git commit -m` example suggests: any multi-clause prose containing an apostrophe and a shell example is a candidate.

3. **The guard denies reads of its own source.** `grep -n 'rm -rf' .claude/hooks/block-rm-rf.sh` denies via the `.claude` arm. Same deliberate direction, unchanged here; the workaround is the Read tool. A predictable snag for anyone auditing this file.

4. **`<abs>/node_modules/dist` is now allowed** where the absolute arm previously denied it, since it matches `/?*/dist`. Removing a `dist` directory nested inside `node_modules` is not a footgun the guard advertises, so this reads as acceptable, but it is a real consequence of the suffix match and is recorded rather than buried.
5. **Advisory, not fixed, agreed with the reviewer.** `rm -"r"f /` (a quote *inside* the flag) is allowed, but base allowed it too, so it is pre-existing rather than introduced here. The `.gaia/local/*` whitelist entries take `dir/*` but not a bare `dir`, an asymmetry `dist` and `build` do not share. And the bats suites carry some SC2016/SC2317 shellcheck noise. None is a regression from this diff; all three are left alone deliberately rather than rotating audit digests for cosmetic changes.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.UQ7ChWFDcw

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.ZpUdG9Gu1C

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.G0NidocfFD

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.HuNlzpA4Fb

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.uW9j2AUi1Q

PR #957, opened 2026-07-20
fix(hooks): derive the worktree guard's main root from the payload cwd

### LINKED ISSUE #940: block-worktree-path-mismatch.sh: main_root derivation ignores payload cwd, so the guard goes inert when the hook's own process cwd leaves the repo

### LINKED ISSUE #953: gaia-active-plan.sh: symlinked shared-state comment omits .gaia/local/debt/, and the list is hand-copied across five surfaces
`.claude/hooks/lib/gaia-active-plan.sh:17`

## Failure mode

The header comment enumerates the state a linked worktree symlinks as `cache/shared, audit, telemetry, setup-state.json`. That list is missing `.gaia/local/debt/`, which `.gaia/scripts/link-worktree.sh:140` has symlinked since #930.

The comment is the local authority a reader consults while reasoning about this lib's `main_root` resolution, and it is stated as an exhaustive set ("symlinks only shared state"). A maintainer who trusts it concludes that `.gaia/local/debt/` is per-worktree state, and therefore that a `debt:in-progress` claim or a `refresh-requested` sentinel written from inside a linked worktree does not reach the main checkout's debt count. The opposite is true: that directory is symlinked, so those writes are shared. Anyone reasoning about worktree-side debt state from this comment reaches a wrong conclusion about a behavior the comment exists to explain.

No runtime impact: the comment is prose and the code around it derives `main_root` correctly regardless.

## Broader pattern

The symlinked-path list is hand-copied across surfaces and drifts. Copies at filing time:

- `.gaia/scripts/link-worktree.sh:136-140` (the authority)
- `.claude/hooks/block-worktree-path-mismatch.sh:135`
- `.claude/hooks/lib/gaia-active-plan.sh:17-18` (this issue, stale)
- `wiki/concepts/Local Working State.md:39`
- `.claude/commands/setup-gaia.md:28`

PR #952 removed one copy from `.claude/skills/gaia/references/plan.md` by replacing the enumeration with a pointer to `link-worktree.sh`, after that same paragraph was found carrying the identical stale list for a full release cycle. Two independent copies drifting the same way is the argument for reducing the rest rather than re-syncing them by hand.

## Suggested fix

Correct this comment, then apply the same reduction PR #952 used: replace each hand-synced enumeration with a pointer to `.gaia/scripts/link-worktree.sh`, keeping the enumeration only where a reader genuinely needs the membership rather than the exclusion. Each remaining copy is load-bearing only for its own claim, so verify per site before reducing; `block-worktree-path-mismatch.sh:135` in particular uses the list to justify its exemption arms and likely keeps it.

Found by the `code-audit-maintainer-prose` member as a cross-remit finding on PR #952, out of scope for that PR.

Handler: prompt

stevensacks:
Closes #940
Closes #953

## #940: main_root now derives from the same source as current_root

`block-worktree-path-mismatch.sh` had already migrated `current_root` to the PreToolUse payload's `cwd`, but `main_root` still came from `git rev-parse --git-common-dir` evaluated against the hook's own process cwd. When that cwd sat outside any git repository the call failed and the hook exited before any payload-aware logic ran, so the guard went inert. Inertness here is an **allow**, which is why this is worth fixing despite having no known reachable trigger: it is the failure mode that never announces itself.

Both roots now derive per candidate source, and one source supplies the pair:

- **Payload cwd wins** when it is absolute, resolves to a checkout, and does not contradict a process cwd that resolved to one.
- **Process cwd is the fallback** for an absent, relative, unresolvable, or contradicted payload cwd, and for a payload cwd that answers `--git-common-dir` but not `--show-toplevel` (a bare repo, a `.git` directory).

The absolute-payload-cwd invariant is untouched; the `case` that enforces it and its rationale moved position but not substance.

**The one deliberate behavior change.** The old foreign-repo check compared the payload's main root against `main_root`, which was by definition the process cwd's. With no process-cwd repo there is now nothing to contradict, so the payload is trusted. That cannot produce the false deny the check exists to prevent: both roots then come from the payload's own repo, and a target elsewhere can never equal `main_root`. Pinned by test.

**The pinned fail-open still holds.** `a session whose cwd is not inside any git repository fails open (allowed)` passes unchanged: that payload carries no `cwd`, so there is nothing to fall back to. #940 asked for that decision to be explicit, and the test now carries a comment saying why it holds, pointing at the companion case that must keep guarding.

**Coverage note worth flagging.** The loosening costs `a payload cwd inside an unrelated repository falls back to the process cwd` its ability to kill a dropped cross-check: with the process cwd in the main checkout it allows either way. A new case moves the process cwd into the worktree, where honoring a foreign cwd disarms the guard and the two verdicts separate. Mutation-verified against a hook with the check removed; the inertness case is mutation-verified against the pre-fix hook.

Suite: 42 tests, all green.

## #953: five surfaces, reduced where the copy is not load-bearing

All five located, plus two the issue did not name.

| Surface | Disposition |
|---|---|
| `.gaia/scripts/link-worktree.sh` (`link_one` calls + header) | **Authority.** Header said `cache/` where only `cache/shared/` is symlinked; corrected. |
| `.gaia/cli/src/setup/link-worktree.ts` (`SHARED_PATHS`) | Keep, now machine-checked against the authority. |
| `.claude/hooks/block-worktree-path-mismatch.sh` | Keep. The enumeration justifies why four dirs get exemption arms and the fifth (a symlinked file) needs none. Now machine-checked. |
| `.claude/hooks/lib/gaia-active-plan.sh` | **Reduced to a pointer.** This is the issue's headline stale copy. |
| `wiki/concepts/Local Working State.md` | Keep. Reader-facing membership is the point of the page; already correct. |
| `.claude/commands/setup-gaia.md` | **Reduced to a pointer.** Named three of five, not the four the issue reported. |
| `.gaia/tests/hooks/block-worktree-path-mismatch.bats` (two comments) | **Reduced.** One omitted `cache/shared`. Not named in the issue. |

Approach: both, scoped by what each copy is for. Reduction where the load-bearing claim is the *exclusion* (plans and specs are not shared; the symlinks may not exist yet), which never needs the membership. A deterministic check where the copy is exact and machine-comparable:

- `shared paths: the CLI's SHARED_PATHS matches what this script symlinks`, an exact match of the two path sets
- `shared paths: the worktree write-guard exempts every symlinked directory`, a subset test, since the guard carries further arms on unrelated rationale

Both live in the authority's own suite and are mutation-verified. Prose that keeps an enumeration is not machine-checked, and the tests do not pretend otherwise: their comments name those surfaces so the failure is also the checklist for the copies a machine cannot compare.

## Noted, not filed

**`pr-merge-audit-check.sh` resolves the gate's sha from its own process cwd. Investigated, deliberately excluded.**

Confirmed at `:197` (`sha=$(git rev-parse HEAD)`), at `:210` where `root` comes from a bare `--show-toplevel`, and in the comment at `:208-210` stating the assumption outright. Run `gh pr merge` for a worktree PR from the main checkout and every member digest is computed over the wrong tree, so the gate denies with a digest that looks wrong but is not.

Excluded for a reason beyond blast radius, which on its own would not have settled it: this PR already edits a `.claude/hooks/` file, so the members whose scope covers that surface rotate regardless.

The substantive reason is that a payload-cwd swap is not the correct fix, so it is not the "small, safe, clearly correct" case that would earn inclusion. The gate should evaluate the tree of **the PR being merged**, which is neither cwd. Any cwd is a proxy that happens to be right when the session sits where the protocol expects. Swapping one proxy for another leaves the same class of defect with a different trigger, and would look like a fix.

Two further findings support leaving it alone. The hook's cwd dependency is not confined to `:197`: line 135 tests for `repo-scope.sh` with a cwd-relative path, while the sibling lib loads deliberately resolve from `BASH_SOURCE` with a comment explaining why cwd must never be trusted there. The file already holds both models, and a partial retrofit deepens that split. And this is fail-closed gate machinery where a wrong answer either blocks every merge or admits an unaudited one, which is not the place for a heuristic swap landing beside an unrelated fix.

The honest options are: leave it (the protocol expects the session inside the worktree, and the failure is a loud deny, not a silent pass), or resolve the sha from the PR the command names, which the hook already parses for repo-scoping at lines 132-138. That is a design decision, not a repair.

## Quality Gate

Legitimately skipped. Confirmed with the documented grep from `wiki/decisions/Quality Gate.md`: no staged file matches `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.

Run regardless:

- `.gaia/tests/shell-lint.sh`: **passed** (163 scripts, 145 suites, hook array guard clean)
- `.gaia/tests/hooks/block-worktree-path-mismatch.bats`: 42/42
- `.gaia/scripts/tests/link-worktree.bats`: 18/18
- `.gaia/scripts/tests/create-worktree.bats`: 17/17

## CHANGELOG

Entry added under `## [Unreleased]`, `### Fixed`. A guard that stops going inert is adopter-visible, and it reaches existing adopters on their next `/update-gaia`.

## Pre-existing, untouched

`block-worktree-path-mismatch.sh:169` carries a comment line over the file's wrap width, from an earlier change. Mentioned rather than reflowed, per Surgical Changes.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.yExDN0XFT0

Issue #935, opened 2026-07-20
wiki commit-classify: type rules never match scoped commit subjects, so the WORTHY filter is inert


stevensacks:
## `wiki commit-classify` type rules never match this repo's scoped commit subjects

**Location:** `.gaia/cli/src/wiki/commit-classify.ts:251`

### Failure mode

Every type-based rule in the classifier matches an unscoped conventional-commit prefix:

```ts
subject.startsWith('feat:') ||
subject.startsWith('fix:') ||
subject.startsWith('refactor:');
```

with the same bare form at `:167` (`style:`), `:293` (`chore:`), `:299` (`docs:`), and `:303` (`test:`).

This repo writes scoped subjects exclusively. Across `fb3c596..b2ed25c7` (138 commits):

| Subject form | Count |
|---|---|
| scoped, e.g. `fix(hooks):` | 136 |
| unscoped, e.g. `fix:` | 0 |
| non-conforming | 2 |

`'fix(hooks): ...'.startsWith('fix:')` is `false`, so none of these rules can fire. Every commit falls through to Rule 9, which is deliberately fail-open (`Anything else: WORTHY (false positive better than false negative)`). The fail-open default is correct in itself; the defect is that it is reached unconditionally because nothing upstream of it can match.

The file is internally inconsistent on this point: it already handles scoped forms for `chore(adr):`, `chore(cli):`, `chore(deps):`, `chore(release):`, and `docs(decision):`, but not for the generic type rules.

### Measured impact

On the 2026-07-20 wiki sync over that range:

- 94 of 138 commits classified `no matching prefix, defer to human review`
- 137 of 138 marked `WORTHY`; the single `SKIP` was a `chore(deps):` version bump, unscoped by coincidence
- the only other rules that fired were path-based (43 via `touches wiki-heavy domain`)

Step 3 of the sync playbook exists to be a cheap deterministic filter ahead of Step 4's expensive per-commit diff reads. With 137 commits marked worthy it filtered nothing, and Step 4 mandated 137 LLM deep-reads. That run cost roughly $20 and 35 minutes, nearly all of it attributable to commits the classifier was designed to skip.

The defect is silent: the classifier reports a plausible reason string per commit and never signals that its rule table is inert.

### Suggested fix

[...]

PR #958, opened 2026-07-20
debt(cli): declare the conventional-commit grammar and vocabulary once

### LINKED ISSUE #936: wiki commit-classify: no detection when the rule table goes inert
## No detection when `commit-classify`'s rule table goes inert

**Location:** `.gaia/cli/src/wiki/commit-classify.ts:308`

### Failure mode

Rule 9 is the classifier's fail-open default:

```ts
return {
  reason: 'no matching prefix, defer to human review',
  suggestion: 'WORTHY',
};
```

Failing open is the right call for a single unrecognized commit. The problem is that nothing distinguishes *a few* commits landing here from *all* of them. When the rule table stops matching real subjects, every commit funnels into this return, the classifier keeps emitting a plausible per-commit reason string, and the run looks healthy while the deterministic first pass has silently stopped filtering.

Trigger: any divergence between the rule table and the commit subjects actually being written. Introducing a new type (`debt` was added at some point and no rule was ever written for it), changing the subject convention, or refactoring the matching logic all produce it.

Bad outcome: Step 3 of the sync playbook exists to be a cheap filter ahead of Step 4's expensive per-commit diff reads. When it goes inert, every commit is promoted to `WORTHY` and Step 4 deep-reads the entire range. There is no signal that anything is wrong.

This is not hypothetical. See #935: the type rules match unscoped prefixes (`fix:`) while the repo writes scoped subjects (`fix(hooks):`) exclusively, so 94 of 138 commits landed on this exact return and 137 of 138 were marked `WORTHY`. The 2026-07-20 sync consequently deep-read essentially the whole range at a cost of roughly $20 and 35 minutes. The defect had been latent for an unknown period and surfaced only because a large backlog made the cost visible. A smaller weekly sync would have absorbed it indefinitely without anyone noticing.

The existing test suite passes with the rule table fully unreachable, so tests do not cover this either.

### Suggested fix

Add a check that treats the *rate* of Rule 9 fallthrough as the health signal, rather than any individual commit's classification. Sample recent history, classify it, and flag when the share landing on `no matching prefix, defer to human review` exceeds a threshold.

This catches the whole class of divergence regardless of cause, which a commit-message linter does not: every one of those 94 commits was correctly formatted per the project's own convention. They were valid. The classifier simply could not read them, so no lint on the input side would have fired.

Open design questions the surrounding code does not settle, and which should be decided as part of this work:

- Where it lives: a CLI subcommand, a test asserting a floor on match rate over the last N commits, or a session-start or CI check.
- Whether it warns or fails.
- The threshold and sample size. Note that a legitimately unusual batch can spike the rate briefly, so the check should tolerate that without crying wolf.

Fixing #935 corrects today's divergence but restores no mechanism to catch the next one, which is why this is worth doing separately.

Handler: plan

### LINKED ISSUE #937: Commit-type vocabulary is undeclared, and squashed PR titles are unvalidated
## Commit-type vocabulary is undeclared, and squashed PR titles are unvalidated

**Location:** `.gaia/cli/src/wiki/commit-classify.ts:290` (`classifyCatchAll`, the de-facto vocabulary)

### Failure mode

There is no declared list of permitted commit types anywhere in the repo. The nearest thing is the set of prefixes `commit-classify.ts` happens to test for, which is not a specification and is not kept in sync with practice by anything.

Types in active use across `fb3c596..b2ed25c7` (138 commits), against whether the classifier has any rule for them:

| Type | Count | Rule exists |
|---|---|---|
| `fix` | 75 | yes |
| `feat` | 23 | yes |
| `docs` | 12 | yes |
| `test` | 9 | yes |
| `debt` | 7 | **no** |
| `refactor` | 6 | yes |
| `chore` | 2 | yes |
| `ci` | 1 | **no** |
| `build` | 1 | **no** |

Trigger: someone introduces a new commit type. Nothing registers it, nothing warns, and the classifier silently treats every commit of that type as unrecognized and promotes it to `WORTHY` via the Rule 9 fail-open. `debt` is the proven case, in steady use with no rule ever written for it.

Note this is independent of #935. Correcting the scoped-prefix matching there still leaves `debt`, `ci`, and `build` with no disposition, because the gap is a missing vocabulary rather than a broken matcher.

### Secondary: PR titles are the load-bearing artifact and go unchecked

Because PRs are squash-merged, the PR title becomes the commit subject on `main` and is therefore what the classifier reads. Branch commits are discarded. Both non-conforming subjects in the range are PR titles:

```
Harden the Code Audit Team merge gate (#793)
Remove the adaptive mentorship layer and the event pipeline that feeds it (SPEC-038) (#763)
```

At 2 of 138 the direct cost is small, and this is explicitly not a call for commit-level linting: individual commits are 136 of 138 conforming and a per-commit gate would add friction while defending something already observed. The title is the only point where validation is both load-bearing and cheap.

### Suggested fix

Declare the permitted type vocabulary in one place, and have both consumers read it rather than each carrying its own copy:

1. `commit-classify.ts` derives its type matching from the declared list, so a type without a disposition is a visible gap rather than a silent fallthrough.
2. A PR-title check validates against the same list on `pull_request` (`opened`, `edited`, `synchronize`). No workflow covers this today; the existing workflows under `.github/workflows/` are tests, audit, lint, release, and distribution.

Decide dispositions for `debt`, `ci`, and `build` as part of declaring the list.

Ordering note: this pairs with the drift detector filed alongside it. This issue keeps the vocabulary honest at the input; that one detects when the classifier stops matching regardless of cause. Neither subsumes the other, and both are cheap relative to a single degraded sync run.

Handler: plan

### LINKED ISSUE #945: wiki commit-classify: rules 6/7 discriminate on app/**, unreachable in GAIA's own repo
## `commit-classify` rules 6/7 discriminate on `app/**`, which GAIA's own repo never touches

**Location:** `.gaia/cli/src/wiki/commit-classify.ts:346`

### Failure mode

#935 fixed the type rules so they match this repo's scoped commit subjects, and commits landing on the rule-9 fail-open dropped from 94 of 138 to 1 of 139. But the sync's first pass still barely filters: `WORTHY` only fell from 137 to 122.

The reason is a second, independent inertness with a different root cause. Rules 6/7 discriminate a `feat`/`fix`/`refactor`/`debt` commit by **path**:

```ts
if (touchesOnlyTests(files))            // SKIP
if (onlyInventory && !ARCH_BODY...)     // SKIP  (app/components|hooks|services|pages)
if (touchesAppNonTest(files))           // WORTHY (app/** non-test)
return {reason: 'feat/fix/refactor/debt: defer to human review', ...}  // WORTHY
```

Every one of those predicates is scoped to `app/**`. GAIA's own product source is `.gaia/` and `.claude/`; `app/` is the React template it ships to adopters. Measured over `fb3c596..HEAD` (139 commits), **zero** commits touch `app/` at all. So all three discriminating branches are unreachable and every source commit falls to the final `defer to human review` WORTHY.

That single reason accounts for **76 of the 139** commits, by far the largest remaining bucket:

| reason | count |
|---|---|
| `feat/fix/refactor/debt: defer to human review` (WORTHY) | 76 |
| `touches wiki-heavy domain` (WORTHY) | 44 |
| everything else | 19 |

Same consequence as #935: Step 3 exists as a cheap deterministic filter ahead of Step 4's expensive per-commit diff reads, and for GAIA's own history it promotes essentially every source commit to a deep read.

### Why this is not #935, #936, or #937

- **#935** (merged) was the *type* vocabulary failing to match scoped subjects. This is the *path* vocabulary failing to match the repo's layout. Fixing one does nothing for the other.
- **#936** proposes a detector for a rule table going inert. A fallthrough-rate check keyed on rule 9 would **not** fire here: these commits do not reach rule 9, they get a plausible, specific reason string from rule 7. If #936 is built, this is a concrete case worth making sure it catches.
- **#937** declares the *type* vocabulary. Same axis as #935, not this one.

### Open design questions

The fix is not simply "add `.gaia/` to the prefix lists", because the classifier is a **shipped adopter surface**:

- For an adopter React app, `app/**` is correct and should stay.
- Adopters also have a `.gaia/` (the harness). Making `.gaia/**` source-bearing globally would promote every adopter's harness churn to WORTHY, which is wrong for them.
- So the path vocabulary probably has to become repo-aware or configurable rather than a hardcoded list, which is a contract decision on a shipped surface.

Decide the mechanism (config key, convention, or derive from the manifest) as part of this work.

### Suggested fix

Make the source-bearing and inventory path sets a declared, repo-configurable input rather than `app/**` literals, defaulting to today's values so adopter behavior is unchanged, and give GAIA's own clone a set that matches where its source actually lives.

Handler: plan

### LINKED ISSUE #946: Three private copies of the conventional-commit grammar across the CLI, already drifted
## Three private copies of the conventional-commit grammar, already drifted

**Location:** `.gaia/cli/src/release/bump.ts:105` (plus `.gaia/cli/src/release/changelog.ts:135` and `.gaia/cli/src/wiki/commit-classify.ts:180`)

### Failure mode

The same conventional-commit header grammar (type, optional `(scope)`, optional `!`, `:`) is parsed by three file-private regexes in two directories:

```ts
// release/bump.ts:105
/^(?<type>[a-z]+)(?:\([^)]*\))?(?<bang>!?):/u
// release/changelog.ts:135
/^(?<type>[a-z]+)(?:\([^)]*\))?!?:(?<rest>.*)$/u
// wiki/commit-classify.ts:180
/^(?<type>[a-z]+)(?:\((?<scope>[^)]*)\))?(?<breaking>!?):/u
```

None is exported, so nothing can reuse another, and they have **already drifted**: only `commit-classify` captures the scope by name, `changelog` names neither scope nor bang, and the three disagree on what they expose. This is not hypothetical drift risk, it is observed divergence.

It also already caused a real defect. `commit-classify` originally consulted the `!` marker for `feat` only, while `bump.ts:122` has always treated `!` on *any* type as a major bump. That inconsistency was caught in review during #935 (a `chore(cli)!:` would have been silently SKIPped as tooling-internal) and fixed in `commit-classify` alone; the underlying reason the two could disagree at all is that they share no parser.

Bad outcome: the next grammar change (multi-segment scopes like `feat(cli/wiki):`, digits in a type, a different bang position) has to be found and re-verified in three regexes and three test suites, and any one that is missed fails silently, since every consumer here is fail-open by design.

### Suggested fix

Extract a single `parseConventionalCommitHeader` returning `{type, scope, breaking, rest}` into a shared module (e.g. `.gaia/cli/src/util/conventional-commit.ts`, sibling to the existing `.gaia/cli/src/util/atomic-write.ts`) and have `bump.ts`, `changelog.ts`, and `commit-classify.ts` all consume it. Keep each caller's own disposition logic; only the grammar is shared.

Surfaced twice independently during #935's review (a reuse pass and the maintainer-node audit), and deliberately left out of that PR because rewiring two release-critical modules is well outside a classifier matcher fix.

Handler: plan

### LINKED ISSUE #947: release bump: PATCH_TYPES omits debt and build, so such commits contribute no version bump
## `bump.ts` PATCH_TYPES omits commit types the repo actively writes

**Location:** `.gaia/cli/src/release/bump.ts:93`

### Failure mode

`classifyCommit` maps a commit subject to a bump kind:

```ts
if (match.groups?.bang === '!') return 'major';
if (match.groups?.type === 'feat') return 'minor';
if (PATCH_TYPES.has(match.groups?.type ?? '')) return 'patch';
return null;
```

`PATCH_TYPES` omits `build`, `debt`, and `wiki`, all of which are in active use:

| Type | Occurrences in recent history | In `PATCH_TYPES` |
|---|---|---|
| `debt` | 7 | no |
| `build` | 1-2 | no |
| `wiki` | 5 | no |
| `ci` | 1-2 | yes |

A commit of an omitted type returns `null` and contributes nothing to the aggregate bump. A release window containing **only** such commits aggregates to no bump at all, and `bump` reports `no conventional-commit prefixes found` even though every commit in the range was correctly formatted.

This is reachable rather than theoretical: `/gaia-debt` emits `debt(...)` commits, and a maintenance window of debt drains plus a rebundle (`build(cli):`) is a realistic release range.

Note `wiki:` is arguably correct to omit, since sync commits are the tool's own bookkeeping and probably should not force a release. `debt` and `build` are the substantive gap.

### Suggested fix

Decide which of `debt`, `build`, and `wiki` are patch-worthy and add those to `PATCH_TYPES`. Worth deciding at the same time whether an unrecognized-but-well-formed type should be a loud failure rather than a silent `null`, since the current behavior makes an omission invisible until a release produces no bump.

Surfaced by the maintainer-node audit during #935; `bump.ts` was unchanged by that PR, so it was left out of scope.

Handler: prompt

### LINKED ISSUE #948: commit-classify docstring cites a nonexistent README contract and a wrong sync.md path
## `commit-classify` docstring points at a contract document that does not exist

**Location:** `.gaia/cli/src/wiki/commit-classify.ts:9`

### Failure mode

The module docstring names its authoritative sources:

```
 * Rule precedence (first match wins). Mirrors the rules in
 * `wiki/sync.md` Step 3 and the README "Wiki primitives JSON schemas"
 * contract:
```

Two problems:

1. **The README section does not exist.** There is no "Wiki primitives JSON schemas" section anywhere in the repo, and `.gaia/cli/` contains no `.md` file at all. `git log -S` dates the reference to `433abf00` (#97), so it has been stale for a long time.
2. **The `wiki/sync.md` path is wrong.** The file lives at `.claude/skills/gaia/references/wiki/sync.md`, not `wiki/sync.md`.

No live failure mode; the cost is that a reader looking for the authoritative WORTHY/SKIP contract is sent to two places, neither of which resolves. That matters more than usual here because the rules are subtle and two open issues (#936, #937) are about keeping them honest.

Worth noting the docstring's *intent* is also now inverted: `.claude/skills/gaia/references/wiki/sync.md:50` explicitly says "Trust the CLI's classification, do not re-derive WORTHY/SKIP rules in prose." So the code is the contract and the doc mirrors it, not the other way round, which is the opposite of what "Mirrors the rules in ..." implies.

### Suggested fix

Drop the README clause, correct the `sync.md` path to `.claude/skills/gaia/references/wiki/sync.md`, and reword so the docstring reads as the source of truth that `sync.md` defers to.

Surfaced by the maintainer-node audit during #935; pre-existing and outside that PR's changed lines, so left untouched per surgical-changes.

Handler: prompt

### LINKED ISSUE #949: Root tsconfig claims .gaia/cli, producing phantom type errors across the whole maintainer CLI in-editor
## Root tsconfig pulls `.gaia/cli/**` into the app project, so every CLI file shows phantom type errors in the editor

**Location:** `tsconfig.json:2` (the `include` array)

### Failure mode

The root `tsconfig.json` includes `**/*.ts` and excludes only `node_modules`, `build`, `coverage`, and `storybook-static`:

```json
"include": ["**/*.ts", "**/*.tsx", ...],
"exclude": ["node_modules", "build", "coverage", "storybook-static"]
```

`.gaia/cli/` is a **separate pnpm workspace** with its own `tsconfig.json`, its own `@types/node`, and its own dependency tree. Because the root config claims its files anyway, an editor language server resolves them against the app's project, where none of that is configured. Opening any file under `.gaia/cli/src/` produces a wall of phantom diagnostics:

```
Cannot find name 'process'. Do you need to install type definitions for node?
Cannot find module './commit-classify.js' or its corresponding type declarations.
Cannot find name 'node:child_process' ...
Parameter 'file' implicitly has an 'any' type.
```

None are real. `pnpm -C .gaia/cli typecheck` exits 0 on the same files, and CI's `Vitest (.gaia/cli)` job passes.

Bad outcome: the editor's diagnostics are untrustworthy for the entire maintainer CLI, which is a large and actively edited surface. A real error introduced in `.gaia/cli/src/` is indistinguishable from the ~20 phantom ones already on screen, so the fastest feedback loop a maintainer has is effectively disabled there and regressions surface only at the workspace typecheck or in CI.

The command-line `pnpm typecheck` is unaffected in practice, so this is specifically an editor/language-server problem, which is why it has gone unnoticed.

### Suggested fix

Exclude `.gaia` from the root `tsconfig.json` so the CLI workspace is resolved only by its own config. Verify afterwards that `pnpm typecheck`, `pnpm lint`, and the Vitest projects still see exactly the file sets they see today, since the root `include` is also what feeds them.

Handler: plan

stevensacks:
Seven tech-debt issues, one root cause: the conventional-commit grammar and its type vocabulary were never declared anywhere, so every consumer inferred them privately and each inference drifted on its own axis.

Closes #946
Closes #947
Closes #948
Closes #937
Closes #945
Closes #936
Closes #949

## The decision

**Declare the grammar and the vocabulary once; keep each consumer's disposition local.**

`.gaia/cli/src/util/conventional-commit.ts` exports `parseConventionalCommitHeader` and `COMMIT_TYPES`. The three consumers each keep their own table, keyed by `CommitType`:

| Consumer | Table | Disposition it owns |
|---|---|---|
| `release/bump.ts` | `TYPE_TO_BUMP` | semver bump |
| `release/changelog.ts` | `TYPE_TO_SECTION` | Keep-a-Changelog section |
| `wiki/commit-classify.ts` | `PATH_DISCRIMINATED_TYPES` | whether rules 6/7 judge it by path |

Because all three are `Record<CommitType, ...>`, adding a type to the vocabulary is a compile error in all three until each decides, and removing one is an excess-property error. That is the whole mechanism: the previous failure mode was a type in steady use that no consumer had an opinion about, and nothing anywhere reported it.

**Rejected: one central disposition table.** The three legitimately disagree (`docs` is patch / Changed / SKIP; `wiki` is null / null / suppressible). Centralizing would put semver policy, changelog policy, and wiki-worthiness policy in one module that no one owns, and would leak the next consumer's concerns into the same place.

**Rejected: a per-commit message lint.** Branch commits are 136 of 138 conforming. Squash-merge means the PR title is the subject that lands, so that is the only point where validation is both load-bearing and cheap.

## Per issue

**#949 — premise is wrong, no change made.** The root `tsconfig.json` does **not** claim `.gaia/cli/**`. TypeScript's wildcard `include` patterns skip dot-prefixed directories, so `.gaia/` is never matched. Verified directly:

```
$ npx tsc --noEmit --listFiles | grep -c "\.gaia/cli"
0
```

The real cause of the phantom diagnostics is an **uninstalled `.gaia/cli` workspace**. Reproduced in a fresh worktree: with `.gaia/cli/node_modules` absent, `tsc -p .gaia/cli/tsconfig.json` emits exactly the issue's symptoms (`Cannot find module 'js-yaml'`, and `Cannot find name 'process'` when root `node_modules` is also absent). After `pnpm -C .gaia/cli install`, they all vanish. Adding `.gaia/cli` to the root `exclude` would have been a no-op that looked like a fix. The actual remedy is `pnpm -C .gaia/cli install`.

**#946 — done.** Three drifted regexes replaced by one parser. `rest` is derived by slicing the matched header rather than extending the pattern with `(?<rest>.*)$`, so a subject carrying a newline still parses.

**#947 — done.** `debt` and `build` now contribute a patch bump. `wiki` stays release-neutral deliberately: a sync commit should never be the reason a version ships. An undeclared-but-well-formed type still contributes nothing rather than failing loudly, since `ci-check-subject` now catches it at the input instead.

**#937 — done.** Vocabulary declared; `gaia ci-check-subject` validates a subject against it; `pr-title.yml` runs it on `opened`/`edited`/`reopened`. No `synchronize`, since a push cannot change a title.

**#945 — done, with an honest result.** Path sets are now read from `gaia.wikiClassify` in `package.json`, defaulting to today's `app/**` values. Measured over the same 139 commits the issue cites:

| | before | after |
|---|---|---|
| fail-open share | 77/139 (55%) | 16/139 (12%) |
| WORTHY share | 86% | 85% |

**The filtering barely moves, and that is the real finding.** Rules 6/7's `source-bearing → WORTHY` branch and its fallthrough both produce WORTHY, so the only filtering value in those rules was ever in their two SKIP branches. GAIA's own source genuinely is what its wiki documents, so a high WORTHY rate here is correct rather than a fault. What actually changed is that the classifier now *knows* what it is looking at instead of guessing, which is what makes the health signal trustworthy. Cost control for a large window remains sync.md Step 2's drift cap, not this filter.

`package.json` over a `.gaia/` file because `/update-gaia` never touches keys outside its managed sections, so adopter tuning survives an update; `.gaia/audit-ci.yml` is manifest class `shared` and gets a field-aware merge that can emit a conflict patch.

**#936 — done.** Health block on every run: `evaluated`, `deferred`, `deferral_rate`, `worthy_rate`, `threshold`, `inert`, plus a stderr warning when inert. Two calibration points that mattered:

- **Both fail-open sites count**, not just rule 9. Rule 7's tail emits a specific-sounding reason (`defer to human review`) and is exactly how #945 hid. A rule-9-only check would not have fired on it, as #945 itself predicted.
- **Threshold is 0.4, not 0.6.** The #935 type failure sat at 68% and the #945 path failure at 55%. My first draft used 0.6, which would have caught the first and missed the second. Measured healthy state is 12%, so 0.4 separates cleanly.

`worthy_rate` is reported alongside deferral so that configuring the path vocabulary cannot quietly convert "I could not tell" into "definitely read this" while the deep-read cost stays identical.

**#948 — done.** Dropped the nonexistent README clause, and reworded so the module reads as the WORTHY/SKIP contract that `sync.md` defers to, rather than a mirror of it.

## Found while testing

- **`perf` had no classifier rule.** The new every-declared-type-reaches-a-rule guard failed on `perf` the first time it ran. It was in `PATCH_TYPES` and `TYPE_TO_SECTION` but had no wiki disposition, so it would have landed on the rule-9 fail-open exactly as `debt` did. Added to the path-discriminated set.
- **The guard's first version was too weak to catch that class.** It substring-matched the rule-9 reason string and used a neutral fixture path, so the five path-discriminated types passed by landing on rule 7's *deferred* tail, the very outcome it exists to reject. Rewritten to assert `health.deferred === 0` against a source-bearing fixture. Mutation-checked: flipping `debt` out of the table now fails the guard.

## Noted, not filed

Per the batch instruction, further defects in this machinery, recorded rather than filed:

1. **`.gaia/cli/src/release/preflight.ts:336`** does `driftSubjects.every((subject) => subject.startsWith('wiki:'))`, and its own comment claims it uses "the same marker `gaia wiki commit-classify` uses". It no longer does: the classifier matches through the parsed header, so `wiki(sync):` and `wiki!:` are self-referential to it but are *not* benign drift to preflight. This is the same bug class this PR exists to kill, and it is the only remaining instance in the tree. Left alone as out-of-diff per surgical-changes; a one-line fix plus an import.
2. **The health metric only sees a rule dying *into a fail-open default*.** A rule that dies into another *discriminating* rule stays invisible: renaming the `chore(deps):` convention would strand rule 5 while those commits landed on rule 8's `chore` arm, keeping deferral at zero. Catching that needs per-rule counts against a persisted cross-sync baseline. Recorded in the module header rather than built.
3. **The "deliberately not in `required_status_checks`, that is an ask-first cutover" rationale is now written three times** (`pr-title.yml`, `verify-required-checks.yml`, `read-audit-ci-config.sh`). The policy should live once with the workflows pointing at it.
4. **`commitDetails` in `wiki/util/git.ts` spawns one `git diff-tree` per commit.** Pre-existing N+1 subprocess, and by far the dominant cost of a classify run.

## Quality Gate

| Step | Result |
| --- | --- |
| Simplify | Pass, 2 review agents, 8 findings applied |
| Localization | N/A, no user-facing app strings |
| Type checking | Pass, root + `.gaia/cli` |
| Linting | Pass, `pnpm lint` + `pnpm lint:cli`, zero warnings |
| Unit tests | Pass, 155 root + 1659 CLI |
| E2E tests | Pass, 7 passed 1 skipped |
| Dev server | Pass, via the Playwright web server boot |
| Build | Pass |

Bundles verified reproducible from src (`cmp` against a fresh `pnpm bundle`).


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.P5iXMKsoWO

PR #962, opened 2026-07-21
feat(release): guard bare-prose mentions of excluded wiki titles (#911)

### LINKED ISSUE #911: No guard for a release-excluded wiki page title referenced as bare prose in a shipped page
**Location:** `.gaia/cli/src/release/scrub.ts:467` (the derived `wikilink-to-excluded` check).

## Failure mode

A release-excluded wiki page's **title**, referenced as bare prose in a shipped page, reaches adopters as a dangling reference to a page they never receive. No current check sees it:

- `wikilink-to-excluded` derives the excluded-slug set correctly but only matches `[[wikilinks]]`.
- The scrub's `maintainer-paths` check and the health audit's Bucket B match **path** prefixes, not titles.
- `gaia wiki dead-paths` matches backticked filesystem paths only.

A bare title mention is invisible to all three. Concretely, two shipped pages carried the sentence-level phrase `Forensics Triage Workflow` (a page release-excluded at `.gaia/release-exclude:57`) with no wikilink and no path:

- `wiki/decisions/Quality Gate.md`
- `wiki/concepts/Code Review Audit CI.md`

Both instances are already fixed (wrapped in `gaia:maintainer-only` markers in #908), so there is no active leak. The **class** has no guard, so the next one ships silently.

## Why this is not a one-liner

The obvious implementation, "derive the excluded page-title set at scan time and flag bare mentions", is a non-starter as stated. Of the 47 release-excluded `wiki/**` pages, the titles split three ways:

| Group | Count | Examples | Scannable |
|---|---|---|---|
| Distinctive proper nouns | ~6 | `Forensics Triage Workflow`, `Bundle-time Scrub`, `Folding Shell Scripts into the CLI Binary`, `CLI-Binary-Split` | Yes, this is the real leak surface |
| Dated generated artifacts | ~35 | `lint-report-2026-07-08`, `consolidate-report-2026-07-09`, `staleness-audit-2026-06-24` | Structurally unambiguous, never named in prose |
| Generic words | ~4 | `GAIA`, `dashboard`, `Steven Sacks`, `Release Workflow` | **No.** Scanning for `GAIA` flags every shipped page |

Flagging on the full derived set would fire on essentially every page. That reproduces the cry-wolf failure mode fixed in #910 (a scan that reports known-false anomalies every cycle trains readers to discount it), but far louder.

## Suggested design

Treat the leak-sensitive title set as **explicitly declared**, not fully derived, mirroring how `PROSE_PATH_ALLOWLIST` handles per-entry false positives in `release/runtime-deps.ts`:

1. Add an opt-in title list to the `wikilink-to-excluded` entry in `.gaia/release-scrub.yml` (or a sibling check id), each entry carrying a one-line justification.
2. Scan the **post-strip staging tree**, so text inside `gaia:maintainer-only` markers is already gone and needs no special-casing.
3. Skip backticked spans and fenced code blocks; those are path references already owned by other checks.
4. Fail the release build on a hit, matching the existing leak-check severity.

Open design questions for the SPEC: whether the declared list should be derivable-with-an-opt-out instead of opt-in (a heuristic such as "multi-word AND title-case AND not a single common noun" might cover the ~6 real cases automatically, at the cost of a fuzzier rule); and whether the check belongs in `release scrub` or as a `gaia wiki` primitive that the audit can also call directly.

Handler: spec

---

Found by the false-clean challenger during a `/health-audit` run. Analysis and the two fixed instances come from #908; the cry-wolf precedent is #910.

stevensacks:
## What

Adds a third derived leak-check, `excluded-titles`, inside `gaia-maintainer release scrub`. It fails the release build when a shipped `wiki/**` page names a release-excluded wiki page's TITLE in **bare prose**: a Title-Case reference that is not a `[[wikilink]]`, a backticked path, a fenced code block, or a stripped `gaia:maintainer-only` marker block. Such a reference would otherwise reach an adopter as a dangling pointer to a page their clone never received.

This closes the last dangling-reference gap the three existing derived checks each miss: `wikilink-to-excluded` matches only `[[wikilinks]]`, `maintainer-paths` matches path prefixes, and `gaia wiki dead-paths` matches backticked filesystem paths. A bare Title-Case mention is invisible to all three.

## How

- **Derivation (case-preserving, page-basenames only):** the checked-title set is derived at scan time from the `.md` page basenames named in `.gaia/release-exclude` (read from the source repo, which excludes itself), case-preserved, contributing every `.md` exclude's basename and every `.md` page beneath a bare-directory exclude, but **never** the bare-directory basename itself. So the set cannot drift from the exclude list; a newly excluded distinctive page is covered with no config edit.
- **Matcher:** its own full-line-stream walk with per-file fenced-code-block state, span-level stripping of `[[wikilink]]` and inline-backtick spans, and case-sensitive whole-token matching (alphanumerics and hyphens are token-internal, so `Bundle-time Scrub` does not fire inside `Bundle-time Scrubbing`).
- **Opt-out:** a config-declared `title-opt-out` (seeded with the four known generics `GAIA`, `dashboard`, `Steven Sacks`, `Release Workflow`, each `#`-justified) is subtracted by case-sensitive equality. Mis-nesting `title-opt-out` onto any other check fails config load loudly (exit 2) rather than silently dropping it.

## Scope

Maintainer-only framework work under `.gaia/cli/src/release/`. The adopter `gaia` binary never imports release code, so no adopter-observable behavior changes; the guard runs only on the GAIA maintainer's own release cut.

## Verification

- CLI-package gate green: typecheck clean, lint clean (`--max-warnings=0`), 1698 tests passing (16 new `excluded-titles` acceptance tests).
- Maintainer binary rebundled (`.gaia/cli/gaia-maintainer` only).
- Introduction-clean gate: stage-and-scrub against the real tree reports **zero `excluded-titles` leaks**.

Closes #911


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.gJTYq8yIsm

PR #965, opened 2026-07-21
debt(release): scan past an unbalanced fence in the excluded-titles leak-check

### LINKED ISSUE #963: excluded-titles leak-check: unbalanced fence leaves fence state stuck (fail-open)
**Location:** `.gaia/cli/src/release/scrub.ts:999` (`findTitleLeaksInFile`)

**Failure mode.** The `excluded-titles` release leak-check tracks fenced-code-block state per file by toggling `insideFence` on every fence delimiter (line 999), with no handling for an unbalanced fence. A shipped `wiki/**` page containing an odd number of ``` ``` ``` / `~~~` fence delimiters — an opened-but-never-closed fence, or otherwise malformed markdown — leaves `insideFence` stuck `true` after the final unmatched delimiter. Every line after it is then skipped (the match branch at line 1000 requires `!insideFence`), so a genuine bare-prose Title-Case mention of a release-excluded wiki page that appears *after* the unbalanced fence goes unreported. This is fail-open: the release build passes when it should fail, and the dangling reference ships to adopters — exactly the leak the check exists to catch.

Triggering input: a `wiki/**` staging page with a lone unmatched ``` ``` ``` (or `~~~`) delimiter, followed later in the same file by a bare mention of an excluded page title (for example `Forensics Triage Workflow`).

Discovered as an advisory, non-blocking finding by the pre-merge `code-audit-maintainer-node` audit on PR #962 (the change that introduced the check). The current toggle behavior matches that PR's frozen design contract ("a fence delimiter always toggles state") and requires malformed markdown to trigger, so it was recorded as a follow-up rather than fixed in place. No live failure mode exists on the current tree.

**Suggested fix.** Make the fence tracker resilient to an unbalanced fence rather than silently swallowing the tail of the file. The direction is a design choice the surrounding code and the sibling checks settle: either treat the fence as implicitly closed at EOF for matching purposes (scan the tail anyway), or surface the unbalanced fence as its own diagnostic. An unclosed `gaia:maintainer-only` marker already fails the release build, so aligning an unbalanced code fence with that fail-closed posture is a reasonable model. Pair the fix with a `scrub.test.ts` case: a page with an odd fence-delimiter count and a bare excluded-title mention after it must be flagged.

**Handler:** prompt

stevensacks:
## What

The `excluded-titles` release leak-check (`gaia-maintainer release scrub`) tracked fenced-code-block state with a per-line `insideFence` toggle. An **odd (unbalanced) fence-delimiter count** left the toggle stuck `true` through end-of-file, so every line after the unclosed fence was silently skipped. A bare Title-Case mention of a release-excluded wiki page appearing *after* such a fence went unreported.

This is **fail-open**: the release build passes when it should fail, and a dangling reference to a page the adopter's clone never received could ship. Maintainer-only, release-excluded tooling, so no adopter clone is affected today, and the introducing check (#911) is still unreleased.

## Fix

Derive fenced-code-block spans from delimiter **pairs** instead of a running toggle. Delimiters pair left-to-right; a trailing lone (odd) delimiter closes nothing, so it and every line after it stay scannable, restoring the fail-closed posture the sibling checks hold. Balanced fences behave exactly as before.

## Test

Adds a `scrub.test.ts` case: a page with a closed fence pair followed by an **unclosed** fence and a bare excluded-title mention after it. The mention must be flagged (it was swallowed before), while the closed pair above it is still skipped. Fails on the old toggle, passes on the fix.

Closes #963

### DIFF
--- a/.gaia/cli/gaia-maintainer
+++ b/.gaia/cli/gaia-maintainer
@@ -23060,11 +23060,19 @@ var isFenceDelimiter = (line) => {
 var stripSkippedSpans = (line) => line.replaceAll(WIKILINK_SPAN, "").replaceAll(INLINE_CODE_SPAN, "");
 var findTitleLeaksInFile = (file2, matchers, lineAllowlist) => {
   const leaks = [];
-  let insideFence = false;
-  for (const [index, line] of file2.content.split("\n").entries()) {
-    if (isFenceDelimiter(line)) {
-      insideFence = !insideFence;
-    } else if (!insideFence && !lineAllowlist.some((rx) => rx.test(line))) {
+  const lines = file2.content.split("\n");
+  const fenceIndices = lines.flatMap(
+    (line, index) => isFenceDelimiter(line) ? [index] : []
+  );
+  const insideClosedFence = /* @__PURE__ */ new Set();
+  for (let pair = 0; pair + 1 < fenceIndices.length; pair += 2) {
+    for (let index = fenceIndices[pair]; index <= fenceIndices[pair + 1]; index += 1) {
+      insideClosedFence.add(index);
+    }
+  }
+  for (const [index, line] of lines.entries()) {
+    const scannable = !insideClosedFence.has(index) && !lineAllowlist.some((rx) => rx.test(line));
+    if (scannable) {
       const residue = stripSkippedSpans(line);
       for (const { regex, title } of matchers) {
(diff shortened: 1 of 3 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.AkU0xllP5F

PR #969, opened 2026-07-21
debt(audit): retire superseded refusals and detect lost member reports (#931, #941)

### LINKED ISSUE #931: Audit gate: a superseded .refused marker permanently blocks merge, and the documented remedy cannot clear it
`.claude/hooks/lib/audit-clearance.sh:98` (`clearance_member_refused`)

## Failure mode

A Code Audit Team member that writes a `.refused` marker and later writes an **earned `.ok` for the same content digest** leaves a permanent merge blocker that no documented path can clear.

`clearance_member_refused()` decides refusal purely on the `.refused` file existing with a matching digest, member, and `provenance`. It never compares timestamps and never looks for a same-digest `.ok`. `pr-merge-audit-check.sh` then gives that result absolute priority, stating "a refusal always takes precedence over any earned marker for the same content". Nothing reconciles the pair afterwards: `local-janitor.sh` reaps `.ok`/`.refused` only by age and retention, and `post-audit-status.sh` never removes a superseded refusal.

The result is a dead end, because the hook's own remedy text is unreachable. It instructs the operator to "re-spawn the member's agent to address the finding", but a re-spawn against unchanged content produces another `.ok` while the `.refused` persists, so the gate stays shut no matter how many times the documented remedy is followed. The only exits are mutating the content to shift the digest, or deleting the refusal by hand.

Observed live on PR #930: `code-audit-maintainer-shell` wrote `.refused` for digest `d5bc5718…` at 08:10Z, then an earned `.ok` for that identical digest at 08:23Z. The merge gate stayed shut, and the only resolution was a human manually `rm`-ing the refusal, which is precisely the bypass the design forbids an agent from performing.

## Suggested fix

Settle the supersession semantics first, then implement. The design tension is the substance of the work and must not be skipped: refusal-precedence exists specifically to stop someone re-running an auditor until it passes, so "newest marker wins" is **not** an acceptable fix, it converts the anti-gaming control into a no-op.

A fix has to distinguish legitimate supersession from gaming. Options worth weighing, none obviously correct:

- Have a member delete its own `.refused` when it writes an `.ok` for the same digest, making the reversal an explicit act by the same author rather than an inference by the gate.
- Record a resolution pointer in the `.ok` (naming the refusal it supersedes and why), so supersession is auditable rather than silent.
- Keep strict refusal-precedence and instead fix the hook's remedy text, which currently promises a re-spawn will clear a state a re-spawn provably cannot clear.
- Require a human-attested override artifact, making the current manual `rm` a first-class, recorded action instead of an undocumented bypass.

Whichever is chosen, the gate should stop advertising a remedy that cannot work.

Handler: spec

### LINKED ISSUE #941: Code Audit Team member report can be lost even under synchronous dispatch, when the return channel isn't the orchestrator-visible one
`.claude/agents/code-audit-maintainer-shell.md:107` (the "## Output Format" section). The three sibling Code Audit Team member specs share the identical reporting shape at the equivalent section: `code-audit-github-workflows.md`, `code-audit-maintainer-node.md`, `code-audit-maintainer-prose.md`. `code-audit-frontend`, the default member, is not in scope here: it already routes its report through a different, ledger-based return contract for a related reason (see its "Output Format" section, "Return contract").

## Failure mode

Observed twice in one session: a dispatched Code Audit Team member completed its audit, wrote a valid earned clearance marker, and emitted its full findings report as plain assistant text at the end of its turn. The report never reached the dispatching orchestrator, which was left holding a green marker gate and zero visible findings.

`wiki/concepts/PR Merge Workflow.md` documents a no-op classifier (`.gaia/scripts/audit-noop-detect.sh --shape audit-team-member`, see "No-op detection and retry for each dispatched member") that reads a dispatched member's returned text to tell a real audit apart from a silent no-op. A member whose report is lost in transit is indistinguishable from one that no-op'd, except that its marker exists on disk, and the marker's presence is the only thing that keeps it from being classified a no-op and retried.

The practical consequence: an orchestrator can reach the merge gate believing "earned" means no unaddressed Critical or Important finding, while never actually seeing the report, including any Suggestions the workflow's clean-pass contract says the operator must resolve or acknowledge before merging.

Related: #918 diagnosed and fixed the case where a member's report was lost because the dispatch ran in the background (the `Agent` tool's default). The wiki page now mandates `run_in_background: false` for exactly that reason, reasoning that "a backgrounded member routes no text back... A synchronous dispatch is what makes the returned-text contract hold." This finding is a distinct failure mode surfacing after that fix: dispatch was synchronous, the member did the work and wrote its marker, and the report was still lost. The plain-text-output channel itself dropped it before it reached the orchestrator.

## Suggested fix

Starting point, not a settled fix. The wiki's synchronous-dispatch reasoning holds for a plain foreground `Agent` call: a synchronous tool call's return is visible to its caller automatically. It does not hold for every dispatch topology. Under a team-based dispatch, where a member runs as an addressable teammate rather than a plain subagent call, a teammate's final assistant-text turn is not automatically visible to whichever agent dispatched it; only an explicit message back to that agent is. The member spec currently gives no guidance on this: it defines the report's content (this "Output Format" section) but never says which channel carries it back to the orchestrator.

Confirm which dispatch topology (or topologies) actually reach this spec before prescribing a single fix. The fix likely has to name the return channel explicitly for whichever topology is in play, and, once settled, should apply uniformly to all four sibling members sharing this reporting shape rather than to this one alone.

Handler: plan

stevensacks:
Batch F of the tech-debt drain, and the last of it. Closes the two remaining open `tech-debt` issues, both `severity:important` + `difficulty:hard`.

Batched deliberately: both fixes touch shared gate machinery, which rotates **every** member's digest repo-wide, so batching pays that audit trip once instead of twice.

## #931 — a refusal nothing could clear

The earned and refused clearance families live at **different filenames**, so an earned write leaves a same-digest refusal on disk, and the gate checks the refusal family first and treats it as absolute. The one legitimate refused→earned path on identical content is an operator acknowledging an unaddressed Important **with a stated reason** — which moves no bytes, so the digest never rotates and the gate stays shut permanently. The remedy the hook itself printed ("re-spawn the member's agent to address the finding") provably could not clear it, leaving a hand `rm` as the only exit — precisely the bypass the design forbids.

**Settled design: member-authored supersede + recorded reason.** The shared writer takes `--supersede-refusal <reason>`; on an earned write it records a `supersedes` block in the marker body and removes that member's *own* sibling refusal.

- **Ordering is load-bearing.** The earned `.ok` publishes first, the refusal is removed second, so a crash between them leaves *both* artifacts and the gate **shut** (fail-safe) rather than neither.
- **The anti-gaming invariant is preserved and tested.** A plain earned write still never touches a refusal, so "re-run the auditor until it passes" remains blocked. Refusal precedence never decays into "newest marker wins".
- The gate reader needs **no logic change** — it clears because the refusal is gone. A comment records why it deliberately does no timestamp inference.

## #941 — a lost report read as a clean pass

The no-op classifier short-circuits on **marker presence**, so a member that completed, wrote a valid earned marker, and whose report never reached the orchestrator was indistinguishable from a genuine clean pass: dispatch classified real, retry suppressed, operator left holding a green gate with zero visible findings — including the Suggestions the clean-pass contract obliges them to resolve.

**Topology finding worth recording:** the filed premise points at a "team-based dispatch, where a member runs as an addressable teammate." GAIA has no such thing — audit dispatch is a depth-1 star of foreground `Agent(run_in_background: false)` calls (`wiki/concepts/Task Orchestration.md`). The failure is unreachable via the sanctioned path. But the underlying gap is topology-independent, and the durable artifact needed to close it **already exists**: every specialized member writes a findings sidecar on each local pass.

**Settled design: the durable sidecar is the report of record.** `audit-noop-detect.sh` takes `--findings` and requires *both* the earned marker and that sidecar before short-circuiting. Omitting the flag preserves the previous marker-only behavior for the default member and for a run whose base sha did not resolve.

## Adopter impact

**This ships.** The merge gate, clearance lib, writer, classifier, `code-audit-frontend`, `code-audit-github-workflows`, and the wiki page are all in the manifest; only the three `code-audit-maintainer-*` specs are release-excluded. Reaches existing adopters on their next `/update-gaia`. `code-audit-maintainer-prose` is correctly excluded from the supersede prose — it never refuses by contract.

## Verification

- **2046 bats, zero failures** across `.gaia/scripts/tests/`, `.gaia/tests/hooks/`, `.gaia/tests/lib/`, `.gaia/tests/prose-audit/` — run under the bash-5 guard to match CI, including the doc-grep contract suites that a prose edit can break invisibly.
- **10 new cases**: supersede round-trip (refusal removed + reason recorded), the anti-gaming plain-earned-write case, both usage errors, idempotent no-refusal write, lost-report detection, malformed-sidecar detection, back-compat without `--findings`, and orphan-sidecar fall-through.
- `pr-merge-audit-check.bats`'s frozen `code-audit-maintainer-shell: REFUSED` literal preserved — only the parenthetical remedy text changed.
- **shell-lint clean**: 167 scripts, 145 suites.
- Quality Gate skipped per its own documented rule (no staged source or gate-affecting config); the binding gates here are shell-lint and bats.

## Notes

No new tech-debt issues filed — this is a machinery fix, per the `machinery_waived` policy. No `.gaia/cli/src/**` touched, so no binary rebuild is owed.

Closes #931
Closes #941

---

## Round 2: clean pass, three Suggestions accepted-not-fixed

`code-audit-maintainer-shell` refused round 1 over a real Important finding (the findings gate validated shape but not identity, letting one member's sidecar vouch for another's lost report — the exact #941 failure). Fixed, plus all three of that round's Suggestions folded in free while its digest was already rotating. Round 2 is a **clean pass, marker earned**, with the member independently re-deriving the mutation kills in a scratch sandbox rather than accepting my account.

Its round-2 Suggestions are **accepted, not fixed**. The PR is clean and the member is marked, and every file involved is one it owns, so folding any of them in buys a full ~200k-token re-dispatch purely to re-earn the marker — the over-rotation the digest-economics policy exists to prevent.

1. **Ordering pin is textual, not behavioural** (`audit-write-clearance.bats`). It is the only test that catches the `mv`/`rm` swap and every failure mode fails *closed*, but it cannot see through indirection and would red on a benign extract-to-helper refactor. A PATH-shim ordering log would be strictly stronger. Fold in free if that suite is touched again.
2. **The `supersedes` block records nothing from the artifact it deletes** (`audit-write-clearance.sh`). It carries a constant, the author's reason, and its own timestamp, so after supersession the reversal is self-attested rather than corroborated — a slightly weaker property than the header comment claims. Folding the prior refusal's body into the block is a one-line `jq` read. No live failure mode. Fold in free if that writer is touched again.
3. **A remit line names a path the roster assigns to nobody** — `code-audit-maintainer-shell.md:26` claims `.gaia/cli/templates/workflows/code-review-audit.yml.tmpl`, which the roster does not give it or anyone, so a diff touching only that tracked file dispatches zero members and opens the gate unaudited. Graded Suggestion because the *authored* source under `src/` classifies to `code-audit-maintainer-node`, the build overwrites the generated copy, and the two are currently byte-identical, so no authored change escapes review.

**Suggestion 3 is machinery-waived, not filed**, per the `machinery_waived` policy: it is a finding about the audit gate's own machinery surfaced while auditing a fix to that machinery, and filing it is exactly the regeneration loop that policy closed. Both cited paths are machinery, so it is eligible. This listing is its durable record.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.VMjcWVEHRw

PR #971, opened 2026-07-21
debt(audit): reconcile the shell member's remit with the ownerless-by-design template artifact

### LINKED ISSUE #970: Audit gate: a tracked file in the shell member's remit is owned by no roster member, so a diff touching only it merges unaudited
`.claude/agents/code-audit-maintainer-shell.md:26` (the member's declared remit) versus `.gaia/audit-ci.yml:86-98` (the roster entry for that same member).

## Failure mode

`code-audit-maintainer-shell`'s own definition claims `.gaia/cli/templates/workflows/code-review-audit.yml.tmpl` in its remit list. The roster entry for that member carries no glob reaching it: the ten it declares are `.gaia/**/*.sh`, `.gaia/**/*.bats`, `.claude/hooks/**/*.sh`, `.specify/extensions/gaia/lib/*.sh`, `.github/**/*.sh`, `.github/**/*.bats`, `.gaia/audit-ci.yml`, `.gaia/VERSION`, `.claude/agents/code-audit-*.md`, and `.claude/rules/**`, and a `.tmpl` path matches none of them. No other member's globs reach it either.

Run through the real ownership classifier, that tracked path returns owner `-` and `audit_out_of_scope_allowlisted` true. A diff touching only that file therefore dispatches zero Code Audit Team members, and the merge gate opens with no audit at all.

Concrete trigger and outcome: an engineer reads the remit line, believes the file is covered, edits that tracked path directly, and it merges unreviewed. The edit is then silently discarded, because `bundle:adopter` (`.gaia/cli/package.json:8`) runs `rm -rf templates && cp -r src/automation/templates/workflows/. templates/workflows/`, regenerating the committed copy from source.

## Why this is Suggestion rather than Important

Three mitigations bound the blast radius, each verified rather than assumed:

- The authored source `.gaia/cli/src/automation/templates/workflows/code-review-audit.yml.tmpl` classifies to `code-audit-maintainer-node`, so no authored change escapes review.
- The build regenerates the committed copy wholesale, so a hand edit to it does not survive a bundle.
- The two files are currently byte-identical.

The unreviewed-merge window is real but narrow and self-correcting. The auditor that surfaced this explicitly offered to be overridden to Important if that window is treated as live.

## Suggested fix

Reconcile the two declarations in one direction. Choosing the direction is the substance of the work:

**(a)** Add the glob to the `code-audit-maintainer-shell` roster entry in `.gaia/audit-ci.yml` so the file dispatches that member. This has to stay in lockstep with the built-in fallback roster, which `verify-audit-roster.sh` enforces.

**(b)** Drop line 26 from `.claude/agents/code-audit-maintainer-shell.md` and note that the generated copy is covered transitively through its `src/` source.

The adjacent code leans toward (b): the bundle step proves the committed file is a build artifact rather than an authored one. A third option worth weighing is freshness-gating it in CI the way the committed binaries already are, instead of auditing it.

## Provenance

Surfaced by `code-audit-maintainer-shell` in round 2 of PR #969 (merged as `8f21b1da`), which closed #931 and #941. It was recorded there as `machinery_waived` rather than filed at merge time, per the policy that a finding about the audit gate's own machinery, surfaced while auditing a fix to that machinery, belongs in the PR body instead of the tracker. The maintainer has since asked for it to be tracked deliberately, so that waiver is knowingly overridden here. Both records stand as written: PR #969's body is the account of the disposition taken at merge, and this issue is the later decision to track it anyway.

Handler: prompt

stevensacks:
Closes #970

## What was wrong

`code-audit-maintainer-shell`'s own definition listed `.gaia/cli/templates/workflows/code-review-audit.yml.tmpl` in its remit. No roster glob reaches that path, in `.gaia/audit-ci.yml` or in the built-in fallback roster, so the declaration and the routing disagreed.

Verified against the real ownership classifier rather than assumed: every one of the 34 tracked files under `.gaia/cli/templates/` returns owner `-`.

```
$ audit_scope_init "$ROOT"; ... | audit_owners_for_paths
-  .gaia/cli/templates/workflows/code-review-audit.yml.tmpl
-  ... (33 more, all ownerless)
```

The failure mode is the one the issue describes: an engineer reads the remit line, believes the file is reviewed, edits it, and it merges with no member dispatched.

## Which direction, and why

The issue offered two directions and left the choice as the substance of the work. The evidence points one way: **the ownerless routing is deliberate and already pinned**, so the remit line is the stale half.

Four independent records say so:

- `.gaia/tests/hooks/audit-scope-routing-parity.bats` names the path as one of five sets that **deliberate roster changes move**, expecting `-`.
- `.gaia/tests/hooks/audit-scope-lib.bats` carves it out of the SEC-007 "every machinery path is owned" invariant, by exact name so every other path still fails closed.
- `.gaia/cli/src/automation/__tests__/audit-template-dogfood.test.ts` pins all twelve templates byte-identical to their source, and calls itself "the pin that trade rests on".
- `.claude/rules/maintainers/github-workflow-distribution.md` labels the file a **build artifact**: "Do not hand-edit; regenerate it."

The issue's third option, freshness-gating it in CI, already exists: that drift-guard is the freshness gate.

So this takes direction (b), and adds a short note recording *why* a glob does not belong there, so the carve-out reads as intentional rather than as an oversight for a future reader to "fix" by re-adding the glob.

## Audit findings, resolved

Round 1 came back clean but raised two findings of the same stale-count class, both resolved in the second commit:

- **Suggestion**, `.claude/agents/code-audit-maintainer-shell.md:10`: said "the **three** `code-audit-*` agent definitions"; the glob matches five.
- **Cross-remit**, `wiki/decisions/Code Audit Team.md:82`: said "**four**", same error.

Both numerals are **dropped** rather than corrected, so neither can go stale again when a sixth member is seeded. Neither count fed dispatch, which reads the roster, so no routing changed. Round 2 verified both resolutions and returned clean.

Round 2 raised one further cross-remit finding: `wiki/decisions/Code Audit Team.md` still says "the **twelve** pinned build-artifact templates" (also at line 26, outside this diff).

Left as-is deliberately, and the distinction is the point: "three" and "four" were **false**, while "twelve" is **true** and verified on disk. Removing a correct, orienting count to pre-empt hypothetical future drift trades real information for a speculative gain, and the same numeral sits on a second line this PR does not touch, so fixing only the one in the diff would leave the page inconsistent. Recorded here rather than filed, since a currently-correct number is not a defect.

## Verification

- Declared remit now matches **both** rosters exactly, ten globs each (`.gaia/audit-ci.yml` and `_audit_scope_builtin_roster`), checked by diffing the parsed lists.
- `audit-scope-lib.bats` + `audit-scope-routing-parity.bats`: 27 tests pass.
- `audit-template-dogfood.test.ts`: 5 tests pass, so the drift guard this note cites genuinely holds.
- Code Audit Team gate: `code-audit-maintainer-shell` dispatched and cleared with an earned marker; spawn set now resolves empty.
- Quality Gate: **skipped, nothing to check.** The diff is markdown only, no staged source or gate-affecting config.
- CHANGELOG: `## [Unreleased]` → `### Fixed` entry added.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Ye3lWyALKc

PR #978, opened 2026-07-21
feat(audit): derive Code Audit Team remits from the roster

### LINKED ISSUE #972: Audit gate: a member's declared remit and the roster that dispatches it can drift, with no check comparing them
`.gaia/scripts/verify-audit-roster.sh:26` (the enumerated invariant list) versus `.claude/agents/code-audit-maintainer-node.md:12` (a member's declared remit).

## Failure mode

Every Code Audit Team member declares its file ownership **twice**, in two surfaces with different consumers:

| | Where | Consumer | Decides |
|---|---|---|---|
| Machine-readable | `.gaia/audit-ci.yml` `auditors[].globs`, mirrored in `_audit_scope_builtin_roster` (`.claude/hooks/lib/audit-scope.sh`) | `audit_owner_for_path` / `resolve-audit-spawn.sh` | who is dispatched, whose marker the merge gate demands |
| Prose | `.claude/agents/code-audit-*.md`, the "Remit and self-skip" section | the agent model, once spawned | which changed files it reviews, and whether it self-skips |

Nothing enforces that the two agree. `verify-audit-roster.sh` declares five invariants at the cited line (claimant disjointness, default-member exclusion, undecidable-pair failure, agent-file plus machinery registration, and the `code-audit-` name prefix). None of them compares an agent definition's declared remit against the globs the roster assigns that member.

When the two drift, the prose is what a human reads and what the spawned agent filters against, so a file can read as reviewed while dispatching nobody.

## This is live, not hypothetical

`.claude/agents/code-audit-maintainer-node.md:12` declares "You own changed files under `.gaia/cli/src/**`, plus the CLI's build/config surface", and its filter instruction repeats "Filter `changed` to paths under `.gaia/cli/src/`". The roster enumerates only four extensions under that prefix: `.gaia/cli/src/**/*.ts`, `.gaia/cli/src/**/*.tmpl`, `.gaia/cli/src/**/*.snap`, `.gaia/cli/src/**/.gitkeep`.

Run through the real ownership classifier:

```
.gaia/cli/src/release/scrub.ts        code-audit-maintainer-node
.gaia/cli/src/config/defaults.json    -
.gaia/cli/src/schemas/thing.yml       -
.gaia/cli/src/release/fixture.txt     -
```

The extension narrowness is deliberate: `wiki/decisions/Code Audit Team.md` records it as "a named, accepted cost", noting that a future file under an unlisted extension "reaches no member". So the routing is correct and the **agent prose is the drifted surface**, over-claiming in the dangerous direction. Because `.gaia/**` is out-of-scope-allowlisted, a diff touching only such a file dispatches nobody and clears the merge gate through the legacy bypass, while the owning member's own definition says it covers it.

## Why this class recurs

The same two-surface drift has now been filed three times, each time against a different pair, and each time closed by fixing that one pair:

- #781: the builtin fallback roster omitted globs the committed roster granted. Closed by mirroring them; the dual-roster SEC-007 assertions in `.gaia/tests/hooks/audit-scope-lib.bats` now cover that pair.
- #970: the shell member's prose claimed a path no roster glob reached. Closed by dropping the claim.
- This issue: the node member's prose claims paths no roster glob reaches.

#781 and #970 are two directions of the same drift on the same `.tmpl` glob: the roster once granted it (per #781's body), then deliberately dropped it, and the prose kept the stale claim. Each instance was cheap to fix; the seam that generates them is not guarded.

## Why the fix is not mechanical

The five members express remit in five different shapes, so a naive bullet-list diff parses three of them:

- `code-audit-maintainer-shell`: bullet list, 10 globs, exact.
- `code-audit-github-workflows`: bullet list, 4 globs, exact.
- `code-audit-maintainer-prose`: one prose sentence, 1 glob, exact.
- `code-audit-maintainer-node`: prose sentence, over-claims (above).
- `code-audit-frontend`: bullet list that collapses five `*.config.*` globs onto one line, and which already says "Do not re-derive that set by hand" before calling `resolve-audit-spawn.sh --no-carry-forward` to decide its self-skip.

## Suggested fix

Pick a direction; choosing it is the substance of the work.

**(a)** Normalize the declaration shape first, giving every agent definition a machine-readable remit region, then add parity as a sixth invariant in `verify-audit-roster.sh`.

**(b)** Delete the duplication: have each definition defer to the roster and resolve its globs at runtime. `code-audit-frontend` is the in-repo precedent, since it already treats the oracle rather than its own bullet list as authoritative. Cost: an agent definition stops being self-contained documentation of its own domain.

**(c)** Keep the prose explicitly descriptive and non-normative, and check only that it does not *contradict* the roster.

Whichever direction wins must also reconcile the `code-audit-maintainer-node` drift, since any working check fails red on it until it is fixed.

## Grading note

Filed `severity:important` rather than `suggestion` because there is a live uncorrected instance, no automated guard, and the class has now recurred three times. A drainer who weighs the mitigations differently (maintainer-only tooling, and `.gaia/**` routing narrowness that is itself a documented accepted cost) could reasonably demote it to `suggestion`, which is how the single-instance case #970 was graded.

## Provenance

Surfaced while draining #970 through `/gaia-debt` and confirmed against the classifier during that run. Out of scope for PR #971, which fixed the single shell-member instance.

Handler: plan

stevensacks:
Makes `.gaia/audit-ci.yml` the single authority on which files each Code Audit Team member owns.

Today a member's `.claude/agents/code-audit-*.md` definition states its remit in hand-written prose, independently of the roster that actually decides dispatch. The two drift. The defect this closes is a definition claiming a path the roster grants nobody: the file reads as reviewed while dispatching no auditor.

Each definition now carries one marker-delimited remit region generated verbatim from the roster, `.gaia/scripts/write-audit-remits.sh` generates it, and `.gaia/scripts/verify-audit-roster.sh` gains a remit-parity invariant that fails loudly, names the offending glob, and prints the literal repair command. The four over-claiming definitions are narrowed by hand once. A dispatched member is denied the writer at the self-heal repair boundary: it edits audit-machinery paths, so a mid-audit run would rotate every member's content digest and void every clearance marker on the PR.

Closes #972

## Deferred: the update-flow tension

`.gaia/audit-ci.yml` is manifest class `shared` (adopter-tunable) while the two shipped agent definitions are `owned`. An adopter who tunes the roster and regenerates therefore produces `owned`-class drift that every later `/update-gaia` either prompts to overwrite or freezes into a sidecar conflict patch. This change resolves neither defensible repair (an announced post-update regeneration step carried as a CHANGELOG action-required marker, or a region-aware update flow) and adds no `**Action required:**` marker. Naming it here keeps it a live maintainer decision rather than a silent foreclosure. Maintainer decision, recorded at plan time. Tracked as #979 (not closed by this PR).

## Accepted-and-noted audit suggestions

The pre-merge `code-audit-maintainer-shell` pass cleared this PR and left two Suggestions, filed rather than folded in under the PR Merge Workflow's digest-economics guidance: the branch was already clean with every dispatched member marked, so applying them here would have rotated both members' digests and bought a full re-dispatch solely to re-earn the markers. Tracked as #988 and #989 (neither closed by this PR).


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.DzAXjUF5VE

PR #985, opened 2026-07-21
debt(audit): refuse the Playwright and Storybook trees in the self-heal boundary

### LINKED ISSUE #982: Audit self-heal boundary: the refusal set guards test/** but not .playwright/**, so a member can edit the spec that catches its own repair
**Location:** `.claude/hooks/lib/audit-selfheal-paths.sh:44`

## Failure mode

`AUDIT_SELFHEAL_REFUSE_ERE` is the single sourced refusal set naming the paths no Code Audit Team member may edit during a self-heal. It refuses `test/**` for a specific reason: a member must not be able to edit the test that would catch its own bad repair. It does not refuse `.playwright/**` or `.storybook/**`, which hold tests for the same reason.

Verified directly against the sourced ERE:

```
test/foo.ts                            REFUSED
.playwright/e2e/hydration.spec.ts      ALLOWED
.playwright/utils.ts                   ALLOWED
.storybook/preview.ts                  ALLOWED
wiki/x.md                              REFUSED
```

`.playwright/**` is claimed by no roster member in `.gaia/audit-ci.yml`, so it falls to `code-audit-frontend` as an in-scope-but-ownerless path, and that member runs with `push_fixes: true`.

The concrete bad outcome: a member repairs an `app/` file wrongly and, in the same self-heal pass, weakens or deletes the `.playwright/e2e/` assertion that would have caught it. Both enforcement points source this one ERE and carry no second path list, so neither stops it:

- `.claude/hooks/block-selfheal-paths.sh` (local, `PreToolUse`)
- the self-heal step in `.github/workflows/code-review-audit.yml` (CI, push-time staged-diff check)

No bats suite asserts that `.playwright/**` is refused, so nothing else guards it either. `.playwright/**` is additionally outside the ESLint-covered set, so lint does not backstop it.

## Scope

Maintainer-only and release-excluded machinery; no adopter clone runs this gate. The exposure is to the integrity of this repo's own merge gate, not to shipped code.

## Suggested fix

Add `.playwright` (and decide on `.storybook`) to the first alternation group, so the line reads `^(\.claude|\.specify|wiki|test|\.playwright|\.github/workflows)/|…`, then add a bats assertion pinning the refusal so the set cannot silently regress.

Two decisions the fix carries. First, whether `.storybook/**` belongs: its stories are test-adjacent but the directory is also ordinary config, and the `test/**` rationale reaches the stories more clearly than the config. Second, the comment directly above the ERE records that its root-config half is the `has_source` pattern from `code-review-audit.yml` "reused verbatim, so the two sets can never drift apart"; the addition targets the first group rather than that half, so confirm the invariant is untouched before editing, and mirror to the workflow template if it is not.

## Provenance

Raised as an Important cross-remit finding by `code-audit-frontend` during the pre-merge audit of #977, independently re-confirmed on a later round of the same PR. Owner is `code-audit-maintainer-shell`; no member repaired it, since a member may not edit the gate machinery that binds it.

Handler: plan

stevensacks:
Closes #982

## What

`AUDIT_SELFHEAL_REFUSE_ERE` is the one refusal set naming the paths no Code Audit Team member may edit during a self-heal. It refused `test/**` for a specific reason, that a member must not be able to edit the test that would catch its own bad repair, but did not refuse `.playwright/**` or `.storybook/**`, which hold tests for that same reason.

The concrete bad outcome: a member repairs an `app/` file wrongly and, in the same self-heal pass, weakens or deletes the `.playwright/e2e/` assertion that would have caught it. `.playwright/**` is claimed by no roster member, so it falls to `code-audit-frontend` as an in-scope-but-ownerless path, and that member runs with `push_fixes: true`.

## The change

One alternation token each for `.playwright` and `.storybook`, plus five bats assertions pinning the refusal.

Both trees are refused **whole**, matching how `test/**` is refused whole rather than narrowed to its assertion files. `.storybook/` was the open decision the issue carried: it holds no stories (they sit beside their components under `app/**/tests/`), but its config and decorators shape what Chromatic snapshots, and Chromatic is a required merge check, so a member that may edit them may suppress the visual regression its own repair caused. The cost of over-refusing is that a member reports a finding instead of repairing it; the cost of under-refusing is a silently weakened gate.

## What did not need to change

- **Both enforcement points** source this one set and carry no second copy: the local `PreToolUse` hook (`.claude/hooks/block-selfheal-paths.sh`) and the CI push gate in `code-review-audit.yml`. A repo-wide grep confirms no second copy of the alternation group exists.
- **The `has_source` invariant.** The comment above the ERE records that its root-config half is `code-review-audit.yml`'s own `has_source` file pattern, reused verbatim so the two can never drift. This addition targets the first alternation group, leaving that half untouched.
- **The three `code-review-audit.yml` copies** stay byte-identical (asserted by `self-heal-scope-gate.bats`, still green).

## Verification

| Check | Result |
| --- | --- |
| `.gaia/tests/hooks/block-selfheal-paths.bats` | 47/47 (5 new, RED before the fix) |
| `.github/audit/tests/self-heal-scope-gate.bats` | 11/11 |
| `.gaia/tests/shell-lint.sh` | passed (shellcheck + hook-array-guard) |
| Quality Gate | skipped, no staged source or gate-affecting config |

Direct check against the issue's own table:

```
test/foo.ts                                    REFUSED
.playwright/e2e/hydration.spec.ts              REFUSED   (was ALLOWED)
.playwright/utils.ts                           REFUSED   (was ALLOWED)
.storybook/preview.ts                          REFUSED   (was ALLOWED)
wiki/x.md                                      REFUSED
app/foo.ts                                     ALLOWED
.playwrightfoo/x.ts                            ALLOWED   (sibling-prefix, correctly not caught)
```

## Note on the issue's scope claim

The issue records this as "maintainer-only and release-excluded machinery; no adopter clone runs this gate." That is wrong: `.gaia/manifest.json` classes both `.claude/hooks/lib/audit-selfheal-paths.sh` and `.claude/hooks/block-selfheal-paths.sh` as `owned`, so both ship. Only the bats suite is release-excluded. The fix therefore reaches adopters on their next `/update-gaia`, and the CHANGELOG entry is written accordingly.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.5e03wNdt8H

PR #990, opened 2026-07-21
debt(audit): cover the remit writer's region-body emptiness guard

### LINKED ISSUE #989: The remit writer's region-body emptiness guard is never reached by any test
The region-body emptiness guard in the remit writer has no test that reaches it. It is correct defense-in-depth against a destructive silent failure, but nothing proves it works.

**Location:** `.gaia/scripts/write-audit-remits.sh:242` (`if [ ! -s "$body_file" ]; then`).

## Failure mode

The guard exists because awk's `while ((getline line < bodyfile) > 0)` cannot distinguish an unreadable body (`getline` returns -1) from an empty one (returns 0), and emits nothing either way. Without it, a failed body write installs a region holding the two markers and nothing between them over every `.claude/agents/code-audit-*.md` definition, dropping every glob bullet and the canonical sentence, while the writer still exits 0.

The only robustness case covering that path, in `.gaia/scripts/tests/write-audit-remits.bats`, drives the failure with `TMPDIR=<nonexistent>`. That trips the earlier `mktemp -d` guard at `:134`, which exits before the member loop, so the `-s` branch at `:242` is never executed by any suite. The guard's real targets are ENOSPC and EIO during the body write, with `mktemp -d` having succeeded, and no test induces that state. A future edit could therefore break or delete the guard with every suite still green.

## Suggested fix

Add a case that lets `mktemp -d` succeed and then makes the created `$tmpdir` unwritable (e.g. `chmod 500`) so the body write fails specifically, then assert the writer exits non-zero and leaves every definition byte-identical, mirroring the existing `bad-tmpdir` case's assertions. Skip the case when the suite runs as root, where the mode change does not bite.

Reported by `code-audit-maintainer-shell` on the pre-merge audit of PR #978, accepted-and-noted rather than folded in: that PR was already clean with every dispatched member marked, so applying it there would have rotated both members' digests and bought a full re-dispatch solely to re-earn the markers.

Handler: prompt

stevensacks:
## What

Adds the bats case that reaches the region-body emptiness guard in `.gaia/scripts/write-audit-remits.sh`.

## Why

The guard is correct defense-in-depth against a destructive silent failure: without it, a failed region-body write installs a region holding the two markers and nothing between them over every `.claude/agents/code-audit-*.md` definition, dropping every glob bullet and the canonical sentence, while the writer still exits 0. awk's `while ((getline line < bodyfile) > 0)` cannot tell an unreadable body (-1) from an empty one (0) and emits nothing either way, which is why the guard checks the body rather than any single cause.

Nothing proved it worked. The existing robustness case drives its failure with `TMPDIR=<nonexistent>`, which trips the earlier `mktemp -d` guard and returns before the member loop runs. The emptiness guard's real targets, ENOSPC and EIO during the body write with `mktemp -d` having succeeded, were unreachable by any suite, so a future edit could break or delete the guard with every suite still green.

## How

A PATH shim is the only handle on the directory: the writer names it with a random `mktemp` suffix, so no fixture can `chmod` it by name beforehand. The shim delegates to the real `mktemp`, then strips write permission from what it just created, leaving the directory readable and traversable so the body write is the only thing that fails. The writer carries no `set -e` by design, so the failed writes fall through to the guard rather than aborting.

The case asserts the emptiness guard is what refused (per member), that the earlier `mktemp` guard is *not* what fired, and that every definition is byte-identical. Skipped as root, where mode 500 does not deny the write.

## Verification

- Full suite green under bash 5 (`.gaia/scripts/bats5.sh`): 20/20.
- Mutation check: with the guard neutralized to `if false; then`, **only** the new case fails; the sibling `bad-tmpdir` case and the other 18 stay green. That is the coverage gap this closes, demonstrated rather than asserted.
- `.gaia/tests/shell-lint.sh` passes (shellcheck over 146 tracked bats suites).
- Quality Gate skips by its own rule: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.

No CHANGELOG entry: test-only, zero behavior change, and `.gaia/scripts/tests/` is release-excluded so no adopter clone is affected.

Closes #989

### DIFF
--- a/.gaia/scripts/tests/write-audit-remits.bats
+++ b/.gaia/scripts/tests/write-audit-remits.bats
@@ -605,6 +605,67 @@ YAML
   [ "$before_a" = "$after_a" ]
 }
 
+@test "robustness: a body write that fails after mktemp succeeds is refused, definitions byte-identical" {
+  # The sibling case above drives its failure with an unusable TMPDIR, which
+  # trips the mktemp guard and returns before the member loop ever runs. This
+  # one covers the state that guard cannot reach: mktemp -d succeeds and the
+  # region-body write inside the directory it created then fails, the
+  # ENOSPC/EIO class the region-body emptiness guard exists for. Without that
+  # guard the run installs a region holding the two markers and nothing
+  # between them over every definition, and still exits 0.
+  [ "$(id -u)" -eq 0 ] && skip "running as root: mode 500 does not deny the write"
+
+  local r="$BATS_TEST_TMPDIR/unwritable-tmpdir"
+  fixture_root "$r" <<'YAML'
+auditors:
+  - name: code-audit-default
+    globs:
+      - "app/**"
+    default: true
+  - name: code-audit-a
+    globs:
+      - "a/**"
+YAML
+
+  # A PATH shim is the only handle on that directory: the writer names it with
+  # a random mktemp suffix, so no fixture can chmod it by name beforehand. The
+  # shim delegates to the real mktemp, then strips write permission from what
+  # it just created, leaving it readable and traversable so the body write is
+  # the only thing that fails.
+  local shim real_mktemp
+  shim="$BATS_TEST_TMPDIR/mktemp-shim"
+  mkdir -p "$shim"
+  real_mktemp="$(command -v mktemp)"
+  cat > "$shim/mktemp" <<SH
+#!/usr/bin/env bash
+d="\$("$real_mktemp" "\$@")" || exit 1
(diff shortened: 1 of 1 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.AVSpGjZI3I

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.JkBIDJaU4X

PR #997, opened 2026-07-22
debt(ci): pass Storybook a literal placeholder, not the real SESSION_SECRET

### LINKED ISSUE #996: Chromatic passes the real SESSION_SECRET into a Storybook build that never uses it
**Location:** `.github/workflows/chromatic.yml:96`

## Failure mode

The Chromatic job passes the repository's real `SESSION_SECRET` into the Storybook build environment:

```yaml
SESSION_SECRET: ${{ secrets.SESSION_SECRET || 'ci-placeholder-session-secret-not-real' }}
```

Storybook never uses that value for its purpose. The secret signs cookies, and both consumers are server-only paths Storybook does not execute: `app/entry.server.tsx:20` (`setToastCookieOptions`) and `app/sessions.server/language.ts:9`.

The variable has to be *present*, not real. `app/env.server.ts:21` runs `schema.parse(process.env)` at module scope and the full schema declares `SESSION_SECRET: z.string()` (required). `app/root.tsx:16` imports from that module, so the parse can execute in the browser under `storybook dev`, which is why `.storybook/env.ts` shims the value onto `window.process.env`. Only a non-empty string is needed to satisfy the parse; the value is never read. The existing `|| 'ci-placeholder-session-secret-not-real'` fallback already demonstrates that a literal works, since the job passes when the secret is absent.

Nothing reaches the published artifact today. `.storybook/env.ts` exports nothing and is imported purely for side effect, and `package.json` declares `"sideEffects": false`, so the module is tree-shaken out of the preview bundle. A build of `storybook-static` contains zero occurrences of any key in the `viteFinal` `define` map, not only this one. The gap is that a bundling optimization is the only thing standing between a real secret and a build that publishes a public artifact, and no test asserts that property.

## Suggested fix

Pass a literal placeholder instead of the secret, so the build environment never holds the real value:

```yaml
SESSION_SECRET: 'storybook-placeholder-not-used-for-signing'
```

This keeps the Zod parse satisfied, leaves `storybook dev` and the Chromatic job working unchanged, and removes the dependence on tree-shaking for a value that has no reason to be there.

## Scope: `SITE_URL` needs no change

`SITE_URL` on the adjacent line (`chromatic.yml:97`) has the same shape, is shimmed by `.storybook/env.ts` for the same reason, and is omitted from `clientSchema` alongside `SESSION_SECRET` (`app/env.server.ts:16-19`). It was checked and deliberately left alone. Do not fold it in:

- **It is not a secret.** A deployed site URL is public by definition, so there is nothing to keep out of a public artifact. The rationale above does not transfer.
- **Its only consumer is server-only and never runs here:** `app/utils/http.server.ts:5` compares `request.headers.get('host')` against `new URL(env.SITE_URL).host`.
- **A placeholder would have to be a real URL anyway.** `new URL(...)` throws on a non-URL string, so unlike `SESSION_SECRET` (where any non-empty string satisfies `z.string()`), `SITE_URL` cannot take an arbitrary literal. The existing `|| 'http://localhost:3000'` fallback is already the correct shape.

Handler: prompt

stevensacks:
The Chromatic job handed the repository's real `SESSION_SECRET` to a Storybook build that never uses it for its purpose. The secret signs cookies, and both consumers are server-only paths Storybook does not execute: `app/entry.server.tsx` and `app/sessions.server/language.ts`.

The variable only has to be *present*. `app/env.server.ts` parses the environment at module scope and declares the field as a required `z.string()`, and `app/root.tsx` imports that module, so the parse can run in the browser under `storybook dev`, which is why `.storybook/env.ts` shims the value. Any non-empty string satisfies it, and the existing `|| 'ci-placeholder-session-secret-not-real'` fallback already proved a literal works, since the job passes when the secret is absent.

Nothing reached the published artifact before this either: `.storybook/env.ts` exports nothing and `package.json` declares `"sideEffects": false`, so the module is tree-shaken out of the preview bundle. That left a bundler optimization as the only thing standing between a real secret and a build that publishes a public artifact, with no test asserting the property. Passing a literal removes the dependence rather than continuing to rely on it.

## Verification

- `SESSION_SECRET: z.string()` in `app/env.server.ts` carries no length or format constraint, so the placeholder satisfies the parse.
- The workflow YAML parses and the step's `env` map resolves as intended.

## Scope

`SITE_URL` on the adjacent line is deliberately left alone, per the issue: a deployed site URL is public by definition, and `new URL(...)` throws on a non-URL string, so its existing `|| 'http://localhost:3000'` fallback is already the right shape.

`.github/workflows/tests.yml` carries the same `secrets.SESSION_SECRET || <literal>` shape and is also left alone. It publishes no artifact, and its Playwright run exercises the real cookie-signing path, so a real value there is coherent in a way it is not for a Storybook build. Noted rather than folded in.

Closes #996

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -93,5 +93,8 @@ jobs:
           COMMIT_SHA: ${{ github.event.pull_request.head.sha || github.sha }}
           MSW_ENABLED: ${{ secrets.MSW_ENABLED || 'true' }}
           npm_package_version: ${{ steps.package-version.outputs.current-version}}
-          SESSION_SECRET: ${{ secrets.SESSION_SECRET || 'ci-placeholder-session-secret-not-real' }}
+          # Storybook never signs a cookie, so this only has to satisfy the
+          # required z.string() in app/env.server.ts. Keep it a literal: the real
+          # secret has no reason to enter a build that publishes a public artifact.
+          SESSION_SECRET: 'storybook-placeholder-not-used-for-signing'
           SITE_URL: ${{ secrets.SITE_URL || 'http://localhost:3000' }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ A release change that requires the adopter to act, run a command or hand-migrate
 
 ### Changed
 
+- the Chromatic workflow no longer passes the repository's real `SESSION_SECRET` into the Storybook build. Storybook never signs a cookie, so the variable only ever had to satisfy a required environment schema, and it now receives a literal placeholder instead. Nothing reached the published static build before either, but that depended on a bundler optimization rather than on the secret never being there (#996)
 - ESLint now checks `.playwright/` and `.storybook/`. The lint preset ignored both by default, which silently disabled the very rules it configured for them, so every e2e spec, Playwright helper, and Storybook config file went unchecked while `pnpm lint` reported a clean tree. `@gaia-react/lint` 1.11.0 stops ignoring them, and the 78 findings that surfaced in GAIA's own tree are fixed here. **Action required:** run `pnpm lint` after updating. Most of what surfaces in your own e2e and Storybook code auto-fixes; what remains is real (#980)
 
 ### Added


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.pxafzgJmMg

PR #998, opened 2026-07-22
debt(worktree): derive the create hook's root from the common git dir

### LINKED ISSUE #993: Worktree hook derives its root from the current worktree, so a nested run silently skips typegen
`.gaia/scripts/create-worktree.sh:44`

## Failure mode

The `WorktreeCreate` hook derives its project root at line 44:

```bash
project_root="$(git rev-parse --show-toplevel)"
```

`--show-toplevel` resolves to the **current** working tree. When the hook fires from inside a linked worktree, that is the worktree, not the main checkout, so `project_root` names the wrong root for the rest of the run.

The visible consequence is in the typegen step. It borrows the CLI at `$project_root/node_modules/.bin/react-router`, and a linked worktree has no `node_modules`: `link-worktree.sh` symlinks the five `.gaia/local/` shared-state paths and the root `.env` / `.env.*` files, never `node_modules`. The `[ -x "$cli" ]` presence guard therefore returns early, and because that guard is deliberately silent (a checkout before its first install owes no typegen and should not say so), the nested worktree gets no `.react-router/types` **and no message explaining why**.

Bad outcome: the exact condition the typegen step exists to prevent, an agent linting against `error` typed values in a worktree, reappears in the nested case, with the one diagnostic that would name it suppressed.

Reproduced by the shell audit in a scratchpad repo: creating from the main checkout prints `create-worktree: generated typed routes in ...`; creating from inside an existing worktree produces a nested worktree with no typed routes and no output at all.

`worktree_path` (line 55) is built from the same `project_root`, so a nested run also lands its worktree under the inner worktree's `.claude/worktrees/` rather than the main checkout's.

## Mitigating context

The supported paths all prevent nesting, which is why this is filed as a suggestion rather than higher:

- The `EnterWorktree` runtime tool refuses to create a new worktree when the session is already in one.
- `.claude/skills/gaia/references/isolation.md`'s first decision arm resolves an already-inside-a-worktree session to "stay in it", explicitly declining to nest.

The reachable route is a direct invocation of the hook from a worktree cwd, which is how the audit reproduced it.

## Suggested fix

Sibling scripts already use the derivation that resolves the main checkout from anywhere, `dirname "$(git rev-parse --git-common-dir)"`:

- `.gaia/scripts/remove-worktree.sh:25`
- `.gaia/scripts/lib/ledger-path-lib.sh:16`
- `.gaia/scripts/mentorship-cleanup-sweep.sh:81`
- the `WorktreeRemove` hook registered in `.claude/settings.json:170`

Either align line 44 with that convention, or, if nesting is to stay unsupported, keep line 44 and split the two silent guard cases so the nested one still reports something rather than failing invisibly.

Note that aligning line 44 changes where a nested run places its worktree, not just where it looks for the CLI, so decide the intended nesting behavior as part of the fix.

Context: surfaced by the `code-audit-maintainer-shell` audit of the `/gaia-debt 987` fix, out of scope for that fix (line 44 predates it and was not changed by it).

Handler: prompt

stevensacks:
Closes #993

## Problem

The `WorktreeCreate` hook resolved its project root with `git rev-parse --show-toplevel`, which answers with whichever working tree the hook fires from. Every path it derives belongs to the **main checkout**:

- the `.claude/worktrees/` base it creates under
- the `node_modules/.bin/react-router` CLI its typegen step borrows
- the `.gaia/local/worktree-locks/` dir its per-name mutex uses (`link-worktree.sh` symlinks five `.gaia/local/` shared-state paths into a worktree, and the lock dir is not one of them)

So a run from inside a linked worktree nested the new worktree under the inner one, and found no `node_modules` to borrow. The typegen presence guard is deliberately silent, since a checkout before its first install owes no typegen, so the nested worktree got **no typed routes and no diagnostic either**: the exact lint-against-`error`-typed-values condition the typegen step exists to prevent, with its one explanatory message suppressed.

## Fix

Resolve the root through `dirname(absolute(git rev-parse --git-common-dir))`, which is identical from every tree of the repo. This is already the convention in `remove-worktree.sh`, `link-worktree.sh`, and the `WorktreeRemove` registration in `.claude/settings.json`.

The issue asked for the nesting behavior to be decided as part of the fix. **Nesting creates a sibling worktree under the main checkout**, because every consumer of these paths already assumes that root: `remove-worktree.sh` derives its prune base from the main checkout, the borrowed toolchain only exists there, and a lock under an inner worktree serializes against no peer.

Outside a git repo `--git-common-dir` exits 128, so the hook still fails fast under `set -e` exactly as `--show-toplevel` did. `pwd -P` keeps the resolved root physical, matching git's own path reporting, so a symlinked checkout resolves the same way.

## Verification

- New test 21 in `create-worktree.bats` runs the hook with cwd inside a linked worktree and asserts the new worktree lands under the main checkout, nothing lands under the inner tree, and typed routes are generated. Confirmed RED before the fix (failed on the placement assertion), GREEN after.
- Full `create-worktree.bats` (21), `link-worktree.bats`, `remove-worktree.bats`, and `shell-lint.bats` all pass under bash 5 via `.gaia/scripts/bats5.sh`.
- Quality Gate skips: no staged file is source or gate-affecting config (`.sh`, `.bats`, `.md` only). shell-lint is the applicable deterministic check and is green.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.0VCEn8QaMI

PR #1002, opened 2026-07-22
feat(update): region-aware merge for machine-generated file regions

### LINKED ISSUE #979: Update flow cannot merge a generated remit region, so a roster-tuning adopter hits a hand-resolved conflict on every update
`.claude/skills/update-gaia/SKILL.md:330`

## Failure mode

The three-way merge table classifies a whole file. Row 330 (`owned`, `A` != `B` and `A` != `L`) emits a conflict patch. Nothing in the table can express "this file has a machine-generated region that the adopter is expected to regenerate."

That collides with the roster-derived remit regions:

- `.gaia/audit-ci.yml` is manifest class `shared`, so tuning it is sanctioned adopter behavior.
- `.claude/agents/code-audit-frontend.md` and `.claude/agents/code-audit-github-workflows.md` are manifest class `owned`, and each carries a `gaia:audit-remit` region generated from that roster.
- `.gaia/scripts/verify-audit-roster.sh` fails when a region and the roster disagree, and prints `bash .gaia/scripts/write-audit-remits.sh` as the repair. That writer ships to adopters (manifest class `owned`).

Concrete sequence that reaches the bad outcome:

1. An adopter edits `.gaia/audit-ci.yml` to add or narrow a glob. Sanctioned; the file is `shared`.
2. The roster check now fails with a remit-parity finding and prints the repair command.
3. The adopter runs `bash .gaia/scripts/write-audit-remits.sh`. It rewrites the remit region inside the two `owned` definitions. This is the only sanctioned way to clear the failure.
4. Their working tree now has `A` != `B` for two `owned` files.
5. On the next `/update-gaia` that changes either definition upstream, row 330 fires: a `.gaia-merge/<path>.patch` conflict the adopter resolves by hand. Taking the release's side discards the regenerated region and the roster check goes red again; taking their own side drops whatever the release changed elsewhere in the definition.

The adopter followed every instruction correctly and still lands in a hand-resolved conflict, on every update that touches those files, indefinitely. Row 328 (`A` = `B`, no drift) never applies to them again.

Note this is not merely latent: prior to the roster-derived regions, an adopter tuning the roster got no signal to touch the agent definitions at all. Shipping a check that fails and hands them a command that edits `owned` files is what routes real traffic through the mismatch.

## Suggested fix

Teach the update flow that a marker-delimited region is generated, so a difference confined to that region is not adopter drift. Roughly: let the manifest declare a generated-region marker pair per path, compare `A` against `B` with each declared region masked out when choosing a row, and after writing the file re-derive the region from the adopter's own roster (or re-run the writer as a post-update step) rather than emitting a patch.

Alternatives worth weighing before committing to that, since they trade very differently:

- Reclassify the two shipped definitions as `shared`. Cheapest, but it hands adopters ownership of files GAIA needs to keep evolving.
- Leave the classes alone and add an announced post-update regeneration step, carried as a CHANGELOG `**Action required:**` marker. Honest, but nothing enforces it, and a missed run leaves the check red.
- Region-aware merge as described above. Most work, and the only option where the adopter never sees a conflict they did not cause.

Deferred deliberately by SPEC-056 (PR #978), which added the regions and named this tension in its PR body rather than silently foreclosing a choice.

Handler: spec

stevensacks:
Teaches `/update-gaia` that a marker-delimited region inside a shipped file is machine-generated, so an adopter who regenerates one never resolves a conflict they did not cause on that path.

Ships as three independently revertible units, in this order:

1. **Maintainer-side declaration**: the shared whole-line marker parser, the optional manifest `regions` key with its build-time scan, and the per-PR parity check. Adopter-visible behavior unchanged.
2. **Region oracle and normalized comparison**: `gaia update merge-region`, the headline value, delivered without any regeneration.
3. **Regeneration runner**: `gaia update regen-regions`. Separately revertible, and the commit that adds it says so.

Then the merge-walk prose, the reporting rows, the distribution and convergence suites, and the wiki pages.

Closes #979


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.WQr3ZItsi6

Issue #1011, opened 2026-07-22
The Storybook preview env key list is hand-maintained in three places and has already drifted from clientSchema


stevensacks:
## Failure mode

The set of environment keys the Storybook preview exposes to the browser lives in three hand-maintained lists that nothing forces to agree:

- `.storybook/main.ts:32-41`, the `define` map that inlines each key into the preview bundle.
- `.storybook/env.ts:2-9`, the shim that copies each inlined key onto `window.process.env`.
- `app/env.server.ts:16-19`, `clientSchema`, which is the app's own definition of which keys are safe to send to a browser and which types `Window['process']['env']` through `app/env.server.ts:25-34`.

They have already drifted. Both Storybook lists carry `SITE_URL`; `clientSchema` omits it. The shim key has no reader: nothing under `app/` reads `window.process.env.SITE_URL`, the declared type of that object has no such property, and the only consumer of the value anywhere is `app/utils/http.server.ts:5`, which runs on the server and reads `env.SITE_URL` directly. The shim key is dead weight that survives because no check compares the lists.

Nothing signals the drift, and nothing signals the next one either. A key added to the `define` map but not to the shim is inlined and never surfaces on `window.process.env`. A key added to the shim but not to the `define` map reads `undefined` at runtime. A key that `clientSchema` deliberately withholds from the browser reaches the browser anyway if someone adds it to the Storybook pair, which is exactly the shape `SITE_URL` is in now.

Separately, `.storybook/vite.config.ts:5` merges every key of every `.env` file into the build process's `process.env`:

```ts
Object.assign(process.env, loadEnv(mode, process.cwd(), ''));
```

Six of those keys have a consumer. The rest are ambient in a Node process that runs Vite plugins, Storybook addons, and their transitive dependencies, and that produces an artifact uploaded to a public snapshot host. A developer whose local `.env` holds real values makes all of them readable by every package in that build.

## Suggested fix

Give the key set one definition and derive the three consumers from it:

- Export the list from a single module.
- Build the `define` map in `.storybook/main.ts` from that list rather than writing each entry by hand.
- Read the loaded env into a local binding in `.storybook/vite.config.ts` and pass only the needed values through, instead of assigning into `process.env`.
- Keep the literal object in `.storybook/env.ts`, which needs its keys statically for the typed assignment, but make it the one place a human edits.

[...]

PR #1013, opened 2026-07-22
debt: drain five issues (Storybook shim, workflow permissions, guard docs)

### LINKED ISSUE #995: The remit-writer guard test names its fixture root for the sibling case's mechanism, and its mktemp shim's single-call-site assumption is unrecorded
Two clarity defects in the region-body emptiness-guard test case. Both are comprehension-level: neither changes what the suite proves, and the case is correct and green as written (its assertions were mutation-verified against a neutralized guard). What each costs is a future reader's time, and both mislead in the same direction, toward the wrong mechanism.

**Locations:** `.gaia/scripts/tests/write-audit-remits.bats:618` and `.gaia/scripts/tests/write-audit-remits.bats:638`.

## Failure mode

**1. The fixture root is named for the wrong thing (`:618`).** The variable is `$BATS_TEST_TMPDIR/unwritable-tmpdir`, but nothing about `TMPDIR` is unwritable in this test. That is the *sibling* case at `:582`, which drives its failure with `TMPDIR=<nonexistent>` and correctly names its root `bad-tmpdir`. Here `$r` is just the fixture root, and the unwritable thing is the anonymous directory the `mktemp` PATH shim `chmod 500`s.

The two cases exist precisely to cover two different guards: the sibling trips the `mktemp -d` guard at `write-audit-remits.sh:134`, and this one is the only test that reaches the region-body emptiness guard at `:242`. A maintainer debugging a failure here reads the name, concludes the test is about `TMPDIR`, and looks at the guard this case was written *not* to hit. The name actively points away from the distinction the case exists to draw.

**2. The `mktemp` shim's single-call-site assumption is unrecorded (`:638`).** The shim intercepts every `mktemp` in the writer's whole process tree and unconditionally `chmod 500`s whatever it returns, including a plain file if `-d` were ever absent. The blast radius today is exactly the one intended call site (`write-audit-remits.sh:134`); neither `verify-audit-roster.sh` nor `audit-scope.sh` invokes `mktemp`.

If a collaborator later gains one, the coupling fails **red** rather than greening hollow, so the suite stays trustworthy. But the failure it produces is misdirected: the writer exits 1 via `--emit-roster failed`, and the assertion that trips is the body-guard `assert_contains`, so the message points at the roster reader rather than at the shim that actually broke it. The assumption is load-bearing and written nowhere.

## Suggested fix

1. Rename the fixture root to `unwritable-body-dir` (or similar), so the name describes the directory the test actually makes unwritable rather than borrowing the sibling case's mechanism.
2. Add a one-line comment at the shim recording that it assumes the writer's process tree has exactly one `mktemp` call site, and that a second one would surface as an `--emit-roster failed` exit rather than as an obvious shim problem.

Both are confined to the one test case and change no assertion.

Reported by `code-audit-maintainer-shell` on the pre-merge audit of PR #990, accepted-and-noted rather than folded in: that PR was already clean with its one dispatched member marked, so applying either nit would have rotated the member's digest and bought a full re-dispatch solely to re-earn the marker.

Handler: prompt

### LINKED ISSUE #1000: Storybook inlines SESSION_SECRET into the preview bundle, and a local .env can supply the real value
**Location:** `.storybook/main.ts:40`

## Failure mode

`SESSION_SECRET` is inlined into the Storybook preview bundle at build time, and the value it inlines can come from a developer's local `.env`.

Three pieces make the chain:

- `.storybook/vite.config.ts:5` runs `Object.assign(process.env, loadEnv(mode, process.cwd(), ''))`. The empty prefix means `loadEnv` returns **every** key in `.env`, not just `VITE_`-prefixed ones, so `SESSION_SECRET` lands in `process.env`. The `Object.assign` direction also means `.env` overwrites the ambient environment rather than deferring to it.
- `.storybook/main.ts:40-42` define-substitutes `'import.meta.env.SESSION_SECRET'` with `JSON.stringify(process.env.SESSION_SECRET)`, which bakes the value in as a string literal.
- `.storybook/env.ts:8` reads `import.meta.env.SESSION_SECRET` into `window.process.env`, and `.storybook/preview.ts:6` imports that module, putting it in the preview graph.

Anyone with a real value in `.env` who runs `pnpm build-storybook && pnpm chromatic` (both are package scripts) publishes that value inside a JS chunk on a public snapshot host. No CI workflow is involved, so a workflow-level placeholder does not prevent it.

Today the value happens not to survive into `storybook-static`, because `.storybook/env.ts` exports nothing, `package.json` declares `"sideEffects": false`, and the module is tree-shaken out. That is a bundler optimization, not a guarantee: `"sideEffects": false` is an assertion about this package's own modules, and `.storybook/env.ts` assigns to `window.process.env`, which is precisely a side effect. The claim is already false for that file, so the tree-shake that currently saves it is not something to rely on.

## Suggested fix

Remove `SESSION_SECRET` from the shim entirely rather than feeding it a safer value. `clientSchema` in `app/env.server.ts:16-19` already omits `SESSION_SECRET`, so no browser code legitimately reads it:

- drop the `SESSION_SECRET` line from `.storybook/env.ts:8`
- drop the `'import.meta.env.SESSION_SECRET'` entry from the `define` map in `.storybook/main.ts:40-42`

The shim exists because `app/root.tsx:16` imports `app/env.server.ts`, whose module-scope `schema.parse(process.env)` requires `SESSION_SECRET` as a `z.string()` and would otherwise throw in the browser. So the fix depends on which of these holds:

- **No story pulls `app/root.tsx` into the client graph** -> the parse never runs in the browser and both lines can simply go.
- **Some story does** -> the parse still needs a non-empty string, so the shim keeps a hardcoded literal in `.storybook/env.ts` instead of a `define`-substituted build-time value. That keeps the real environment out of the bundle while satisfying the parse.

Establish which case applies before choosing; that is the decision this issue leaves open.

Consider also narrowing the `loadEnv` prefix at `.storybook/vite.config.ts:5` so a local `.env` cannot introduce arbitrary server-only keys into a browser build in the first place. That is defense in depth for the whole `define` map, not just this one key.

## Scope

`SITE_URL` is shimmed by the same mechanism and is deliberately not part of this: a deployed site URL is public by definition.

Related: #996 removed the real secret from the Chromatic workflow's build environment. This is the same exposure one layer down, at the source rather than at the CI boundary.

Handler: plan

### LINKED ISSUE #1001: chromatic.yml declares no permissions block, so a third-party action gets the default GITHUB_TOKEN scope
**Location:** `.github/workflows/chromatic.yml:11`

## Failure mode

Neither the workflow top level nor either job in `chromatic.yml` declares a `permissions:` block, so both jobs inherit the repository default `GITHUB_TOKEN` scope.

`.github/workflows/chromatic.yml:83` passes `secrets.GITHUB_TOKEN` to the third-party `chromaui/action`. When the repository default is "read and write", that action receives a write-scoped token across every scope (contents, issues, pull requests, packages, and the rest) for a job whose only legitimate need is to report snapshot results. A compromised or malicious release of that action, or any transitive dependency it executes, would run with far more authority than the task requires.

`chromatic.yml` and `cli-tests.yml` are the only two of the thirteen workflows in `.github/workflows/` with no `permissions:` block. `tests.yml:20-24` is the in-repo convention this diverges from.

## Suggested fix

Declare an explicit least-privilege `permissions:` block on the `chromatic` job.

Do not copy `tests.yml` verbatim: Chromatic's action posts commit statuses, so it needs more than `contents: read`, and pinning too narrow a set breaks the status reporting the required check depends on. Confirm the minimum set against Chromatic's own documentation for the action version pinned at `.github/workflows/chromatic.yml:81`, then declare exactly that.

The `has-chromatic-token` job does no API work and can take `permissions: {}`.

Note that `chromatic.yml` ships verbatim to adopters (manifest class `shared`), so whatever is pinned becomes the default every adopter inherits.

Handler: prompt

### LINKED ISSUE #1009: Self-heal guard follow-up: the SC2016 directive covers one test of six, and the ( padding's over-deny surface names three instances of four
Two documentation-accuracy defects left behind by the self-heal exec-position fix (#988, PR #1006). `code-audit-maintainer-shell` reported both as Suggestions on that PR's final, clean round; they were accepted-and-noted rather than folded in, because the PR was already marked and both files sit inside that member's globs, so applying them would have rotated its content digest and bought a full re-dispatch for cosmetics.

Neither has a live failure mode today. Both are cheap to fold into any later PR that already rotates `code-audit-maintainer-shell`'s digest, which is the intended way to drain this.

## 1. The SC2016 directive covers one `@test` out of the six it was written for

**Location:** `.gaia/tests/hooks/block-selfheal-paths.bats:530`

A `# shellcheck disable=` directive scopes to the next command or compound structure, and in a bats file each `@test` parses as its own function. The directive at line 530 therefore suppresses SC2016 for only the `@test` at line 534. Confirmed against the merged tree:

```
$ shellcheck --severity=info --format=gcc .gaia/tests/hooks/block-selfheal-paths.bats | grep SC2016
.gaia/tests/hooks/block-selfheal-paths.bats:540:39: note: Expressions don't expand in single quotes [SC2016]
.gaia/tests/hooks/block-selfheal-paths.bats:545:39: note: Expressions don't expand in single quotes [SC2016]
.gaia/tests/hooks/block-selfheal-paths.bats:550:39: note: Expressions don't expand in single quotes [SC2016]
.gaia/tests/hooks/block-selfheal-paths.bats:555:39: note: Expressions don't expand in single quotes [SC2016]
.gaia/tests/hooks/block-selfheal-paths.bats:560:39: note: Expressions don't expand in single quotes [SC2016]
```

### Failure mode

The directive's own comment claims block-wide reach ("hollow out every assertion below"), so a reader who trusts it is misled about five of the six tests. The risk it exists to prevent is concrete and already-demonstrated: a contributor "fixing" SC2016 the way the message suggests would double-quote the payloads, which expands `${R}` and `$(pwd)` away at test-authoring time. The command reaching the hook would then carry no `${VAR}` or `$(cmd)` at all, and all six regression tests would keep passing against a guard that no longer defends what they were written to pin. Those tests exist because exactly that bypass shipped twice during PR #1006's own review, so a silent hollowing-out here is a real regression path, not a hypothetical one.

Not currently enforced: `.gaia/tests/shell-lint.sh` holds `*.bats` to a `warning` floor and SC2016 is `info`. Note that #690 already tightened `*.sh` to `style`; if `*.bats` ever follows, these five become hard failures.

### Suggested fix

Repeat the directive above each of the remaining five `@test`s. Do **not** hoist a single file-scope directive to the top of the file: that would also suppress SC2016 in every future test in this suite, including one where an unexpanded expression is a genuine defect, which is a strictly worse trade than five repeated lines. Then trim the comment's "every assertion below" wording so its stated scope matches its actual one.

## 2. The `(` padding's over-deny surface names three of its four instances

**Location:** `.claude/hooks/block-selfheal-paths.sh:270`

The comment block documents the deny-widening consequences of padding `(` into a standalone token, naming the array-literal case and the `(cd foo && ...)` case explicitly, and leaving a fourth to the general clause "Padding can also split a quoted argument, which only ever widens the deny surface". That fourth instance is paren-led quoted prose:

```
$ echo "(.gaia/scripts/write-audit-remits.sh)"     -> DENY
```

### Failure mode

No wrong behavior: the deny is inside the documented rationale and over-denying is the deliberate safe direction for this guard, so **no code change is wanted here**. The defect is that this is the likeliest of the four instances to actually be hit, by a dispatched member writing ordinary prose that happens to parenthesize the writer's path, and it is the one instance a reader has to derive from a general clause instead of finding named. A member that hits it sees a deny with no matching example in the comment that explains it, and the natural next move is to assume the guard is broken and try to work around it.

### Suggested fix

One clause added to the existing comment, naming `echo "(<writer>)"` alongside the two instances already named. No code change.

## Scope note

Two files, both edits comment-only, no behavior change, no public-contract change, no cross-module ripple. They are filed together because they are one economic unit: both sit in `code-audit-maintainer-shell`'s globs, so fixing either one alone costs the same digest rotation and re-dispatch as fixing both.

Handler: prompt

### LINKED ISSUE #1012: cli-tests.yml declares no permissions block, so both jobs inherit the default GITHUB_TOKEN scope
**Location:** `.github/workflows/cli-tests.yml:13`

## Failure mode

Neither the workflow top level nor either job in `cli-tests.yml` declares a `permissions:` block, so both `cli-tests` (`:14`) and `distribution-harness` (`:94`) inherit the repository default `GITHUB_TOKEN` scope. When that default is "read and write", every step in both jobs runs with write authority across every scope for work whose only need is to check out the repo and run a test harness.

This is the same class as the `chromatic.yml` finding, and with that one resolved `cli-tests.yml` is the last of the thirteen workflows in `.github/workflows/` with no `permissions:` block. `tests.yml:20-24` and the top-level default-deny in `forensics-triage.yml:22` are the two in-repo conventions it diverges from.

Blast radius is smaller than the `chromatic.yml` case in two ways: the jobs invoke no third-party action that receives the token, and the workflow is maintainer-only and release-excluded, so no adopter inherits it. That is why this is a suggestion rather than an important-tier finding. It is still the one remaining workflow where a job added later silently starts with ambient write scope.

## Suggested fix

Declare `permissions: {}` at the workflow top level so the file is fail-closed by default, then grant each job exactly what it needs. Confirm the minimum per job against what its steps actually do rather than copying another workflow's set; both jobs check out the repo and run tests, so `contents: read` is the expected answer, but verify each step's own action inputs before pinning it.

The top-level default-deny is the shape that makes this durable: it means a third job added to the file inherits no scope rather than the repository default, which is the failure mode this issue exists to close.

## Scope

`cli-tests.yml` is maintainer-only and release-excluded, so this changes nothing an adopter receives.

Handler: prompt

stevensacks:
Drains five tech-debt issues in one unit. The five are unrelated by file, so they were fixed concurrently by four agents on disjoint paths, then gated once as a combined diff.

Closes #995
Closes #1000
Closes #1001
Closes #1009
Closes #1012

## What changed

**#1000, Storybook inlined `SESSION_SECRET` into the preview bundle.** The issue left one decision open: whether any story pulls `app/root.tsx` into the client graph, which would mean the module-scope `schema.parse(process.env)` still needs a non-empty string and the shim has to keep a hardcoded literal. It does not. Nothing value-imports `app/root.tsx` (the only two references are type-only), the stories' decorators build routes from inline literals, and the empirical check is decisive: none of `clientSchema`'s own key names appear anywhere in a built `storybook-static`, so `app/env.server.ts` is not in the client graph at all. Both the `define` entry and the shim line are therefore removed outright.

**#1001, `chromatic.yml` declared no `permissions:` block.** The issue's stated premise turned out to be wrong and was not acted on. It asserts Chromatic's action posts commit statuses and so needs more than `contents: read`. At the pinned `v16.9.1`, `chromaui/action` makes no GitHub API calls: `getOctokit` is never called in the deployed bundle, `token` is not among the 40 inputs the action source reads, and upstream removed the need for it. The statuses come from the Chromatic GitHub App via `projectToken`. So `contents: read` is provably sufficient and cannot break the required check.

**#1012, `cli-tests.yml` had the same gap** and was filed and folded in here, since it is the same class in the same directory and the workflow auditor is already dispatched for this diff. With both fixed, no workflow in the repo leaves a job on the ambient default.

**#995 and #1009 are comment and naming accuracy** in guard tests and a hook rationale. No behavior change, no assertion changed.

## Scope decisions worth reviewing

**The `loadEnv` narrowing #1000 suggests as optional defense-in-depth was implemented, then reverted.** Tracing the actual bundle paths showed it closes none of them. `SESSION_SECRET` reached the bundle because it had a `define` entry, not because `loadEnv('')` populated `process.env`; and Storybook's own env plugin (`stringifyProcessEnvs` in `@storybook/builder-vite@10.5.0`) inlines only `allowedEnvVariables` and `VITE_`/`STORYBOOK_`-prefixed keys, sourcing from its own env preset, so it is unaffected by `.storybook/vite.config.ts` either way. The narrowing bought a different threat class than the issue describes (what third-party build code can read out of `process.env`) while adding a third hand-maintained copy of the preview key list, propagating an existing `SITE_URL` drift, and shipping a comment that overclaimed: `loadEnv`'s third argument is a prefix list, so `API_URL` also admits `API_URL_SECRET`. Filed as #1011 with the tracing recorded, so nobody re-derives it.

**`SITE_URL` is deliberately not removed**, though its shim key has no reader. #1000's Scope section excludes it explicitly, and the reason for removing it (dead key) is not the reason the issue gives for keeping it (public by definition). It belongs to #1011, which pairs it with the guard that would have caught the drift.

**#1001 grew slightly past its stated fix**, twice, both toward the issue's own failure mode. Workflow-level `permissions: {}` was added so each file is fail-closed for any job added later, matching the existing idiom in `forensics-triage.yml` and `code-review-audit.yml`; per-job blocks alone would have left the next job on the repository default, which is the hole the issue exists to close. And the dead `token: ${{ secrets.GITHUB_TOKEN }}` input, which the issue's Failure mode names directly, is deleted rather than only scope-limited.

Note that deleting the token input does not fully close that exposure on its own: `actions/checkout` persists credentials into `.git/config` in the workspace by default, so a compromised action can read a token off disk regardless. The scope pin is what bounds it; the deletion removes the explicit hand-off.

`chromatic.yml` ships verbatim to adopters (manifest class `shared`), so its permission set becomes every adopter's default. The three granted consumers (`checkout`, `paths-filter`, `setup-node`) are unconditional steps in every rendered copy and their needs do not vary by repo, so the set is correct generically rather than tuned to this repo.

## Verification

Quality Gate on the combined diff: typecheck 0, lint 0, unit tests 29 files / 155 passed / 1 skipped, Playwright 8 passed / 1 skipped, dev server HTTP 200, `pnpm build` 0, `build-storybook` 0. A grep of the built `storybook-static` finds no occurrence of `SESSION_SECRET`.

Shell and workflow surfaces: `shell-lint.sh` passed (169 scripts at `style`, 147 bats suites at `warning`, array-guard clean), bats 107 passing across both edited suites under the bash-5 guard, `shellcheck --severity=info` clean of SC2016 on the self-heal suite, `actionlint` clean on both workflows.

The `cli-tests.yml` and CHANGELOG edits landed after the full gate run and are YAML and markdown only, verified with `actionlint` plus a YAML parse asserting the resolved permissions on every job.

## Follow-ups filed

- #1011, single-source the Storybook preview env keys, remove the dead `SITE_URL` shim key, and add the guard asserting the shim's keys equal `clientSchema`'s. `Handler: plan`.
- #1012, folded into this PR rather than deferred.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Lpavp2PEzv

PR #1017, opened 2026-07-22
debt(cli): rebuild the stale maintainer binary and require its gate

### LINKED ISSUE #1015: The committed gaia-maintainer binary is stale, and the check that caught it is not a required context
**Location:** `.gaia/cli/gaia-maintainer`, and the gate at `.github/workflows/cli-tests.yml`

## Failure mode

The committed `.gaia/cli/gaia-maintainer` binary does not match the source it is built from, so the maintainer CLI in the tree runs code that is one fix behind `.gaia/cli/src/`.

`a0c9e35e` (#984, "drop glob fragments truncated mid-basename in the leak scan") changed `.gaia/cli/src/` and did not re-bundle. The adopter binary `.gaia/cli/gaia` is unaffected and reproduces byte-identically; only `gaia-maintainer` drifted, which matches the change, since #984 fixed release code that only the maintainer bundle contains.

The practical consequence is that #984's fix is not in the binary a maintainer actually executes. Its own commit message describes what that leaves in place: the leak scan reports a phantom runtime-dependency leak against a legitimate line of prose, the gate fails closed, and that "blocked every distribution scenario that builds a staging tree, and `/gaia-release` with it". Anyone running `/gaia-release` today hits the bug #984 believed it had fixed.

Reproduced locally, matching CI exactly:

```
cp .gaia/cli/gaia-maintainer /tmp/committed
pnpm -C .gaia/cli bundle
cmp -s /tmp/committed .gaia/cli/gaia-maintainer   # differs
```

`gaia` compares identical in the same run, so this is real drift rather than a nondeterministic build.

## Why the existing gate did not stop it

The gate works. It cannot block.

`.github/workflows/cli-tests.yml` runs a reproducibility step that rebuilds both binaries and fails when either differs from the committed copy. That step is what #754 added, and it did fire: #984 touched `.gaia/cli/**`, so the job's `paths-filter` matched and the check ran. It went red and the pull request merged anyway, because `Vitest (.gaia/cli)` is not in the repository's required status checks. `REQUIRED_CONTEXTS` in `.gaia/scripts/verify-required-checks.sh:39-45` declares five contexts and this is not among them.

So the drift is not a gap in detection. It is a gate whose verdict nothing consumes, which makes it advisory in practice no matter how correct its result is.

This surfaced only because a later pull request edited `cli-tests.yml` itself. That path is in the job's own `paths-filter`, so the job ran on a change where someone was reading its output. Absent that coincidence the stale binary would have reached the next release.

## Suggested fix

Two parts, and the second is the one that matters.

1. Rebuild and commit the binary: run `pnpm -C .gaia/cli bundle` and commit the resulting `.gaia/cli/gaia-maintainer`. Verify `gaia` stays byte-identical in the same run, so the commit carries only the drifted artifact.

2. Make the reproducibility verdict blocking. Add `Vitest (.gaia/cli)` to `REQUIRED_CONTEXTS` in `.gaia/scripts/verify-required-checks.sh` and promote it in the live branch ruleset. Note that the script deliberately does not make this decision on its own: its own comment says to keep the list "in sync with any future decision to promote or demote a check; that decision is what this script exists to guard, not to make on its own". So promoting the context in the GitHub ruleset is a human call that has to be made alongside the list edit, not inferred from it.

Weigh one thing before promoting the whole job: the check currently covers both the CLI vitest suite and the reproducibility step, and it self-skips via `paths-filter` when a pull request touches no CLI path, reporting green. Confirm the skip path still reports the context so a required check cannot hang pending on unrelated pull requests. If that shape is a problem, splitting the reproducibility step into its own always-reporting job is the alternative.

## Scope

Maintainer-only. `.gaia/cli/**` and `cli-tests.yml` are release-excluded, so no adopter clone receives either the binary or the gate. The blast radius is the maintainer release path, which is why this is filed as important rather than a suggestion: it silently disarms `/gaia-release`.

Related: #754 established the PR-time check and is closed. This is the follow-on that the check's verdict is not enforced.

Handler: prompt

stevensacks:
## What

`.gaia/cli/gaia-maintainer` did not reproduce from `.gaia/cli/src/`. `a0c9e35e` (#984) fixed the release leak scan without re-bundling, so the binary a maintainer actually executes was one fix behind: `/gaia-release` still hit the phantom runtime-dependency leak that commit removed, and the gate fails closed, so it blocked every distribution scenario that builds a staging tree.

`gaia` was unaffected and reproduces byte-identically, matching a fix that only touched maintainer-bundle code.

## Why the existing gate did not stop it

`cli-tests.yml`'s reproducibility step **did** fire on #984 and **did** go red. It could not block the merge because `Vitest (.gaia/cli)` was never in the branch ruleset's `required_status_checks`, so its verdict had no consumer. Detection was never the gap; enforcement was.

## Changes

- **Rebundle** `gaia-maintainer` (`pnpm -C .gaia/cli bundle`). `gaia` stays byte-identical in the same run, so the commit carries only the drifted artifact. The delta is exactly #984's `PATH_TRUNCATING_METACHAR` fix.
- **Declare** `Vitest (.gaia/cli)` in `REQUIRED_CONTEXTS` (`.gaia/scripts/verify-required-checks.sh`) and **promote** it in the live `main` ruleset. The script deliberately never writes to the ruleset, so promotion was taken as the explicit maintainer call its header documents; it was made before this PR so declared and live stay in sync and the advisory drift job is green here.
- **Correct two comments** that described the job as not declared-required (`cli-tests.yml`, `.gaia/tests/distribution/README.md`).

## Safety of promoting the context

The job is `if: github.event_name == 'pull_request'` with per-step paths-filter guards, so it always runs on a PR and reports green when the filter misses. Confirmed empirically on #1003, #1005, #1006, and #1010, all non-CLI PRs where it reported `pass`. A required context therefore cannot hang pending on unrelated work.

## Verification

- `pnpm -C .gaia/cli bundle` then `cmp`: both binaries reproduce
- `pnpm -C .gaia/cli test --run`: 124 files, 1707 passed
- `.gaia/scripts/tests/*.bats` under bash 5: 781 passed, 0 failed
- `bash .gaia/scripts/verify-required-checks.sh` against the live ruleset: exit 0, and the context no longer appears in the advisory list

## Scope

Maintainer-only. `.gaia/cli/**`, `.gaia/scripts/verify-required-checks.sh`, and `cli-tests.yml` are all release-excluded, so no adopter clone receives any of it.

Closes #1015


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.ZbryAEBzsY

PR #1018, opened 2026-07-22
debt(storybook): pin the preview env key list to clientSchema

### LINKED ISSUE #1011: The Storybook preview env key list is hand-maintained in three places and has already drifted from clientSchema

stevensacks:
Closes #1011

## What

The set of environment keys the Storybook preview exposes to the browser lives in hand-maintained lists that nothing forces to agree:

- `.storybook/main.ts`, the `define` map that inlines each key into the preview bundle
- `.storybook/env.ts`, the shim that copies each inlined key onto `window.process.env`
- `app/env.server.ts`'s `clientSchema`, the app's own definition of which keys are safe to send to a browser

They had drifted. #1000 removed `SESSION_SECRET` from the Storybook pair; `SITE_URL` is still inlined and copied onto `window.process.env` even though `clientSchema` withholds it. Nothing reads it: no `app/` code reads `window.process.env.SITE_URL`, the declared type of that object has no such property, and the only consumer of the value is server-side.

Removing `SITE_URL` fixes this instance. The rest of the change fixes the class.

## How

`clientSchema` already types `Window['process']['env']` through the `declare global` in `app/env.server.ts`, so both lists pin themselves to that type with `satisfies` rather than restating it:

- a key `clientSchema` withholds fails the excess-property check
- a key it allows but the list omits fails the `Record` check

Drift in either direction is now a `pnpm typecheck` error instead of something that surfaces at runtime, if at all. No new module, no new export, and no runtime test is needed to hold the lists in agreement.

`.storybook/vite.config.ts` also stops merging every `.env` value into the build process's `process.env`, where every Vite plugin, Storybook addon, and transitive dependency in that process could read it. `.storybook/main.ts` reads the env into a local binding instead, which is the only place the values are needed.

## Verification

- All four drift directions were confirmed to fail `tsc`: a withheld key added to either list, and an allowed key (including an optional one) dropped from either list. The restored state typechecks clean.
- Verified against a live `storybook dev` server that `import.meta.env` carries exactly the five allowed keys plus Vite's own built-ins, with real values inlined, and no `SESSION_SECRET` or `SITE_URL`.
- Full Quality Gate, re-run after the rebase onto current `main`: typecheck, lint, `pnpm test --run` (29 files, 155 passed / 1 skipped, matching the `main` baseline), `pnpm pw` (8 passed / 1 skipped), dev smoke (`/` and `/privacy` both 200), `pnpm build`, `pnpm build-storybook`.

## Out of scope

`vitest.config.ts` still does the same `Object.assign(process.env, loadEnv(...))` this PR removes from the Storybook config, and then forwards the merged env into every test environment. That is a defensible boundary (test processes are not preview builds, and `test/setup.ts` already seeds the required keys), but it is deliberately untouched here rather than quietly folded in.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.6iHpc9NK3C

PR #1024, opened 2026-07-22
debt(test): default npm_package_version in the Vitest setup

### LINKED ISSUE #1021: test/setup.ts is missing the npm_package_version fallback, so running Vitest outside a pnpm script aborts a suite
`test/setup.ts` supplies fallbacks for three of the five env vars `app/env.server.ts` requires, so running Vitest outside a package-manager script aborts a suite before any test runs.

**Location:** `test/setup.ts:11-13`

**Failure mode**

`app/env.server.ts:21` runs `schema.parse(process.env)` at import time, and that schema requires `API_URL`, `NODE_ENV`, `SESSION_SECRET`, `SITE_URL`, **and** `npm_package_version`. `test/setup.ts` defaults only the first, third, and fourth. `NODE_ENV` is supplied by Vitest, but `npm_package_version` is only ever set by npm/pnpm when the process is launched through a `package.json` script.

Reproduced on a clean checkout:

```
$ pnpm exec vitest --run app/utils/tests/http.server.test.ts
ZodError: [{"expected":"string","code":"invalid_type",
  "path":["npm_package_version"],
  "message":"Invalid input: expected string, received undefined"}]
 ❯ app/env.server.ts:21:27
 ❯ app/utils/http.server.ts:2:1
 ❯ app/utils/tests/http.server.test.ts:2:1

 Test Files  1 failed (1)
      Tests  no tests
```

The same file passes via `pnpm test:ci`, because pnpm sets `npm_package_version` for script-launched processes. So the failure hits exactly the invocations a developer is most likely to reach for: `npx vitest <file>`, `pnpm exec vitest`, and IDE/editor Vitest integrations, which spawn the binary directly rather than through a script. The suite does not fail partially, the module never imports, so the file reports zero tests.

`vitest.config.ts:12` masks this on machines that have a `.env` only if that file happens to define `npm_package_version`; the committed `.env.example` does not, so the mask is not something the repo provides.

**Suggested fix**

Add the missing default alongside the existing three, keeping the same `??=` shape so a real value always wins:

```ts
process.env.npm_package_version ??= '0.0.0';
```

Handler: prompt

stevensacks:
## Problem

`app/env.server.ts` parses `process.env` at import time and its schema requires `npm_package_version`, but pnpm sets that variable only for the processes it launches from a `package.json` script. `test/setup.ts` supplied fallbacks for three of the required vars and not that one, so a directly spawned Vitest aborted the suite before collecting a single test:

```
$ pnpm exec vitest --run app/utils/tests/http.server.test.ts
ZodError: [{"expected":"string","code":"invalid_type",
  "path":["npm_package_version"],
  "message":"Invalid input: expected string, received undefined"}]
 ❯ app/env.server.ts:21:27

 Test Files  1 failed (1)
      Tests  no tests
```

`pnpm test --run` was unaffected, so the gap hit exactly the invocations a developer reaches for while iterating on one file: `pnpm exec vitest` and editor Vitest integrations. The module never imports, so the file reports zero tests rather than failing partially.

Note the issue's own prose also named `npx vitest`; that turned out to be wrong. npm sets `npm_package_*` for anything it launches, `npx` included, so `npx vitest` was never affected. Measured on this tree: `npx` -> `1.6.1`, `pnpm exec` -> `undefined`, `node_modules/.bin/vitest` -> `undefined`. The pre-merge audit caught the claim after it had propagated into the code comment and the CHANGELOG; both are corrected here.

## Fix

One `??=` fallback alongside the three siblings already in `test/setup.ts`, keeping the same shape so a real value always wins:

```ts
process.env.npm_package_version ??= '0.0.0';
```

The surrounding comment now also covers this var's distinct trigger (no package.json script, as opposed to no `.env`).

## Verification

Reproduced red, then confirmed green with the exact command from the issue. Full Quality Gate on the combined diff: typecheck, lint, `pnpm test --run` (29 files, 155 passed / 1 skipped), Playwright (8 passed / 1 skipped), dev smoke (HTTP 200), and `pnpm build` all pass. The `HTMLCanvasElement.getContext()` console line is pre-existing and reproduces on an unmodified tree.

Closes #1021


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.GL2sa1IKlV

PR #1053, opened 2026-07-24
fix(hooks): the secret guard's remaining allowlist holes, and the audit handshake's ambient root

stevensacks:
Four defects surfaced by the audit of #1052, all pre-existing on `main`, all confirmed live before being fixed.

## Three holes in the secret guard's allowlist

Each is an arm that does not mean what it says. Verified against `main`'s hook, these all **passed**:

    API_KEY=<a>sk-live-9f3a1c4e8b7d2064<b>          # `<.+>` matches its own `>`
    API_KEY=example-sk-live-9f3a1c4e8b7d2064        # `^example` anchors nothing
    API_KEY=your-key-sk-live-9f3a1c4e8b7d2064       # `^your[-_]` likewise
    export API_KEY=sk-live-9f3a1c4e8b7d2064         # never scanned at all
    local API_KEY=sk-live-9f3a1c4e8b7d2064          # likewise

- `^<.+>$` carried the identical defect the command-substitution arm shed in #1052. Now `^<[^>]+>$`.
- `^your[-_]` and `^example` matched a **prefix**, so a secret rode through behind a placeholder-shaped lead-in. This was the widest hole in the allowlist and predates the substitution arm. Both are now anchored and length-bounded.
- The name grep required the suspicious name at line start, so a declaration keyword took the line out of the scan before any arm was consulted. `export`, `declare`, `local`, and `readonly` are now recognized.

The allowlist still reads shape rather than meaning, and the entry says so: a nested `$(… $(…))` is denied, a substitution that prints a secret is allowed, and a short enough secret behind a placeholder prefix still fits.

## The audit handshake's ambient root

Every Code Audit Team member derived its working root from the ambient cwd. Under worktree isolation that is the **main checkout**, not the tree the member just read, so its marker, trailer, status, and sidecar all described content it never reviewed. The usual symptom is a merge that will not clear; the worse case is a clearance that looks earned for an unreviewed tree.

Members now scope the whole handshake to the root named at dispatch, and the two hooks with no way to be told one — `audit-stamp-trailer.sh`, `post-audit-status.sh` — take `--root`. A single-checkout session is unaffected: there the two roots are the same directory.

## Verification

Every fix is pinned by tests that fail without it, checked by running the new tests against `main`'s scripts:

- 11 new cases in `block-secrets-write.bats` (29 total, all green)
- 5 new cases in `audit-stamp-trailer.bats` (24 total, all green)
- new `post-audit-status.bats` covering the option parse, 5 of its 6 RED against `main`

Full suites: `.gaia/tests/hooks/` 1047 pass / 0 fail, `.gaia/scripts/tests/` 800 pass / 0 fail, `shell-lint passed`.

## Known limits, not fixed here

Two further instances of the same ambient-cwd class, both distinct defects left for their own change:

- `pr-merge-audit-check.sh` resolves one root for **both** the marker store and the digest computation. The store is correctly main-anchored; the digest is not, so the gate digests whatever branch the main checkout holds rather than the tree being merged.
- `block-pr-create-unanswered-dist.sh` resolves the branch under review from the ambient cwd, so from a worktree it evaluates the main checkout's branch instead. It blocked this very PR with 14 files belonging to an unrelated branch.

## Type of change

- [x] Bug fix (patch)

## Checklist

- [x] Quality gate: no `.ts/.tsx/.js/.css` or gate-affecting config staged, so the gate skips per `wiki/decisions/Quality Gate.md`; `shell-lint` and every affected bats suite are green.
- [x] Updated `CHANGELOG.md` under `## [Unreleased]`.
- [x] Updated `wiki/concepts/Claude Hooks.md`, `wiki/concepts/PR Merge Workflow.md`, and `wiki/modules/Claude Integration.md`.
- [x] Touches no `/gaia-init` surface.


stevensacks:
Split. The secret-guard half is #1054, rebuilt clean off `main` and self-contained.

The audit-root-resolution half is withdrawn: it is not a bug but a property of the audit machinery, and it kept growing under review (two audit refusals, each finding a real defect in the previous round's repair). It is written up as an issue instead, with the verified evidence and the options.

The work-in-progress is preserved on `spike/audit-root-resolution` for reference. It is **unaudited and unreviewed** and should not be merged as-is.

PR #1054, opened 2026-07-24
fix(hooks): the secret guard's placeholder allowlist means "wholly", not "starts like"

stevensacks:
Three holes in `block-secrets-write.sh`'s dotenv allowlist, surfaced by the audit of #1052. All pre-existing on `main`, all confirmed live before being fixed.

Verified against `main`'s hook, every one of these **passed**:

    API_KEY=<a>sk-live-9f3a1c4e8b7d2064<b>          # `<.+>` matches its own `>`
    API_KEY=example-sk-live-9f3a1c4e8b7d2064        # `^example` anchors nothing
    API_KEY=your-key-sk-live-9f3a1c4e8b7d2064       # `^your[-_]` likewise
    export API_KEY=sk-live-9f3a1c4e8b7d2064         # never scanned at all
    local API_KEY=sk-live-9f3a1c4e8b7d2064          # likewise

## The fixes

- **`^<.+>$` → `^<[^>]+>$`.** The same delimiter-swallowing defect the command-substitution arm shed in #1052: `.` matches `>`, so anchoring both ends never meant "wholly".
- **The `your-` / `example` arms bound each segment, not the total length.** What separates a placeholder from a secret is structure, short words joined by punctuation versus one unbroken run. A length cap gets both directions wrong: it denies `your-github-personal-access-token` (a canonical `.env.example` line) while admitting `your-aB3xK9pQ7zR2wL5t` (a 17-char high-entropy secret). The separator between segments is mandatory on purpose; making it optional would let one unbroken run be read as several short ones.
- **The name grep tolerates a leading `export` / `declare` / `local` / `readonly`.** Previously the suspicious name had to start the line, so a declaration keyword took it out of the scan before any arm was consulted. This is the likeliest leak shape of the three.

Recognizing the declaration keywords pulls shell lines into a rule written for dotenv files, so the ordinary expansion forms are allowed as whole values alongside them: a positional, an expansion whose operand is **empty**, and an expansion followed by a path (`${1}`, `${VAR:-}`, `${ROOT}/dev.pem`). A non-empty default is not, because that is exactly where a real secret would sit and a segmented secret (a UUID, say) is shaped just like a placeholder.

The allowlist reads shape, never meaning, and the CHANGELOG says so. The honest limits are unchanged: a nested `$(… $(…))` is denied, a substitution that prints a secret is allowed, and a short enough secret behind a placeholder prefix still fits.

## Verification

- 42 cases in `block-secrets-write.bats`, up from 20. Every new case was run against the unfixed hook first and fails there.
- Two shapes an earlier draft of this fix let through are pinned as tests: a segmented secret in an expansion default, and a substitution whose body references a variable.
- I measured the real population before widening the name rule: **8 lines** in the entire repo match it, so the change costs nothing in practice rather than in theory.
- `.gaia/tests/hooks/` 1066 pass / 0 fail, `.gaia/scripts/tests/` 800 pass / 0 fail, `shell-lint passed`.

## Scope

This is the guard half of #1053, split out. The audit-root-resolution half of that PR is withdrawn and written up as an issue instead: it turned out to be a property of the whole audit machinery rather than a bug, and it does not belong in a secret-guard fix.

## Type of change

- [x] Bug fix (patch)

## Checklist

- [x] Quality gate: no `.ts/.tsx/.js/.css` or gate-affecting config staged, so it skips per `wiki/decisions/Quality Gate.md`; `shell-lint` and every affected bats suite are green.
- [x] Updated `CHANGELOG.md` under `## [Unreleased]`.
- [x] Updated `wiki/concepts/Claude Hooks.md` and `wiki/modules/Claude Integration.md` (the latter claimed the hook detects "high-entropy secret-like values", which it never has).
- [x] Touches no `/gaia-init` surface.


stevensacks:
Converting to draft. Not abandoned, and not the same situation as #1053: this branch is one coherent change, blocked on findings that are all fixable in place. Recording the state so the next person does not rediscover it.

**Correction to the first version of this comment.** It said the branch had "one identified defect, a two-line fix." That was wrong. The audit's full report has since been recovered: there are five Important findings, two of them secret-exposure, **three of them regressions this branch introduces against `main`**, plus documentation on this branch that is now known to be false. The scope below supersedes it.

## Current state

`code-audit-maintainer-shell` is the only member this diff dispatches. It ran against `534ddbb0` (tree `595d3cbc`) and **refused**, writing `.gaia/local/audit/6a780c02….code-audit-maintainer-shell.refused`. A refusal keyed to the current digest is absolute, so the merge gate stays shut until the digest rotates.

Its report did not reach the orchestrator and the findings sidecar carries only class and severity, which is what made this look opaque for three sessions. The full report was recovered verbatim from the subagent transcript and is posted below: https://github.com/gaia-react/gaia/pull/1054#issuecomment-5072061228 Background on why that recovery was needed: #1055.

Everything not listed below is sound. The member checked and confirmed: bash 3.2 clean, BSD/GNU portable, no `cd`, no hardcoded absolute paths, `set -euo pipefail` intact, `deny()` builds its JSON through `jq -n --arg` so the reason cannot inject, and the bats diff removes nothing, loosens nothing, and adds no `skip`. 42/42 pass, shellcheck clean on the hook.

## The five findings, three of them regressions this branch introduces

The distinction matters more than the count. Findings 2 and 3 are an incomplete fix and a test gap: bad, but the branch is no worse than `main` on either. Findings 1, 4 and 5 are different in kind. Each one changes a verdict that `main` gets right:

| # | Kind | base | HEAD |
| --- | --- | --- | --- |
| 1 | security loosening: a literal secret behind `${VAR}/` or `${VAR}.` | DENY | **ALLOW** |
| 4 | false positive: `export …_TOKEN="$VAR" # comment`, `local …_KEY="$1" \|\| true` | ALLOW | **DENY** |
| 5 | false positive: `local …_KEY="$1"`, `export …_KEY="${A}${B}"` | ALLOW | **DENY** |

So this branch, whose stated purpose is to tighten the allowlist, opens a hole `main` does not have, and hard-blocks four shapes of ordinary secret-free bash that `main` permits. That is the case for not merging it as authored, independent of the two findings where it merely falls short of its goal.

Every verdict was re-verified independently against the branch content, base compared with HEAD.

**1. The path arm admits arbitrary trailing text, so the commit is a net loosening.** `block-secrets-write.sh:113`, third alternative. After the closing `}`, one `/` or `.` unlocks an unbounded run of `[A-Za-z0-9._/-]`, which is the character set a literal secret uses.

| value assigned to a `_KEY` name | base | HEAD |
| --- | --- | --- |
| `${X}/sk-live-9f3a1c4e8b7d2064` | DENY | ALLOW |
| `${X}.sk-live-9f3a1c4e8b7d2064` | DENY | ALLOW |
| `${X}/550e8400-e29b-41d4-a716-446655440000` | DENY | ALLOW |

The comment at `:105-107` claims the arm "admits a path suffix without admitting arbitrary trailing text, which is how a secret would ride along." The behavior falsifies that sentence. Suggested repair: bound each segment the way the placeholder arms already do, e.g. `^\$\{[A-Za-z_][A-Za-z0-9_]*\}([/.][A-Za-z0-9_-]{1,12})+$`, which keeps `${ROOT}/dev.pem` and rejects the rows above.

**2. The declaration-keyword widening misses every flagged form.** `block-secrets-write.sh:117`. The group requires an identifier immediately after the keyword, so any option between them takes the line out of the scan entirely, which is the exact failure this commit exists to close. All four still ALLOW a live secret at HEAD: `declare -r`, `local -r`, `readonly -g`, `export --`. `local -r` and `declare -r` are the idiomatic forms in careful bash, so the headline claim is substantially unmet. Repair: allow flags in the group, e.g. `((export|declare|local|readonly)[[:space:]]+(-[A-Za-z-]+[[:space:]]+)*)?`, plus deny tests.

**3. The suite pins the path arm's allow direction only, which is what let finding 1 ship.** `block-secrets-write.bats:241-244`. Established by mutation testing: replacing `dev.pem` with a live secret in test 29 leaves the suite green. Every other new arm has a paired deny case. Repair: add a deny test for a secret behind the separator.

**4. A trailing shell comment or operator defeats the value extraction, denying secret-free lines.** `block-secrets-write.sh:59`. The sed strips quotes and surrounding whitespace but nothing past the value, so the rest of the line becomes the value and no arm matches. Both go base ALLOW to HEAD DENY: `export GITHUB_TOKEN="$GH_PAT" # for gh cli` and `local API_KEY="$1" || true`. A trailing comment on an `export` is about the most ordinary bash line there is, the decision is a hard block, and the reason tells the author to use environment variables, which is what the line already does.

**5. Two secret-free expansion shapes the widening now denies.** `block-secrets-write.sh:113`. Base ALLOW to HEAD DENY for `local API_KEY="$1"` (unbraced positional; the design comment claims positionals are covered but only the braced form was added) and `export API_KEY="${A}${B}"` (concatenated references, no literal at all, not mentioned in the design comment).

Not a finding, for contrast: `readonly SIGNING_KEY="${1:?usage}"` is also newly denied, but `:101-104` states the empty-operand-only rule and its reasoning explicitly. Acknowledged tradeoff, not a bug.

## Suggestions and cross-remit

- **S1**: 9 new `SC2016` shellcheck hits in the bats file. Deliberate, but the file already carries a justified file-wide disable for another rule; adding one for SC2016 clears the oracle output.
- **S2**: the `your|fake|dummy` arm admits a segmented secret, which the commit's own stated principle argues against. Low real-world risk, noted because two arms reason from the same premise to opposite conclusions.
- **Cross-remit, and this is the part that must not be lost**: `wiki/concepts/Claude Hooks.md:52` and `CHANGELOG.md:174` both state that a leading `export` / `declare` / `local` / `readonly` no longer takes the line out of the scan. Finding 2 proves that false for the idiomatic spellings. The CHANGELOG line is adopter-facing. Both also present the path arm as bounded, which finding 1 falsifies. Whatever happens to the code, those two claims cannot ship as written.

## Picking this up

1. Fix findings 1 and 2 (the secret-exposure pair), then 4 and 5 (the false-positive regressions).
2. Add the deny tests from finding 3, and the `declare -r` / `local -r` cases from finding 2.
3. Correct the CHANGELOG and wiki claims to match what the code actually does.
4. Commit. That rotates the member's digest and retires the refusal on its own, so no `--supersede-refusal` is needed.
5. Re-dispatch `code-audit-maintainer-shell` on the new HEAD.

One caveat found while verifying: the report's finding 1 lists `export API_KEY="${VAULT}/…"` as base DENY. It was base ALLOW, because base never scanned `export` lines at all. The other three rows hold and the finding stands.

If the re-dispatch refuses again and the report is once more unreadable, recover it from the subagent transcript rather than iterating blind:

```
jq -r 'select(.type=="assistant") | .message.content[]? | select(.type=="text") | .text' \
  ~/.claude/projects/<project>/<session>/subagents/agent-a<name>-<id>.jsonl
```

stevensacks:
## Recovered audit report, verbatim

`code-audit-maintainer-shell` against `534ddbb0` (tree `595d3cbc`), extracted from the subagent transcript. Unedited apart from this header. Absolute paths are the member's own.


## Summary

**Reviewed** (in-remit, both changed):
- `/Users/stevensacks/Development/gaia-react/gaia/.claude/hooks/block-secrets-write.sh`
- `/Users/stevensacks/Development/gaia-react/gaia/.gaia/tests/hooks/block-secrets-write.bats`

Out of remit and not reviewed: `CHANGELOG.md`, `wiki/concepts/Claude Hooks.md`, `wiki/modules/Claude Integration.md` (noted under Cross-remit).

**Verdict: marker withheld.** Five Important findings, no Critical. Correctness core is otherwise sound: bash 3.2 clean (no `declare -A`, no `mapfile`, no `${var^^}`), BSD/GNU portable (`sed -E` and `grep -E` only, no `grep -P`, no `sed -i`), no `cd`, no hardcoded absolute paths, `set -euo pipefail` intact, `deny()` builds its JSON with `jq -n --arg` so the reason cannot inject. The bats diff removes nothing, loosens nothing, and adds no `skip` (verified: zero `-` lines, zero `skip`). All 42 tests pass. `shellcheck` is clean on the hook.

Every claim below was verified by running the hook, not by reading the pattern. I compared each verdict against the base hook (`82b411e4`) to separate a regression from pre-existing behavior.

## Critical Issues

None.

## Important Issues (Should Fix)

### 1. The new path arm admits arbitrary trailing text, so the commit is a net loosening

- **Location**: `.claude/hooks/block-secrets-write.sh:113`, third alternative `^\$\{[A-Za-z_][A-Za-z0-9_]*\}[/.][A-Za-z0-9._/-]*$`
- **Issue**: after the closing `}`, one `/` or `.` unlocks an unbounded run of `[A-Za-z0-9._/-]`, which is exactly the character set a literal secret uses. These were **denied at base and are allowed at HEAD**:

  ```
  API_KEY=${X}/sk-live-9f3a1c4e8b7d2064                  base DENY -> HEAD ALLOW
  API_KEY=${X}.sk-live-9f3a1c4e8b7d2064                  base DENY -> HEAD ALLOW
  API_KEY=${X}/550e8400-e29b-41d4-a716-446655440000      base DENY -> HEAD ALLOW
  export API_KEY="${VAULT}/sk-live-9f3a1c4e8b7d2064"     base DENY -> HEAD ALLOW
  ```

  The adjacent pinned case `API_KEY=${X}550e8400-e29b-41d4-a716-446655440000` is denied (bats test 34). Adding one `.` flips it to allow. The comment at `:105-107` asserts the arm "admits a path suffix without admitting arbitrary trailing text, which is how a secret would ride along" — the behavior falsifies that sentence. This is the arm the dispatch flagged as never having had its own review pass.
- **Fix**: apply the same per-segment bound the placeholder arms use, e.g. `^\$\{[A-Za-z_][A-Za-z0-9_]*\}([/.][A-Za-z0-9_-]{1,12})+$`, which keeps `${ROOT}/dev.pem` and rejects `${X}/sk-live-9f3a1c4e8b7d2064` (the `9f3a1c4e8b7d2064` run exceeds the bound). If the arm is kept as-is deliberately, the comment at `:105-107` and the two docs must stop claiming it bounds the trailing text.

### 2. The declaration-keyword widening misses every flagged form

- **Location**: `.claude/hooks/block-secrets-write.sh:117`, `((export|declare|local|readonly)[[:space:]]+)?[A-Za-z_]`
- **Issue**: the group requires an identifier immediately after the keyword, so any option between them takes the line out of the scan entirely — the same failure the commit exists to close. Verified allowed at HEAD:

  ```
  declare -r API_KEY=sk-live-9f3a1c4e8b7d2064      ALLOW
  local -r   API_KEY=sk-live-9f3a1c4e8b7d2064      ALLOW
  readonly -g API_KEY=sk-live-9f3a1c4e8b7d2064     ALLOW
  export --  API_KEY=sk-live-9f3a1c4e8b7d2064      ALLOW
  ```

  `local -r` and `declare -r` are the idiomatic forms in careful bash, arguably more common than the bare keyword this commit does catch. The bats comment at `:149-153` states the keyword "took the line out of the scan entirely, and no allowlist arm was ever consulted" — still true for these.
- **Fix**: allow flags in the group, e.g. `((export|declare|local|readonly)[[:space:]]+(-[A-Za-z-]+[[:space:]]+)*)?`, and add deny tests for `declare -r` / `local -r`.

### 3. The suite pins the path arm's allow direction only, which is what let finding 1 ship

- **Location**: `.gaia/tests/hooks/block-secrets-write.bats:241-244`
- **Issue**: mutation-tested. Replacing `dev.pem` with a live secret in that test leaves the suite green:

  ```
  run_hook_write "$(printf 'readonly SIGNING_KEY=%s\n' '"${REPO_ROOT}/sk-live-9f3a1c4e8b7d2064"')"
  ok 29 an expansion followed by a literal path is allowed
  ```

  Every other new arm gets a paired deny case (11/12/13 for the placeholder arms, 31-34 for the operand arm). The path arm gets test 29 (allows) and test 34 (`${X}<literal>`, no separator) — neither exercises a secret *behind* the separator, which is the arm's entire failure mode. The assertions themselves are sound: I confirmed both helpers genuinely fail by flipping test 11's and test 23's fixtures, and only those two went `not ok`.
- **Fix**: add `@test "a literal secret behind a path separator is denied"` with `${X}/sk-live-9f3a1c4e8b7d2064` and `${X}.sk-live-9f3a1c4e8b7d2064`, alongside the finding-1 regex tightening.

### 4. A trailing shell comment or operator defeats the value extraction, denying secret-free lines

- **Location**: `.claude/hooks/block-secrets-write.sh:59` (extraction), reaching the deny at `:116`
- **Issue**: the sed strips quotes and surrounding whitespace but nothing past the value, so the whole remainder of the line becomes `$val` and no arm matches. Now that `:117` pulls shell lines in, these carry no literal secret, were **allowed at base, and are denied at HEAD**:

  ```
  export GITHUB_TOKEN="$GH_PAT" # for gh cli    base ALLOW -> HEAD DENY
  local API_KEY="$1" || true                    base ALLOW -> HEAD DENY
  ```

  A trailing `#` comment on an `export` is about the most ordinary bash line there is. `permissionDecision: "deny"` hard-blocks the Write/Edit, and the reason tells the author to "use environment variables / .env" — which is precisely what `"$GH_PAT"` already does, so the message is a dead end.
- **Fix**: strip a trailing ` #…` comment and a trailing `;`/`&&`/`||` clause before the allowlist runs, or match the value with a leading-anchored extraction rather than taking the rest of the line.

### 5. Two secret-free expansion shapes the widening now denies

- **Location**: `.claude/hooks/block-secrets-write.sh:113`, arm set
- **Issue**: both **allowed at base, denied at HEAD**:

  ```
  local API_KEY="$1"            base ALLOW -> HEAD DENY   (unbraced positional)
  export API_KEY="${A}${B}"     base ALLOW -> HEAD DENY   (concatenated references)
  ```

  The comment at `:97-98` says the arm set covers "a positional (`${1}`)" but only the braced form was added; `"$1"` is the far more common spelling and arm 1's bare-identifier alternative requires `[A-Za-z_]` as the first character. Concatenating two references contains no literal at all and is not mentioned in the design comment, so it reads as an oversight rather than a choice.
- **Fix**: add `^\$[0-9]$|^\$\{[0-9]+\}$` to the positional arm and an all-references alternative such as `^(\$\{[A-Za-z_][A-Za-z0-9_]*\})+$`.
- **Not a finding, for contrast**: `readonly SIGNING_KEY="${1:?usage}"` is also newly denied, but the comment at `:101-104` states the empty-operand-only rule and its reasoning explicitly. That one is an acknowledged tradeoff, not a bug.

## Suggestions

### S1. shellcheck SC2016 on the new fixtures (shellcheck)

- **Location**: `.gaia/tests/hooks/block-secrets-write.bats:222, 232, 237, 242, 255, 260, 269, 274, 283` (9 new hits; 10 more pre-existing)
- **Issue**: `SC2016 (info): Expressions don't expand in single quotes`. Deliberate here — the single quotes are what keep the fixture literal — but the suite already carries a file-wide `# shellcheck disable=SC2317` with a written justification at `:23-29`, and this warning is now the only noise in the file's oracle output.
- **Fix**: add a file-wide `# shellcheck disable=SC2016` beside the existing directive with the same one-paragraph rationale. Advisory; does not gate the marker.

### S2. The `your|fake|dummy` arm admits a segmented secret, which the commit's own principle argues against

- **Location**: `.claude/hooks/block-secrets-write.sh:90`
- **Issue**: `API_KEY=your-550e8400-e29b-41d4-a716-446655440000` and `API_KEY=example.550e8400.e29b41d4a716.446655440000` are allowed. The comment at `:102-104` reasons that "a segmented secret has the same structure a placeholder does" and uses that to require an *empty* operand on the expansion arm; the placeholder arms do not apply the same conclusion. Low real-world risk (nobody prefixes a live token with `your-`), and the code comment at `:86-88` does concede "a short enough secret behind a placeholder prefix still fits". Noting it only because the two arms reason from the same premise to opposite conclusions.

### S3. Pre-existing, out of scope

`.env.example:2` reads `SESSION_SECRET=local`, which this hook denies at base and at HEAD. Any Edit that re-emits that line is blocked. Not introduced by this diff; mentioned, not fixed.

## Cross-remit Findings

Both are direct consequences of findings 1 and 2 and become correct once those land. Recording them so they are not lost if the arms are kept as-is.

- **Location**: `wiki/concepts/Claude Hooks.md:52`
- **Issue**: states "A leading `export`, `declare`, `local`, or `readonly` does not take the line out of the scan" (false for `declare -r`, `local -r`, `readonly -g`, per finding 2), and lists "an expansion followed by a path (`${ROOT}/dev.pem`)" among the allowed whole-value shapes without noting the arm accepts any trailing text (finding 1). Its closing enumeration of "the honest limits" omits both.
- **Owner**: no Code Audit Team member's declared domain covers `wiki/**`.

- **Location**: `CHANGELOG.md:174`
- **Issue**: same two claims, adopter-facing. "a leading `export`, `declare`, `local`, or `readonly` took the line out of the scan entirely… the declaration keywords are recognized" overstates what the grep at `:117` matches.
- **Owner**: no member covers `CHANGELOG.md`.

## Self-heal

**Refused, by design.** No file was edited. The working tree is byte-identical to the tree I read (`595d3cbc3212f836a52f5ccbd2f03e63bbb719ab`, `git status --porcelain` empty). Every fix above is left to the authoring engineer: this is the audit's own gate machinery, and a rewrite here risks semantic drift with no independent reviewer downstream.

## Gate handshake

- **Marker**: refused. `/Users/stevensacks/Development/gaia-react/gaia/.gaia/local/audit/6a780c02b9f35122f34d697dc8bf1281160fb2b096aa3889c302686f08425c8a.code-audit-maintainer-shell.refused`
- **Stamp**: not called (applies only to a written earned marker).
- **Status**: not called (same).
- **Sidecar**: `/Users/stevensacks/Development/gaia-react/gaia/.gaia/local/audit/82b411e4b7e0e8e14f1c4272cd66b692d8eac63e.code-audit-maintainer-shell.findings.json`, six entries (five `warning`, one `suggestion`). Note: `.gaia/scripts/audit-key-lib.sh` does not exist on this branch, so I keyed the sidecar to the merge base the way `.gaia/scripts/post-findings-block.sh:55` documents (`<base-sha>.<member>.findings.json`), matching every existing sidecar in that directory.

> Audit marker NOT written. Address findings (or explicitly acknowledge the tradeoff), commit, and re-invoke this agent on the new HEAD.

Repairing any of findings 1-5 edits `.claude/hooks/block-secrets-write.sh`, which rotates my digest and retires the refusal with it — no `--supersede-refusal` will be needed.

stevensacks:
## Two repair rounds landed. One decision left, and it is yours.

Picking this up from the recovery comment above. Both rounds are pushed: `e2126202` and `c48c3ef7`.

### What is closed

All five Important findings, both Suggestions, and both cross-remit doc corrections from the recovered report:

| # | Finding | State |
| --- | --- | --- |
| 1 | path arm admitted arbitrary trailing text (net loosening) | closed, per-segment bound |
| 2 | declaration widening missed `declare -r`, `local -r`, `readonly -g`, `export --` | closed, keyword group accepts its options |
| 3 | suite pinned the path arm's allow direction only | closed, three deny tests added |
| 4 | trailing comment/operator defeated value extraction | closed, tail stripped before the value is read |
| 5 | unbraced positional and concatenated references denied | closed, both allowed as whole values |
| S1 | 9 new `SC2016` hits | closed, file-wide directive with rationale |
| S2/B | `typeset` missing from the keyword group | closed |
| cross-remit | `wiki/concepts/Claude Hooks.md` and the CHANGELOG entry both stated things the code did not do | corrected to match the code |

18 tests added across the two rounds, every one confirmed failing against the pre-fix hook before the fix landed. Suite is 55 cases, `.gaia/tests/hooks/` is 1068/1068, `shell-lint` clean.

I re-ran the base-vs-HEAD differential the audit used, and after round 1 the branch was strictly stronger than `main`: four new denies where the base allowed a real key, and not one remaining regression.

### What round 2 caught, and what I fixed

The re-audit refused again, on two findings that **round 1 itself introduced**. I verified both by running the two hooks rather than reading them, and both were real.

The first is fixed in `c48c3ef7`. The tail-strip added for finding 4 deleted a trailing comment or second statement before the allowlist ran, and nothing ever read the deleted text, so four shapes the base guard denied became allowed — a placeholder on the left with the real key parked in the comment, and worse in kind, a second assignment after `;` whose discarded text is executable. The tail still comes off the value, because the false positive it fixes is real, but it is now judged on its own first: denied when it carries a run of 13+ alphanumerics mixing letters and digits, the same structural rule the arms already use. Verified in both directions — an ordinary `# authentication for the gh cli` stays allowed.

### The open decision

The second one I did not fix, because every fix I could find is a call about this guard's security posture rather than a repair.

Teaching the name grep `export` / `declare` / `local` / `readonly` is the third of this PR's three fixes. It pulls shell lines into a rule written for dotenv **values**, and the value allowlist cannot keep up with shell's expression grammar. Nine realistic secret-free shapes now hard-block that `main` permits. All verified by running both hooks, not inferred:

| value assigned to a `_KEY`-family name | base | HEAD |
| --- | --- | --- |
| `"$REPO_ROOT/dev.pem"` | ALLOW | **DENY** |
| `"$A$B"` | ALLOW | **DENY** |
| `"$HOME/.config/key.pem"` | ALLOW | **DENY** |
| `"$1-$2"` | ALLOW | **DENY** |
| `"${A}-${B}"` | ALLOW | **DENY** |
| `"${KEY:-$DEFAULT}"` | ALLOW | **DENY** |
| `"${base}.${branch}"` | ALLOW | **DENY** |
| `"gaia-$RUNNER_OS"` | ALLOW | **DENY** |
| `"${FULL_KEY#Bearer }"` | ALLOW | **DENY** |

Each is a hard block whose stated remedy — "use environment variables" — is exactly what the line already does.

Three options. The third is my recommendation.

**(a) The audit's own proposed fix**: accept unbraced references in the all-references and path arms. I implemented and measured it: it closes **2 of the 9**. The other seven need separator handling, a leading-literal segment, and operand handling that it does not add. The finding is solid; its proposed fix is not sized to it.

**(b) A references-joined-by-short-segments arm.** I implemented and measured this too: it closes **all 9**, and it preserves every deny assertion that predates this PR. But it relaxes one deny added in this PR — `"${VAULT}/<a UUID>"` becomes allowed, because every segment clears the 12-char bound. That is a real change to what the guard permits. It is arguable, and the argument is Suggestion A's: the `your-` arm **already** allows `your-<the same UUID>`, so denying it behind `/` while allowing it behind `your-` is the inconsistency, not the rule. I did not take this unilaterally. A repair round quietly relaxing a deny is precisely what the audit gate exists to catch.

**(c) Split the PR. Recommended.** The `<…>` delimiter fix and the placeholder-segment fix are self-contained, correct, and audited. They close two holes that are live on `main` right now. The declaration-keyword widening is the piece dragging in shell's whole expression grammar, and it wants a value rule *designed* for shell rather than retrofitted onto arms written for dotenv literals. Landing the first two now stops holding two real security fixes hostage to a third that has cost three audit rounds and is not converging.

If you pick (b), say so and I will implement it and take the re-audit. If you pick (c), I will split the branch.

### One process note, so the next session does not pay for it

I spawned the member before checking for a live refusal at the current digest. There was one, keyed to exactly the content on disk, so the spawn bought a report that already existed. Cheap check, worth doing first:

```
ls .gaia/local/audit/ | grep "^$(bash -c 'source .claude/hooks/lib/audit-digest.sh && audit_member_digest "$PWD" code-audit-maintainer-shell')"
```

A refusal is proof the member already ran. Schema-3 markers carry `"sidecar":false`, meaning the report was never written to disk at all — which is why it had to be recovered from the transcript and posted here. Round 2's refusal is schema 3 as well, so this comment is again the report of record.

stevensacks:
## Split done, and this branch is now clean

Option (c). Head is `f089f091`; `GAIA-Audit` is green and `code-audit-maintainer-shell` has written its earned marker.

**This PR now carries exactly two fixes**, both self-contained and both tightening: the `<…>` arm no longer swallows its own closing bracket, and the `your-` / `example` arms bound each segment instead of matching a bare prefix.

**The declaration-keyword widening is withdrawn** to `fix/secret-guard-shell-declarations` at `c48c3ef7`, preserved in full with its tests and both audit-forced repair rounds, so the follow-up starts from a reviewed tree. The gap it was closing is now stated in four places rather than left to be rediscovered: the name grep, the guard's header, `wiki/concepts/Claude Hooks.md`, and the CHANGELOG entry.

### The audit caught something the split's own commit message got wrong

Worth recording, because it is the second time on this PR that a measurement claim turned out to be narrower than the diff.

The split commit asserted the branch "changes exactly three verdicts and all three are the same direction." That was measured against hand-picked probes covering the shapes I was reasoning about, not against the diff. The audit generated its own corpus and found **36 deny-to-allow flips**: `fake` and `dummy` had been added to the placeholder arm alongside `your` back in `534ddbb0`, and they were the diff's only loosening, with no test in a suite that pins every other arm in both directions. Deleting both keywords left all 29 tests green — which is exactly how a widening survives review.

It is also the wrong widening to make silently. A credential labelled `fake-` or `dummy-` is an ordinary live-secret shape: a rotated-but-still-valid key, or a real token on a throwaway account.

`fd22d360` drops both keywords and the matching wiki clause. The claim is now checkable rather than asserted:

| measurement | result |
| --- | --- |
| 424-value corpus, base vs HEAD (mine) | 0 deny-to-allow, 88 allow-to-deny |
| 97-value corpus (the audit's, independent) | 0 deny-to-allow, 24 allow-to-deny |
| 400,000-case fuzz (the audit's) | 0 deny-to-allow |
| structural containment (the audit's) | HEAD's allow-language is a strict subset of base's, so a flip is impossible rather than merely unobserved |

The audit also mutation-tested all 11 new tests: every one fails under at least one mutation of the arms it covers, so none is hollow. 29 cases in the suite, `.gaia/tests/hooks/` 1036 pass, `shell-lint` clean, verified under BSD grep as well as the local ugrep.

### Three advisory suggestions, none blocking, none actioned

- The unguarded `jq` at the payload read. I deferred it on scope grounds; the audit's judgment is that the deferral is right but my reason was the weaker one. There is no defect: only exit 2 blocks a `PreToolUse` call, so both failure modes degrade to a no-op, and **none** of the 15 `block-*.sh` deny-hooks guards `jq` — it is the house convention, documented at `wiki/concepts/Claude Hooks.md:106`. The real residual (a missing `jq` silently disables the scan) is family-wide and wants its own PR.
- `assert_allowed`'s trailing `!`-negation enforces only by position. Pre-existing, correct today, would silently retire if any line were appended after a call.
- SC2016 ×10 in the suite, all pre-existing, below the repo's own `warning` gate.

**Ready for you to undraft and merge.** I have not done either.

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.G7cFtqMp9k

PR #1060, opened 2026-07-26
feat(worktrees): GAIA resolves one main checkout and keys per-tree state by tree

stevensacks:
Makes GAIA work correctly inside a git worktree. Until now it half-worked: some surfaces silently resolved to the main checkout, some read a linked worktree as "not a repository", and per-tree state collided between trees because it was keyed by directory rather than by tree.

## The shape of the fix

**One resolver, one registry.** Every "where is the main checkout" derivation went through its own ad-hoc `git rev-parse`. They are now one shared resolver (`.gaia/scripts/main-root-lib.sh`), and every piece of local working state is declared once in `.gaia/state-registry.json` with an explicit scope: main-anchored (shared across trees) or per-tree (keyed by tree, not by path). The registry is what the janitor, the write guard, `block-rm-rf`, and `settings.json`'s permission allowlist all read, instead of each carrying its own drifting copy.

**Lifecycle moved to the harness.** GAIA no longer creates or removes worktrees; Claude Code does that natively. `create-worktree.sh` and `remove-worktree.sh` are gone. What GAIA still owns is provisioning on entry: a worktree installs its own dependencies and links its local state, rather than borrowing the clone's.

**Keys carry the branch.** The audit key, the PR-capture breadcrumb, and the worthiness ledger were single-valued across trees, so two worktrees on different branches clobbered each other. Each is now per-tree or per-branch.

**Guards that were silently inert.** Four wiki hooks read a linked worktree as "no repo" and stood down. Serena's `activate_project` could hand a worktree the main checkout's symbols. A write whose path pointed at another tree applied cleanly with no error. Each is now caught.

**Refusals that were silent.** `main-only-lib.sh` gives one loud refusal to the three flows that must run from the main checkout. Its own gate was dead for most of Milestone 5 because the refusal runs under the agent's zsh while the meter ran bats under bash; that false green is fixed and recorded.

## Verification

A 22-scenario concurrency meter (`.gaia/tests/concurrency/`) drives real worktrees rather than fixtures, and is armed as a required CI check against an expected-status manifest. Two of its scenarios were rewritten mid-program because they asserted their own fixtures instead of driving GAIA; that is called out in the log rather than quietly fixed.

## Distribution

The shipping harness was red from Milestone 2 to Milestone 8: this program created a shipping file with no manifest entry, and `cli-tests.yml`'s `distribution-harness` job is path-gated on paths that do not overlap the surfaces the leak check scans, so no PR ever ran the check that would have caught it. **That path-gating gap is not fixed here** — it is byte-identical on `main`, is not this program's, and will let the next leak through too.

The manifest is answered and green. Five files ship (four forced: a shipped surface already references them, so withholding would ship a dangling reference), one is withheld under category 1 (`check-wiki-state-collision.sh`: its only runners are the `.gaia/scripts/tests/` bats suite and `audit-ci-tests.yml`, both release-excluded, so an adopter can neither trigger it nor act on its failure). Five stale manifest entries dropped by regeneration.

- `gaia-maintainer release manifest --check`: clean
- staging harness: `leaks: none`, `runtime-dependency leaks: none`
- bare-mode `release runtime-deps`: clean

## Adopter impact

Four `## [Unreleased]` entries. The worktree scripts' removal is the one with an adopter-facing action.

## Type of change

- [x] Feature (minor)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.EAIdpnE17L

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.iQ26zSuTzn

PR #1061, opened 2026-07-28
feat(worktrees): Phase 2 of the worktree program, main-only refusal, statusline scoping, and audit-root resolution (#1055)

stevensacks:
Phase 2 of the worktree program (PLAN-012). Every ruling behind this change is already closed; this pull request builds them.

## What ships

1. **The main-only refusal on the eleven remaining classified flows.** A flow that writes, drives, or provisions state belonging to the main checkout refuses out loud from a linked worktree and names where to go. Plus the `file-tech-debt` sentinel write resolving the main root through the shared resolver, and one sentence of handoff prose.
2. **The statusline scopes its nudges.** Seven of eight nudges suppress outside the main checkout, `Run /setup-gaia` stays, and both background refreshers stop firing from a worktree. The kept nudge and the refusal compose: the nudge says setup is owed, the refusal says which tree to run it from.
3. **Eight documentation and test sites across seven files** that publish the opposite commitment, amended in the same commit as the suppression, so the repository never asserts the inverse of what it does.
4. **A meter that executes each refusal call site under zsh**, the shell those call sites really run in, and asserts placement ordering rather than mere presence.
5. **The remainder of #1055.** Option C at the audit entry points, option D at the two fail-open sites plus one named third, and a harness that drives the full handshake from a session whose own cwd is outside the audited root.

## Phases

- Phase 1: refusal and suppression, one atomic commit covering both waves.
- Phase 2: the meter and the audit-root systemic fix.
- Phase 3: the missed-site harness and the `## [Unreleased]` CHANGELOG entry.

Each phase lands only after its own gate is clean.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.LcwxOCy8g0

PR #1063, opened 2026-07-28
fix(audit): every member resolves the audited root before it decides what to review

stevensacks:
## The defect

All five `code-audit-*` agent definitions derive `AUDIT_ROOT` from the working root the orchestrator dispatches, falling back to the ambient toplevel only when none was supplied. In the four specialized members that derivation sat under **Gate handshake**, well below the **Remit and self-skip** block whose git queries resolve the diff base and list the changed files:

```bash
default_branch=$(git symbolic-ref --quiet refs/remotes/origin/HEAD ...)
BASE_SHA=$(git merge-base HEAD "origin/${default_branch}" ...)
changed=$(git diff --name-only "${BASE_SHA}...HEAD" ...)
```

Those queries decide **what the member reviews**, and they answered from the ambient cwd. Under a dispatch whose resolved root is a linked worktree while the subagent's cwd is the main checkout, a member reads one tree and keys its clearance to another.

Reported as a Suggestion by `code-audit-maintainer-shell` on #1061, and graded that way for good reason: the dominant sub-case fails closed (main sits at `origin/main`, the diff is empty, the member self-skips and writes nothing), and the adverse case needs main on a branch with a non-empty in-remit clean diff, which no current flow produces. It is latent rather than active. It is also the root-resolution instance the eight-stage census behind #1055 missed, and `.gaia/tests/hooks/audit-root-resolution.bats` cannot see it, since the harness drives the derivation rather than the prose ordering around it.

## The change

- The derivation moves above the remit block in the four specialized definitions, and its three git queries take `-C "$AUDIT_ROOT"`.
- `code-audit-frontend` already derives above its own remit call (its earliest root-consuming command is the dispatch-oracle call, which is already anchored), so it needs no move.
- The gate-handshake sections point back at the single derivation instead of repeating it, following the `BASE_SHA` precedent of deriving once in the remit block.
- Each definition states the rule that shell state does not persist between an agent's Bash calls, so every later call re-runs the derivation. The two consumers fail in opposite directions when it is skipped: `--root "$AUDIT_ROOT"` expands to `--root ""` and fails closed loudly, while `cd "$AUDIT_ROOT" && ...` fails open in silence, because `cd ""` returns 0. That was the second open finding from #1061; it rides here because it edits the same five files and would otherwise pay a second full digest rotation.
- **`code-audit-frontend`'s re-run-ledger derivation is anchored too.** `gaia_audit_key <base_sha> [<dir>]` defaults `<dir>` to `"."` and reads that tree's own branch, so from a non-checkout cwd the key names the wrong tree. The two writers also disagree on `--base`: `audit-write-findings.sh` requires it, `audit-write-clearance.sh` does not, so an empty `BASE_SHA` yields an earned marker with no findings sidecar beside it, the lost-report state the sidecar contract exists to prevent. The base resolution, the merge-base, the lib source, the key call and the ledger path now all take `$AUDIT_ROOT`, with the fail-open guard around the key preserved. The prose restatement under "How to run" mirrors it. Anchoring the script call with `cd "$AUDIT_ROOT" && ...`, the idiom this file already uses for the hooks, resolves it without changing `resolve-audit-base.sh`'s signature.
- The dead `audit_key` assignment and its `audit-key-lib.sh` source are deleted from the four specialized definitions. Nothing reads that value, both writers derive their own key from `--root` and `--base`, and left in place it computes a wrong-tree key and prints two stderr lines from a non-checkout cwd.
- `code-audit-frontend`'s citation of a `BASE_SHA` persistence comment is dropped. That comment exists only in the four siblings; the frontend file's own prose tells the reader those values travel forward without recomputation, so adding one would contradict it.
- **Everything `code-audit-frontend` writes beside its marker now lands in the tree that marker certifies.** Anchoring the ledger write alone left its clean-pass cleanup on the ambient cwd, so from a non-checkout cwd the removal matched nothing and a surviving ledger stayed valid and fresh into the next round, re-briefing the audit and the fixer with findings already closed. The cleanup now removes `$LEDGER`, which is what the prose already promises. The dispositions sidecar carried the same split: `audit-write-clearance.sh` builds `$marker` from `--root` and therefore absolute, while the sidecar paths beside it were relative and `audit-member-digest.sh` was invoked by a relative path even as it received `--root "$AUDIT_ROOT"`. Both are anchored.

## Verification

- The derivation precedes the first remit query in all five files (checked by line number, not by reading).
- The extracted derivation block still hashes `0a52dce9…` in all five, byte-identical to what shipped, so the stage 8 extractor reads what it always read.
- `bash .gaia/scripts/write-audit-remits.sh` reports all five `unchanged` and reverts nothing: every edit sits outside the `gaia:audit-remit` generated region.
- Gates: `pnpm typecheck` 0, `pnpm lint` 0, `.gaia/tests/hooks/` 1129/1129 with all three stage 8 tests green, `meter-gate.sh` reads `23 / 23` and every scenario matches `expected-status.txt`, `shell-lint.sh` clean.
- No machine-specific absolute path enters these template-distributed files.

## Expected audit cost

These files are audit machinery, so this rotates **every** member's content digest. Dispatch keys on ownership, not on digest rotation, and only `code-audit-maintainer-shell` owns `.claude/agents/code-audit-*.md`, so it is the single member owed at merge. The dispatch oracle confirms it.

## Not included

`code-audit-frontend`'s remaining ambient git sites stay as they are: the changed-file listing under "How to run" and its twin in the disposition gate (`git diff --name-only "$(.github/audit/resolve-audit-base.sh)"`), the `AUDIT_TREE_SHA` and `HEAD_SHA` captures, and the branch, upstream, and `git push` calls in the marker-write block. The listing is a mechanical anchoring, but the `git push` is not: it runs in the ambient tree while the stamp commit it means to push is created inside `$AUDIT_ROOT` by the subshell above it, so fixing it is a behavior decision rather than a path rewrite, and the whole group belongs in one deliberate pass over that file.

Four advisory findings are also left open, all graded non-gating by the audit and all belonging in that same pass:

- `code-audit-frontend.md:1031` guards the ledger cleanup on `[ -n "$AUDIT_KEY" ]` while the line it guards now removes `"$LEDGER"`. Every block that defines one defines the other, so no documented path reaches the mismatch, but the guard should name what it guards.
- `.gaia/tests/hooks/audit-root-resolution.bats:99` copies only `code-audit-frontend.md` into its stage 8 fixture, so it pins one of the five now-identical derivation blocks rather than all five. `ALL_MEMBERS` is already defined in that suite and the extractor takes a file argument, so looping the two positive tests needs no new helper.
- `.gaia/scripts/check-audit-key-callers.sh:49` asserts a bare `git grep -qF 'gaia_audit_key'`, which the four specialized files now satisfy on descriptive prose alone, since this PR deletes the only executable call they contained. Its first assertion still bites, so the guard is not hollow, but its header's stated guarantee no longer matches what it checks.
- `code-audit-maintainer-prose.md:173` spells the sidecar key `${audit_key}` where its three siblings spell `${AUDIT_KEY}`. Descriptive only, no consumer.

Also noted and not changed: the persistence paragraph added to all five definitions tells the reader to re-run the two derivation lines, which reproduces an *ambient fallback* rather than a *supplied* root, since `AUDIT_ROOT` is unset in a fresh shell. Every supported dispatcher routes through the merge-workflow template, which hands the literal assignment, and the failure is closed rather than silent, so the wording is imprecise rather than wrong.

### DIFF
--- a/.claude/agents/code-audit-frontend.md
+++ b/.claude/agents/code-audit-frontend.md
@@ -29,12 +29,14 @@ Your globs above are a **second precedence tier**: every claimant member's globs
 
 You are the Code Audit Team's **default member**.
 
-Resolve the audited root once, before the first handshake command below. The orchestrator dispatches you with a "Working root:" line and an `AUDIT_ROOT` assignment; that value is authoritative. The ambient toplevel is the fallback only when no working root was supplied.
+Resolve the audited root first, before the dispatch-oracle call below and every later root-consuming command. The orchestrator dispatches you with a "Working root:" line and an `AUDIT_ROOT` assignment; that value is authoritative. The ambient toplevel is the fallback only when no working root was supplied. It resolves here, ahead of the oracle, because that call decides whether you review at all: answered from the ambient cwd while your clearance keys to the supplied root, it reads one tree and certifies another.
 
 ```bash
 AUDIT_ROOT="${AUDIT_ROOT:-$(git rev-parse --show-toplevel)}"
 AUDIT_ROOT="$(git -C "$AUDIT_ROOT" rev-parse --show-toplevel)" || exit 1
 ```
+
+Shell state does NOT persist between an agent's Bash calls, the same rule the `BASE_SHA` derivation states for its own value, so every later call that uses `$AUDIT_ROOT` re-runs those two lines first. A call that skips them sees an empty value, and the two consumers fail in opposite directions: `--root "$AUDIT_ROOT"` expands to `--root ""` and fails closed loudly, while `cd "$AUDIT_ROOT" && ...` fails open in silence, because `cd ""` returns 0 and leaves the command running against whatever tree the session happens to sit in.
 
 Do not re-derive that set by hand. On a **local** run, at the start of every review, ask the dispatch oracle whether this diff dispatches you, with `--no-carry-forward`:
 
--- a/.claude/agents/code-audit-frontend.md
+++ b/.claude/agents/code-audit-frontend.md
@@ -36,7 +36,7 @@ AUDIT_ROOT="${AUDIT_ROOT:-$(git rev-parse --show-toplevel)}"
 AUDIT_ROOT="$(git -C "$AUDIT_ROOT" rev-parse --show-toplevel)" || exit 1
 ```
 
-Shell state does NOT persist between an agent's Bash calls, the same rule the `BASE_SHA` derivation states for its own value, so every later call that uses `$AUDIT_ROOT` re-runs those two lines first. A call that skips them sees an empty value, and the two consumers fail in opposite directions: `--root "$AUDIT_ROOT"` expands to `--root ""` and fails closed loudly, while `cd "$AUDIT_ROOT" && ...` fails open in silence, because `cd ""` returns 0 and leaves the command running against whatever tree the session happens to sit in.
+Shell state does NOT persist between an agent's Bash calls, so every later call that uses `$AUDIT_ROOT` re-runs those two lines first. A call that skips them sees an empty value, and the two consumers fail in opposite directions: `--root "$AUDIT_ROOT"` expands to `--root ""` and fails closed loudly, while `cd "$AUDIT_ROOT" && ...` fails open in silence, because `cd ""` returns 0 and leaves the command running against whatever tree the session happens to sit in.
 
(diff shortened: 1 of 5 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.gaiC7lLVVM

PR #1064, opened 2026-07-28
fix(audit): the frontend member's last ambient git sites answer from the audited root

stevensacks:
Follow-up to #1063, which deliberately declined this pass. Nothing here is owed to that pull request.

## The remaining ambient git sites

`code-audit-frontend.md` ran seven git sites against whatever tree the session sits in while its clearance keyed to the supplied root. All seven now anchor to `$AUDIT_ROOT`:

- the two changed-file queries that decide what the member reviews (the "How to run" bullet and its twin in the in-flight-fix promotion gate), anchored the way the ledger's base resolution already is
- the `AUDIT_TREE_SHA` and `HEAD_SHA` captures
- the branch probe, the upstream probe, and the push in the marker-write block

## The push is a behavior change, not a path rewrite

Step 2 creates the stamp commit inside `$AUDIT_ROOT`; step 3 probed and pushed ambiently. On a supplied root that differs from the session tree, `push.default=simple` sent the session tree's own branch to its own upstream, so the trailer stayed unpushed while `push_status` read `pushed` and the status POST targeted a head with no trailer. Both of the push's preconditions (an attached branch, an upstream) are properties of the audited tree, so they are read there.

The alternative considered and rejected was refusing to push on a foreign tree. It trades a wrong push for a guaranteed non-push, and it makes a worktree-dispatched member structurally unable to finish the mark, stamp, push, status sequence.

## The persistence paragraph, all five definitions

Two corrections, both in the paragraph that governs how a later Bash call reaches the audited root.

The paragraph told the reader to re-run the two derivation lines. In a fresh shell `AUDIT_ROOT` is unset, so that reproduces the ambient fallback rather than a supplied root, the exact failure the derivation exists to prevent. All five now say the dispatched `AUDIT_ROOT=` assignment is re-issued ahead of those lines when the orchestrator supplied one.

The paragraph also enumerated two consumers of an empty value and omitted `git -C "$AUDIT_ROOT"`, which this pull request makes the most common shape in `code-audit-frontend.md`. All five now name three consumers and say which reach silent ambient resolution:

- `--root ""` fails closed loudly.
- `git -C ""` exits 0 against whatever tree the session sits in, silently and regardless of shell. This is the shape the branch probe, upstream probe, and push all use.
- `cd ""` is shell-dependent: it returns 0 on bash 3.2 and runs the chain ambiently, while bash 5 prints `cd: null directory` and returns 1 so the chain never runs. Bash 5 is what CI and the local bats guard both use.

Editing all five costs no extra digest rotation: all five are machinery, so touching one rotates every member's digest, and dispatch keys on ownership.

## Advisory findings from the round-3 clean pass

- The clean-pass ledger cleanup guard now names `$LEDGER`, the path it removes, alongside the `$AUDIT_KEY` fail-open arm. Testing `$LEDGER` alone would have been a regression: an empty key still interpolates into a well-formed path, so the removal would fire in the one case the fail-open rule says to skip.
- Stage 8's two positive tests loop over all five definitions instead of pinning the default member's copy, so drift in any one block goes red. The mutation control stays on one file to bound runtime.
- `check-audit-key-callers.sh`'s header now states what assertion 2 verifies. Narrowing it to an executable call shape would turn the check red against this repo: four of the five definitions delegate keying to the sidecar writer and name `gaia_audit_key` only to say so, so a call-shape check would fail them for delegating correctly. Assertion 1 is the one that bites; assertion 2 corroborates.
- `code-audit-maintainer-prose.md` spells the sidecar key `${AUDIT_KEY}`, matching its three siblings.

## Verification

The three structural invariants hold after every edit to the five definitions: the derivation is the first root-consuming command in all five, the extracted block hashes `0a52dce9…` in all five, and the remit regions are byte-identical.

Four gates green on both commits: `pnpm typecheck && pnpm lint`; the hooks suite at 1129/1129 with zero failures; the meter at 23/23 with every scenario matching `expected-status.txt`; `shell-lint.sh` passed. `check-audit-key-callers.sh` reports 0/0, and its own conformance suite is green at 15/15.

`code-audit-maintainer-shell` is the whole dispatch set and cleared both heads with an earned marker, re-deriving the invariants and every behavioral claim in the consumer sentence independently.

## Not included

- **The FC-2 lede clause** stays frozen by explicit maintainer decision. It is inaccurate at fourteen sites, and changing it means overriding a closed ruling on contract wording that already shipped.
- **The clearance writers' ambient script path** (`code-audit-frontend.md:930`, `:943`, `:1086`, `:1100`, `:1123`, and the same shape in the other four definitions) is scoped to a follow-up. It needs a decision about whether resolving the writer from the session tree is deliberate, so a branch cannot supply its own clearance writer, rather than a mechanical anchoring.
- **The `call` / `derives` wording** in `check-audit-key-callers.sh`'s verdict lines and its conformance suite's test names is pinned verbatim by six assertions. The restated header documents the pin rather than rippling the rename.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.uFjjaJCnB1

PR #1065, opened 2026-07-28
fix(hooks): the secret guard reads shell declarations, not just dotenv lines

stevensacks:
Closes the leak shape `.claude/hooks/block-secrets-write.sh` has been advertising as open since #1054, whose own comment read "That is a real leak shape and it is deliberately NOT closed here … Tracked separately."

#1054 deliberately landed the narrowed half of this work: the whole-value anchoring and the per-segment placeholder bound. It deferred recognizing shell declaration keywords, because pulling shell lines into a rule written for dotenv values needs a value rule designed for shell expressions rather than the existing arms widened one shape at a time. This is that value rule.

## What changes

- **The scan reaches shell declarations.** A leading `export`, `declare`, `typeset`, `local`, or `readonly` no longer takes the line out of the scan, and neither do that keyword's own options, so the idiomatic `declare -r` and `local -r` are scanned like the bare forms. `typeset` is bash's synonym for `declare` and belongs with them.
- **Shell expansion forms are allowed as whole values**, since a declaration's value is a variable reference far more often than a dotenv literal is: a positional in either spelling, an expansion whose operand is empty, a value that is nothing but braced references, and an expansion followed by a path. A non-empty default is *not* allowed, because a default is exactly where a real secret sits and a segmented secret is shaped like a placeholder. The path arm bounds each segment by the same rule the placeholder arms use, so the separator buys a path rather than an unbounded tail.
- **A trailing comment or `;` / `&&` / `||` clause comes off the value before the allowlist reads it**, so an ordinary secret-free `export FOO_KEY="$BAR" # note` stops being hard-blocked by a deny that told its author to use the environment variable the line already used. The tail is judged on its own terms rather than discarded unread: a comment beside a placeholder is exactly where a real key gets parked.
- `fake-` and `dummy-` join `your-` as placeholder prefixes.

Roughly 300 lines of bats pin both the widened allowances and every deny the guard has to keep making.

## Notes for review

- Rebased onto current main. The commit that opened this line of work shipped as #1054, so it is dropped here rather than replayed; these two commits are what remains.
- `.claude/hooks/**` is audit machinery, so this rotates every Code Audit Team member's content digest and dispatches the full roster, not just the shell member.
- A CHANGELOG entry is owed and follows in a separate commit once this PR has a number. The existing #1054 entry stays as written, since it describes what #1054 actually shipped.

## Gates

    pnpm typecheck && pnpm lint                      clean
    bash .gaia/scripts/bats5.sh .gaia/tests/hooks/   1161 ok, 0 not ok
    bash .gaia/tests/concurrency/meter-gate.sh       23 / 23, every scenario matches
    bash .gaia/tests/shell-lint.sh                   passed


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.CwcvbUoCTy

PR #1066, opened 2026-07-28
fix(hooks): the secret guard's tail rescan keeps a guarded substitution whole

stevensacks:
Resolves the three NEW Suggestions from #1065's re-audit. All three land in files `code-audit-maintainer-shell` owns, so one re-dispatch covers them.

A Suggestion is a finding, not a specification, so each claim was reproduced against the pre-#1065 base before anything was written, and the proposed repair was checked before being adopted. Two of the three shipped as proposed; one did not.

## 1. The fragment rescan could still turn an allow into a deny

`.claude/hooks/block-secrets-write.sh`

The operator collapse ran over the whole tail, so a guard sitting inside a parked assignment's value was read as a statement separator. A line parking `export API_TOKEN=$(gh auth token 2>/dev/null || true)` after a `;` truncated to a fragment ending `$(gh auth token 2>/dev/null`, matched no allowlist arm, and hard-denied a line carrying no literal secret. Same defect class the untrimmed-first ordering already fixes for the primary value, relocated into the rescan, which had no equivalent bound.

The auditor's proposed repair (run the fragment's value through `value_allowed` untrimmed **before** the collapse) cannot be implemented as written: the fragments are produced *by* the collapse, so there is no fragment to pre-judge. The repair here targets the same intent one step earlier. An unnested `$(…)` masks to a body-free `$(@)` before `tr` ever sees the tail, so the fragment the allowlist reads is the value the author wrote. The mask carries the allowlist's own `[^)]*` bound rather than a greedy body, which would run to the last `)` and swallow an assignment parked between two substitutions.

The auditor's stated alternative (skip a collapse that would leave an unbalanced `$(`) is rejected deliberately, and there is now a test that fails against it: it drops the whole tail to the shape rule, so a literal parked beside a substitution stops being denied.

## 2. Nothing pinned the flag-survival property

`.gaia/tests/hooks/block-secrets-write.bats`

Added as proposed. Mutation-proved: a pipe-fed rewrite of the fragment loop passes every other test and fails exactly this one.

## 3. The header overstated the rescan's reach

Picked **qualify the header**, not **drop the `sep != "#"` gate**. The two are not equivalent, and the gate option does not deliver what the finding describes.

Dropping the gate alone is very nearly a no-op, confirmed with a mutant rather than assumed. The comment tail still opens with `#`, so it is still the fragment's leading token and `name_re`'s `^` anchor still rejects it. Of the auditor's own four examples, none changes verdict; only a contrived `# note ; API_TOKEN=…` does. Delivering the finding's intent would need the `#` stripped **and** `name_re` un-anchored for non-leading tokens, which is a real behavior change with real false-positive surface, for a hole the `secret_shaped` backstop already covers at every genuine secret shape (verified: all four positions still deny a real key).

So the header was the thing that was wrong. It, the two inline comments that repeated the claim, and the unreleased CHANGELOG entry now state the actual reach: the allowlist judges an executable tail only where the assignment leads its fragment.

## Differential probe

Every tracked line the guard's own grammar matches, base vs HEAD, plus a 27-case table:

- **Zero `allow -> deny` flips.** That was the bar.
- Four `deny -> allow` flips, all whole-value command substitutions the allowlist already admits in the primary position, so none can hide a literal that main would have caught.
- Tracked-line sweep: 9 lines, zero flips either direction.

## Carried forward: decided, so they stop recurring

Both are known and deliberate. Neither is a defect to be fixed later.

**A commit sha or UUID in a trailing comment is a hard block.** `export CACHE_KEY="$X" # baseline 3ea35f17…` is allow at base, deny now. **Accepted as-is.** It is fail-closed, specifically worded, and no tracked line trips it. If it ever becomes real noise the remedy is narrow: exempt a wholly-hex run of 7/8/40/64 characters. Not worth pre-empting a condition that has not occurred.

**`env` is outside the declaration-keyword group**, so `env FOO_KEY=<secret> cmd` still takes the assignment out of the scan. **Acknowledged as a genuine but out-of-scope gap.** It is the same "keyword hides the assignment" shape #1065 closed, one keyword short, and its option forms would already be covered by the existing `(-[A-Za-z-]+[[:space:]]+)*` group. It stays out of this PR because `env` is a command rather than a shell declaration, so admitting it is a fourth behavior change needing its own tests and its own CHANGELOG line, and this PR is bounded to the re-audit's three findings. Worth a dedicated change whenever the maintainer wants it.

## Tests

Five added, all mutation-proved against the specific rewrite each exists to catch (pipe-fed loop, greedy mask, skip-on-substitution, name-only deny). Tests that cannot fail pin nothing, which is the whole point of finding 2.

## Gates

- `pnpm typecheck` and `pnpm lint`: clean
- `.gaia/tests/hooks/`: **1172 ok, 0 not ok** (1167 on main, plus 5)
- `.gaia/tests/concurrency/meter-gate.sh`: **23 / 23**, every scenario matches `expected-status.txt`
- `.gaia/tests/shell-lint.sh`: passed

## CHANGELOG

Folded into #1065's existing `## [Unreleased]` entry rather than adding a second one. That entry has not shipped, so an adopter only ever sees the final state; a separate entry describing a fix to unreleased behavior would be noise.

---

## Audit rounds

Five rounds against `code-audit-maintainer-shell`. Rounds 3 and 4 withheld the marker; round 5 earned it clean. Every finding was reproduced with a mutant or a differential probe before being applied, and two rounds found real bugs I had introduced.

| Round | Verdict | What it found |
|---|---|---|
| 1 | clean | mask's `/g` unpinned; comment invariant over-claimed |
| 2 | clean | narrowed invariant still false (`<…>` arm admits `$(`) |
| 3 | **withheld** | **live `allow -> deny` flip**: masking erased the assignment that set the flag |
| 4 | **withheld** | **live `allow -> deny` flip**: flag pass was a bare pipeline, SIGPIPE under `pipefail` |
| 5 | clean | comment accuracy only; no live failure mode |

The two live bugs are worth naming, because both were introduced by this branch and both were missed by my own differential sweep:

- **Round 3.** Erasing a substitution body erases an assignment inside it, so a tail whose only watched assignment sat in a substitution lost its flag and fell through to the shape rule. The sweep missed it for lacking that *shape*. Fixed by reading the flag off the unmasked tail while the deny judgement stays on the masked fragments.
- **Round 4.** The flag pass was written as a bare pipeline. `grep -q` exits on first match, `tr` takes SIGPIPE, `pipefail` reports 141, the flag drops. Allow at 200 filler bytes, deny from ~20 000 up. The sweep missed it for lacking *length*; every fixture was short enough that `tr` finished first. Fixed with process substitution, the same idiom the deny loop above already uses and documents.

Final sweep: 55 cases including long-tail shapes, **zero `allow -> deny` flips**, 13 `deny -> allow`, tracked-line sweep clean. Hooks suite 1179 ok / 0 not ok.

## Accept-and-note (round 5, none gate)

Three comment-accuracy Suggestions, no live failure mode, recorded rather than folded in. The PR is clean and the member is marked, so folding them buys a full re-dispatch to re-earn a marker for prose.

- The mask is not strictly one-directional: `<$(a>` inside a bracket placeholder, with the closing paren past a separator, flips allow to deny. Requires syntactically invalid shell that would not survive `bash -n`, and the auditor found zero organic instances across 18,138 fuzzed payloads and the repo's own tracked lines.
- The header still says a tail's `$(…)` is "kept whole the way the primary value's is", while the detailed block correctly says the body is erased.
- The masked loop's flag set is subsumed by the unmasked pass; removing it changed zero verdicts across all 18,138 payloads.

## Out of scope: two pre-existing defects found on the way

Both confirmed identical on `main` and this branch, both outside this PR's changed lines. Neither is introduced here and neither gates.

- **`secret_shaped` has the same SIGPIPE shape and fails OPEN.** A comment carrying one `hunter2xyz123` denies; the same comment carrying 4000 of them allows. Wrong direction for a secret guard.
- **A guarded substitution in the PRIMARY value still false-denies when the line carries a tail.** A line reading `API_KEY=$(cat f || true) # note` denies on both sides. Same defect class this PR repairs for tails, unrepaired for primaries.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.03DOOPkdbB

Issue #1068, opened 2026-07-29
A guarded command substitution in the primary value still false-denies when the line carries a tail


stevensacks:
`.claude/hooks/block-secrets-write.sh:308`

## Failure mode

The untrimmed-first ordering earlier in the loop rescues a guarded command substitution only when nothing follows it on the line. When the line also carries a tail, that judgement fails, and the tail-strip then cuts the value at the substitution's own `||`:

```bash
val=$(trim_value "$(sed -E 's/[[:space:]]+#.*$//; s/[[:space:]]+([|][|]|&&|;).*$//' <<<"$rest")")
```

The truncated remainder matches no allowlist arm, so the write is denied and its author is told to use an environment variable by a line that already uses one.

Verified reproducers, DENIED both on `main` at `3593b06f` and at PR #1066's head: a write containing `API_KEY=$(cat f || true) # note`, and a write containing `GH_TOKEN=$(gh auth token 2>/dev/null || true); export GH_TOKEN`. The control with no tail, a write containing `export GH_TOKEN=$(gh auth token 2>/dev/null || true)`, is ALLOWED on both, which isolates the tail as the trigger.

This is the same defect class PR #1066 repaired for the TAIL position, left unrepaired for the PRIMARY one. False-positive/usability, not a security hole: it blocks legitimate writes rather than admitting secrets.

## Suggested fix

Apply the same `$(…)` mask #1066 introduced for the tail rescan before the strip, then judge the masked value. That PR's mask (`s/\$\([^)]+\)/$(@)/g`) and its rationale comment are the reference implementation, in this same file.

Two cautions from #1066's own five audit rounds, because this change is subtler than it looks. Erasing a substitution body also erases whatever the body held, so any flag or downstream read that depends on the body's contents must be taken from the UNMASKED value. And any new pipeline must not put a short-circuiting consumer at the end of a pipe under `pipefail`. Both mistakes shipped in that PR and both produced allow-to-deny flips that only a differential sweep against `main` caught.

Handler: prompt

stevensacks:
## Attempted and reverted, four approaches, none shippable

This was implemented on the #1069 branch alongside #1067, audited twice by `code-audit-maintainer-shell`, and reset off before merge. #1067 shipped alone. Recording what was tried so the next attempt does not repeat it.

[...]

PR #1069, opened 2026-07-29
fix(hooks): the secret guard's shape backstop no longer fails open on dense tails

### LINKED ISSUE #1067: secret_shaped() fails open under SIGPIPE, so the write guard's shape backstop skips the densest secret-shaped tails
`.claude/hooks/block-secrets-write.sh:101`

## Failure mode

`secret_shaped()` is a two-stage pipeline, and under `set -o pipefail` the pipeline's status is the function's return value:

```bash
grep -oE '[A-Za-z0-9]{13,}' <<<"$1" | grep -qE '[0-9].*[A-Za-z]|[A-Za-z].*[0-9]'
```

The consuming `grep -qE` exits on its first match. Whenever the producer still has output to write at that moment, it dies of SIGPIPE, the pipeline returns 141, and the function reports **not** secret-shaped. The shape backstop therefore declines to fire on exactly the tails most densely packed with secret-shaped runs.

Verified reproducer: a write whose content is `export API_KEY=$X # ` followed by a single `hunter2xyz123` is DENIED, while the same line carrying 4000 repetitions of that run is ALLOWED.

The direction is fail-OPEN, which is the wrong default for a secret guard. Confirmed identical on `main` at `3593b06f` and at PR #1066's head, so it is pre-existing and was not introduced there.

Scope note: this guard is a write-time heuristic, not a sandbox or a security boundary, and its header and wiki page already say so. This is a weakness in a best-effort developer guard, not an exploitable production vulnerability.

## Suggested fix

Feed the consumer by process substitution so only its status is read:

```bash
grep -qE '[0-9].*[A-Za-z]|[A-Za-z].*[0-9]' < <(grep -oE '[A-Za-z0-9]{13,}' <<<"$1")
```

PR #1066 repaired this identical SIGPIPE-under-`pipefail` shape in the tail flag pass, so there is a worked precedent in the same file, plus a regression-test pattern (a long-fixture bats case, which is the only way to observe it) already in `.gaia/tests/hooks/block-secrets-write.bats`.

Handler: prompt

stevensacks:
Fixes the secret guard's shape backstop, which fails OPEN.

Closes #1067.

**#1068 was attempted on this branch and dropped.** See "Why #1068 is not here" below; it stays open.

## The defect

`secret_shaped()` was a two-stage pipeline, and under `pipefail` the pipeline's status is the function's return value. Its consuming `grep -q` exits on the first match, so whenever the producer still had output to write it died of SIGPIPE, the pipeline returned 141, and the function reported **not** secret-shaped.

The backstop therefore declined to fire on exactly the tails densest with secret-shaped material. A comment tail carrying one 13-character run is denied; the same line carrying 4000 of them was allowed. The direction is fail-OPEN, the wrong default for a secret guard.

Feeding the consumer by process substitution reads only its status. This is the repair already applied to the tail flag pass in this same file, so there is a worked precedent one screen up.

## Verification

A differential probe compared every verdict against `main` under both `/bin/bash` (3.2.57) and bash 5.3, over 139 cases: guarded substitutions in primary and tail position, bracket placeholders and splices, comment and executable tails, parked assignments, brace groups and pipe prefixes, expansion forms, and 8k/14k/20k/40k/200k tails.

Exactly five verdicts flip, all allow-to-deny, all of them this repair:

| case | short twin on `main` | long form on `main` | at HEAD |
|---|---|---|---|
| secret-shaped comment tail | deny | **allow** at 20k / 40k / 200k | deny |
| secret-shaped separator tail | deny | **allow** at 20k | deny |
| secret-shaped substitution tail | deny | **allow** at 40k | deny |

Each flip has a sub-threshold twin that `main` already denies, so this closes a length-dependent hole rather than widening the rule. The threshold sits between 14k (already denies on `main`) and 20k, which matches the SIGPIPE class. Nothing flips deny-to-allow. A sweep of all 15 unfiltered tracked assignment lines in the repo reports zero flips.

The regression test needs a long fixture; the short twin passes either way, so it pins nothing on its own. Mutation-tested: restoring the bare pipeline kills the new test and only the new test, with the other 79 still green.

## Why #1068 is not here

#1068 was implemented, audited twice, and reset off this branch. It is a false-positive fix, and every attempt at it introduced a defect worse than the one it removed:

1. The mask the ticket suggests fixes only one of its two reproducers. The second carries a `;` with no leading whitespace, which the strip never sees.
2. Widening the strip to accept a bare separator fixed both, but concealed `$(a)||<a-literal>`. Rejected before it left the working tree.
3. Widening the tail *extraction* instead kept that literal denied, and shipped. Audit round 1 found it false-denied any `<…>` placeholder holding its own separator (`<user;pass> # fill in`); following that up found it also false-denied `$(x; export FOO_KEY=$Y) # note`.
4. Reverting the extraction and reading what the bare strip discards fixed those. Audit round 2 found two **Critical fail-opens** in that mechanism: the discard guard skips the pre-tail region entirely (`API_KEY="";<a-literal> # note` allows), and the discard is read from the bracket-masked copy so a wrapped literal is invisible to it (`API_KEY=""||<a-literal>` allows). Both verified against `main`, both with no already-allowed equivalent.

Three of those four rounds gave my differential probe a false green, each time because the probe lacked a *shape* rather than because the reasoning was wrong. Closing round 2's second finding needs a length-preserving mask or an offset map, which is a materially larger mechanism than the ticket describes.

Trading a fail-open in a secret guard for a false-positive fix is the wrong direction, so #1068 comes off and #1067 ships alone. The accumulated reproducers and the dead-end analysis are worth attaching to #1068 before anyone picks it up again.

## Gates

- `pnpm typecheck && pnpm lint`: clean
- `bash .gaia/scripts/bats5.sh .gaia/tests/hooks/`: 1180 ok, 0 not ok (1179 on `main`, +1)
- `bash .gaia/tests/concurrency/meter-gate.sh`: 23 / 23, every scenario matches `expected-status.txt`

CHANGELOG folds into the existing `## [Unreleased]` entry that already credits #1065 and #1066, rather than opening a third.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.cLBcJSCQZb

PR #1070, opened 2026-07-29
fix(hooks): the secret guard exempts .env.example from the assignment rule

stevensacks:
Makes `.env.example` editable again, by dropping the secret guard's *placeholder allowlist* for that file and putting the guard's *shape test* in its place.

## The problem

`.env.example:2` is `SESSION_SECRET` assigned the value `local`. Five letters, so it matches no placeholder arm, and the write guard denies it:

```
BLOCKED: write contains a non-placeholder secret assignment
```

The file could not be written back unchanged, and an adopter adding a variable to their own `.env.example` was told to use a gitignored `.env` instead, which is backwards for the one file whose purpose is to carry placeholder assignments. The line is already committed, so nothing is leaking; this is a false positive, not an exposure.

GAIA has already decided this question twice, on the same file, in the same direction:

| guard | `.env.example` | matcher |
|---|---|---|
| `block-env-read.sh:67` | exempt | Read |
| `block-env-write.sh:15` | exempt, by basename | `Edit\|Write\|MultiEdit` |
| `block-secrets-write.sh` | **not exempt** | `Edit\|Write\|MultiEdit` |

## The fix

Rule 4 keeps running on `.env.example`; it just swaps which test decides. The allowlist asks *"is this a recognized placeholder"*, which every honest example value fails. Shape asks *"is this an unbroken run of 13+ alphanumerics mixing letters and digits"*, which every one of them passes and a pasted key does not.

| value | allowlist | shape |
|---|---|---|
| `local`, `development`, `3001`, `http://localhost:3001/api/`, `changeme`, `your-api-key-here` | refused | **allowed** |
| `sk-live-9f3a1c4e8b7d2064`, `aB3xK9pQ7zR2wL5t`, a JWT | refused | **refused** |

Checks 1-3 (AWS key id, GitHub PAT, PEM header) run on the file as everywhere, and the match is that exact basename, so `.env`, `.env.local`, and `.env.example.local` reach the full allowlist as before.

**Accepted cost, stated rather than buried:** an ordinary value carrying a long unbroken alphanumeric run, a hostname like `myapp123456789.example.com`, is refused in that file. This is the shape rule's own documented limit, applied here deliberately.

The hook reads `file_path` for the first time. Every tool on this matcher carries one; a payload without one resolves to empty, matches no exemption, and is scanned in full. That fail-closed direction is what leaves the 1180 pre-existing tests untouched, and it is pinned by a test.

## Review history

The first commit lifted rule 4 wholesale. **The audit pushed back and was right**: that let a live `sk-live-…` clear into a *tracked* file, and the commit's own test enshrined it. The second commit replaces the wholesale exemption with the shape backstop and inverts that test. Also from the audit: `basename --` (a dash-leading path printed a usage error to stderr; `block-env-read.sh:65` already writes `--`), and `.gaia/hook-scopes.json:80` now describes the `file_path` read, with its `scope: any` verdict unchanged and the manifest checker green on all four arms.

The alternative I rejected up front, and still reject: widening the placeholder allowlist to admit short all-letter values. That solves a file-level problem with a value-level rule, so `DB_PASSWORD=secret` would start clearing in real source. A mutant taking that route is killed by four tests here, so the *choice* of fix is pinned, not only its behavior.

## Verification

TDD: the allow-tests were written first and failed against the unmodified hook; the deny-tests passed from the start, making them regression guards rather than restatements.

Every added test kills a targeted mutant, each built and run:

| mutant | dies to |
|---|---|
| no exemption at all | the six `.env.example` allow-tests |
| exemption hoisted above checks 1-3 | the AWS / PAT / PEM tests |
| wholesale exemption, shape backstop dropped | the three secret-shaped deny-tests |
| keep the allowlist instead of shape | the six allow-tests |
| prefix match instead of exact basename | `.env.example.local` is not exempt |
| widen to the whole `.env*` family | `.env` is not exempt |
| never read `file_path` | the six allow-tests |
| fail OPEN on a missing path | 42 tests, incl. `no file_path is still scanned` |
| `basename` without `--` | the dash-leading-path test |
| widen the placeholder allowlist instead | `the same short value outside .env.example is still denied` |

## Gates

- `pnpm typecheck && pnpm lint`: clean
- `bash .gaia/scripts/bats5.sh .gaia/tests/hooks/`: 1197 ok, 0 not ok (1180 on `main`, +17)
- `bash .gaia/tests/concurrency/meter-gate.sh`: 23 / 23, every scenario matches `expected-status.txt`
- `bash .gaia/scripts/check-hook-scope-manifest.sh`: rc=0, all four arms


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.2OUVRfiKvv

PR #1079, opened 2026-07-29
debt(ci): run the sandbox bats tree in CI behind its own paths filter

### LINKED ISSUE #1076: the .gaia/tests/sandbox bats tree runs in no CI workflow and no pre-commit hook, so a red assertion there gates nothing
`.gaia/tests/sandbox/run-all.sh:12`

## Failure mode

`.gaia/tests/sandbox/` holds 21 tests across five suites, and no CI workflow and no pre-commit hook runs any of them. The only runner is `pnpm test:sandbox` (`package.json`), invoked by hand.

```
doc-setup-gaia.bats           10 tests
wiki-page.bats                 6 tests
spec-028-composition.bats      2 tests
manifest-and-enable.bats       2 tests
operational-enforcement.bats   1 test
```

`run-all.sh:12` documents the wiring it never received:

```bash
# CI: add the following step before this script in your workflow:
#   - run: bash .gaia/tests/sandbox/run-all.sh
```

The bad outcome is that a failing assertion in this tree reds nothing and can sit on `main` indefinitely. That is not hypothetical: `spec-028-composition.bats:15` has asserted a permission rule form that was deliberately removed since `1ac0ece5` (#783), and the resulting `not ok` was noticed only when a Code Audit Team member read the file for unrelated reasons. Ownership was not the gap, since `.gaia/**/*.bats` is owned by `code-audit-maintainer-shell`; the gap is that nothing executes the suite.

The workaround is already load-bearing. `.gaia/tests/lib/doc-setup-gaia-isolation.bats:9` records it in its own header:

> Landed in `.gaia/tests/lib/` (not `.gaia/tests/sandbox/`, which no workflow runs) so `audit-ci-tests.yml` actually gates it.

So a maintainer who wanted a doc-conformance test gated had to file it outside the tree that exists for doc conformance. Every future author faces the same choice, and the sandbox tree accretes coverage that runs nowhere.

## Suggested fix

Wire the tree into `.github/workflows/audit-ci-tests.yml` the way every other gated suite is wired: a `dorny/paths-filter` entry plus a bats step.

The filter needs per-suite attention rather than a single glob, for the reason the workflow's existing comments repeatedly name: a filter that omits a path a suite reads reports `code=false`, skips every bats step, and greens the job having run zero tests. Each suite's read set has to be enumerated, including `.claude/commands/setup-gaia.md`, `.claude/settings.json`, `.claude/hooks/block-env-read.sh`, and the wiki page `wiki-page.bats` reads.

Two notes for whoever takes it:

- Sequence this after the `spec-028-composition.bats:15` fix. Wiring the suite in while that assertion stands turns the new CI job red on arrival.
- Portability looks clean. The only environment guard in the tree is a `command -v curl` skip in `wiki-page.bats:21`, and the suites are static conformance greps rather than stateful harness runs.

A side effect worth having: `.claude/settings.json` currently appears in no member's audit globs and in no CI paths filter, so nothing automated gates it at all. Adding it to this filter gives it one.

Precedent for both the shape and the fix is #837, the same defect class for `.specify/extensions/gaia/test/**/*.bats`. That one needed an ownership half as well; this one does not.

## Out of scope

Whether the 21 tests are individually worth keeping. This issue is about the tree running at all, not about its contents.

Handler: prompt

stevensacks:
Closes #1076.

`.gaia/tests/sandbox/` ran in no CI workflow and no pre-commit hook. A failing
assertion there gated nothing, which is how a stale permission-rule assertion
survived in that tree until a reader opened the file for unrelated reasons.

## What this does

Wires the tree into `audit-ci-tests.yml` through `run-all.sh`, gated on a
**second** `dorny/paths-filter` output rather than the existing `code` one.

The separate output is about cost. `manifest-and-enable.bats` git-greps every
committed `.json`/`.md`/`.yml`/`.yaml` file in the repo, so an honest read set
for it is those four globs. Folding them into `code` would arm this job's two
pnpm installs and the INV-7 concurrency meter on very nearly every PR. As its
own output it arms only the bats step, which needs no toolchain beyond the apt
install already shared.

Every suite's read set is enumerated per suite with the reason beside it, since
a filter that omits a path a suite reads reports `code=false`, skips every bats
step, and greens the job having run zero tests. `wiki/concepts/OS Sandbox.md`
matched no existing entry at all.

Verified rather than assumed: `dorny/paths-filter` at the pinned SHA matches with
picomatch `{dot: true}` (`MatchOptions`, `src/filter.ts`), so one glob per type
covers `.claude/`, `.gaia/` and `.github/` and also matches zero segments for
root-level `package.json` and `CHANGELOG.md`.

## Two suite changes were required first

**The manifest-diff assertion is deleted, not repaired.** It needed
`origin/main`, which a depth-1 `pull_request` checkout does not fetch, so it
failed outright in CI. Repair would not have made it correct:

- it asserted the diff never touches a release-generated file that ordinary
  feature work regenerates (105 commits touch it);
- it contradicted the `Distribution Audit (PR)` required check, whose error text
  instructs the author to commit exactly the regenerated manifest it forbade.

It held on the one feature branch that introduced it and nowhere since. Its
durable sibling, pinning that no committed config carries a raw enable, is kept
and now runs in CI.

**The docs-link reachability check skips under CI**, and stays live by hand.
Reaching a third-party host from a declared-required context means an upstream
docs reorganization returning 404 blocks every merge in the repo, which is a
disproportionate blast radius for a dead documentation link.

## Verification

Ran the step's verbatim command inside a faithful reproduction of the checkout
the workflow produces (`actions/checkout` v5, no `fetch-depth`, PR merge ref, no
`refs/remotes/origin/main`) with `CI=true`:

```
step exit code : 0    (expected 0)
ok             : 20   (expected 20)
not ok         : 0    (expected 0)
skipped        : 2    (expected 2)
```

Both skips are by design: the OS-level enforcement test needs a sandbox-capable
runner, and the docs-link check skips under CI.

Gates on this branch:

| Gate | Result |
| --- | --- |
| `.gaia/tests/hooks/` | 1216 ok, 0 not ok |
| `.gaia/scripts/tests/` | 1097 ok, 0 not ok |
| `.gaia/tests/sandbox/` | 20 ok, 0 not ok |
| `meter-gate.sh` | 23 / 23, every scenario matches `expected-status.txt` |
| `shell-lint.sh` | passed |
| Quality Gate | skips, no staged source or gate-affecting config |

## Also

`.gaia/tests/README.md` now states per tree whether CI gates it or a human runs
it, covering all eleven rather than four. That distinction is the thing whose
absence let this defect sit unnoticed.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.rsM4gooskS

PR #1081, opened 2026-07-29
debt(tests): the npx bats fallback names the package that exists

### LINKED ISSUE #1080: the npx bats fallback in the forensics and lib runners names a package that does not exist, so it 404s instead of falling back
`.gaia/tests/forensics/run-all.sh:25`

## Failure mode

Two bats runners advertise an npx fallback that cannot work, and two READMEs instruct the same command as a supported install route.

The npm package is `bats`. There is no `bats-core` package published to npm: the project and the Homebrew formula are both named bats-core, which is where the wrong spelling comes from, but `registry.npmjs.org/bats-core` returns HTTP 404 while `registry.npmjs.org/bats` returns 200 (`1.13.0`, bin `bats`).

So on a machine with `npx` but no `bats` on `PATH`, the runner announces the fallback and then `exec`s a command that exits with an npm E404. Because it is an `exec`, that status becomes the script's, and the suite never runs. The fallback is dead code in the failure case it exists to serve.

Affected, all four surfaces:

```
.gaia/tests/forensics/run-all.sh:10   header, advertises the fallback
.gaia/tests/forensics/run-all.sh:22   echo announcing it
.gaia/tests/forensics/run-all.sh:25   the exec that 404s
.gaia/tests/lib/run-all.sh:10         header, advertises the fallback
.gaia/tests/lib/run-all.sh:23         echo announcing it
.gaia/tests/lib/run-all.sh:26         the exec that 404s
.gaia/tests/forensics/README.md:74    instructs the command
.gaia/tests/lib/README.md:55          instructs the command
```

Both READMEs word it as `Any platform: npx -y bats-core@latest (the run-all.sh entrypoint falls back to this)`. Both halves of that sentence are false: the command 404s, and the entrypoint does not successfully fall back to it.

Blast radius is bounded. CI installs bats with apt before any step that invokes these runners, so no workflow is affected and no gate is weakened. What breaks is a hand run on a machine that lacks bats, which is a contributor on a fresh checkout, and the failure is a confusing npm error rather than the documented fallback.

The identical defect in `.gaia/tests/sandbox/run-all.sh` was fixed in #1079 and verified end to end: with the name corrected, a run on a `PATH` carrying node but no bats downloads bats and executes the whole suite.

## Suggested fix

Replace `bats-core@latest` with `bats@latest` at all six script locations, and correct the two README sentences so they name the working package and stop claiming a fallback that does not happen.

Scope the fix to all four files. Repairing the scripts alone leaves both READMEs instructing a command that 404s, which is the more visible half of the defect.

Worth carrying across: the sandbox fix added a short note naming why the spelling is wrong (npm package `bats`, project and brew formula `bats-core`), which is what stops the next author reintroducing it.

## Out of scope

The two runners' other pre-existing quirk, that the on-`PATH` loop aborts at the first failing file under `set -e` while the npx path runs every suite in one invocation. Both exit non-zero, so no gate is weakened; it is a diagnostics difference only.

Handler: prompt

stevensacks:
Closes #1080.

The forensics and lib bats runners advertise `npx -y bats-core@latest` as the fallback for a machine without bats on `PATH`, and both READMEs instruct the same command as a supported install route. No `bats-core` package is published to npm. The project and the Homebrew formula are both named bats-core, which is where the wrong spelling keeps coming from.

Because the fallback is an `exec`, the E404 status becomes the script's own and the suite never runs, so the branch is dead code in the one case it exists to serve.

## What changed

Six script sites and two README sentences move from `bats-core@latest` to `bats@latest`, across all four files. Repairing the scripts alone would have left both READMEs instructing a command that 404s, which is the more visible half.

Each runner also picks up the short note the sandbox runner already carries, naming why the two spellings differ. That note is what stops the next author reintroducing the wrong one. The `ERROR:` text for the no-npx case stops naming bats-core too and now points at the header, which lists all three install routes rather than the one whose name is ambiguous.

The READMEs keep their claim that the entrypoint falls back to this command. That half of the sentence was false only because the spelling was broken, and it becomes true once the runners are fixed.

## Verification

Run on a `PATH` carrying node but no bats, with both controls, rather than reasoned about:

| check | expected | actual |
| --- | --- | --- |
| `command -v bats` under the probe `PATH` | MISSING | MISSING |
| negative control, `npx -y bats-core@latest --version` | rc != 0, npm E404 | rc=1, `npm error 404` |
| positive control, `npx -y bats@latest --version` | rc=0 | rc=0, `Bats 1.13.0` |
| `.gaia/tests/forensics/run-all.sh` | takes npx branch, rc=0 | takes npx branch, rc=0, **112 ok / 0 not ok** |
| `.gaia/tests/lib/run-all.sh` | takes npx branch, rc=0 | takes npx branch, rc=0, **360 ok / 0 not ok** |

The precondition assertions are part of the probe on purpose: with bats still reachable, both runners take the on-`PATH` loop, pass, and prove nothing about the branch under test.

Registry, checked directly: `registry.npmjs.org/bats-core` returns 404; `registry.npmjs.org/bats` returns `1.13.0` with bin `bats`.

## Gates

- `.gaia/tests/shell-lint.sh` passed
- `.gaia/scripts/bats5.sh .gaia/tests/hooks/` reads 1216 ok / 0 not ok
- `.gaia/tests/concurrency/meter-gate.sh` reads 23 / 23, every scenario matches `expected-status.txt`
- Quality Gate skips: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config

## Blast radius

Bounded to a hand run on a fresh checkout. CI apt-installs bats before any step that invokes these runners, so no workflow is affected and no gate is weakened. All four files are release-excluded (`.gaia/release-exclude:88`) and absent from the manifest, so no adopter clone receives them.

## Out of scope

The two runners' other pre-existing quirk, that the on-`PATH` loop aborts at the first failing file under `set -e` while the npx path runs every suite in one invocation. Both exit non-zero, so no gate is weakened; it is a diagnostics difference only.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Z1cK8hcVy9

PR #1085, opened 2026-07-30
fix(audit): the local audit path posts its status on the pushed head

### LINKED ISSUE #1084: the secret guard's cap commentary carries three inaccuracies the round-8 audit verified
Three verified inaccuracies in the commentary around the secret-scanning write guard's judged-material caps. The round-8 `code-audit-maintainer-shell` audit on #1082 surfaced all three, graded each a Suggestion, and verified each by measurement or mutant. They were accepted-and-noted at merge time rather than folded in, because the PR was already clean and marked and folding them in would have bought a full re-dispatch of that member solely to re-earn its marker.

All three are comment or test-name only. No behavior changes, and none of them can affect what the guard admits or denies.

## Locations

This issue carries one machine dedup key, anchored on finding 1. Findings 2 and 3 are cited below with their literal `path:line` so the keyless `path:line` fallback covers them too, and a later audit re-finding either one resolves here instead of filing a duplicate.

- `.claude/hooks/block-secrets-write.sh:313` (finding 1, the key's anchor)
- `.gaia/tests/hooks/block-secrets-write.bats:1129` (finding 2)
- `.gaia/tests/hooks/block-secrets-write.bats:1111` (finding 3)

## Finding 1: the worst-payload sentence reads distributively against a cap that is a total

`.claude/hooks/block-secrets-write.sh:313`

**Failure mode.** The size cap is a total over all matching lines: the guard joins them and compares the joined length. The passage describes the worst admitted payload as 200 lines "each packed with minimal-width `;`-separated fragments to just under the size cap". Read distributively that is 200 x 65536 = 13,107,200 characters of matching material, which the size cap denies outright, so the sentence contradicts its own subject ("The worst payload either cap admits"). A maintainer sizing a future `timeout` against this ceiling, which is the passage's stated purpose, reads a worst case inflated 200-fold.

The passage's own arithmetic shows the collective reading is the intended one: 65536 x 1.4 ms/char = 91.8 s against its stated ~90 s, and 65536 x 0.6 = 39322 spawns against its stated ~40000. Both are one size-cap's worth, not 200 of them.

**Suggested fix.** Replace `each packed` with `together packed`. This restates no figure, so it does not reintroduce the newly-precise prose that three consecutive audit rounds each falsified.

**Note on urgency:** the error runs in the fail-safe direction. A distributive reader over-estimates the worst case and would set a longer timeout or none, not a shorter one.

## Finding 2: the at-cap fixture's name claims a last-character pin its uniform value cannot make

`.gaia/tests/hooks/block-secrets-write.bats:1129`

**Failure mode.** The fixture's value is a uniform 65530-character run of the letter `a` following the assignment name, and every prefix of that run is equally non-placeholder, so a judge-a-prefix-then-pass-the-rest regression leaves the test green. Its name mirrors the line-cap sibling `a write at the matching-line cap is judged to its last line`, which does pin that property because the only secret sits on the last line. The two names promise the same guarantee and only one delivers it. Position is what makes it unachievable: at exactly the cap there is nothing beyond the cap to drop, so no character-truncation regression is observable there.

Verified by mutant: against a hook that judges only the first 100 matching lines and passes the rest, the size-cap fixture stays DENY (survives) while the line-cap sibling flips DENY to ALLOW (killed).

**The fixture is not redundant and should not be deleted.** An at-cap wave-through mutant, one that passes a payload of exactly 65536 characters unjudged while leaving the over-cap deny intact, is killed by this fixture alone. The at-cap allow, the 65537 over-cap case, and the line-cap fixtures all stay green under it, and no other fixture in the suite carries exactly 65536 characters of matching material. Only the name overreaches.

**Suggested fix.** Rename to what it pins: `a write at the judged-size cap is judged rather than waved through`.

## Finding 3: the new fixture sits between the halves of the pair its comment describes

`.gaia/tests/hooks/block-secrets-write.bats:1111`

**Failure mode.** The note beginning `The pair below pins the cap's own boundary` introduces the 65536-allow / 65537-deny pair as the two sides that keep the comparison from widening to `-ge`. The at-cap fixture added in #1082 now sits between those two halves, so the two tests immediately below the comment are the at-cap allow and the at-cap deny, a pair for which the comment's stated reason is not the operative one. A reader taking "the pair below" as the next two tests gets the wrong pair.

**Suggested fix.** Move the at-cap block together with its own comment below the 65537 fixture, restoring adjacency for the described pair. Code motion only: no prose edited and no figure restated.

## This should ride the next substantive change, not earn its own branch

Both cited paths are owned by `code-audit-maintainer-shell`, so any touch of either already pays that member's audit round. These three fixes add no marginal audit cost when they ride one, and cost a full round (~110-140k tokens, 20-35 minutes) if they travel alone. Prefer folding them into the next real change to the guard or its suite over scheduling them on their own.

## Declined as settled, do not re-raise

A fourth finding from the same round was declined, and it is recorded here because it has already cost one audit round.

Both cap deny messages end with `Split the write, or keep the real values in a gitignored .env`, and the caps deliberately run ahead of the `.env.example` basename branch, so a large `.env.example` write receives that advice too. Round 8 proposed either dropping the clause or gating it on the basename. Round 7 had already considered and rejected both of those exact alternatives.

The word `real` is what keeps the advice correct: it tells the author where real values belong, which is true guidance even when the file being written is the placeholder file. The phrasing also matches the shape-only deny that `.env.example` already emits. Leave both messages as they are.

Handler: prompt

stevensacks:
## The defect

The `GAIA-Audit` success status was stranded on a superseded commit whenever the last Code Audit Team member to clear was a specialized one.

Each member writes its marker, stamps the trailer commit **locally** (`audit-stamp-trailer.sh` never pushes), then posts the status on the **pushed** PR head (`post-audit-status.sh` resolves it via `gh pr view --json headRefOid`). `code-audit-frontend` already pushed the stamp before posting. The four specialized members did not, and stated the omission as doctrine:

> You never push, here or anywhere else: … and the member-aware status call in step 3 clears independently via the remote head.

That last clause is the bug. It clears via the **pre-stamp** remote head. The orchestrator then pushes the stamp, the PR head advances, and the success status is left on a sha nothing reads. CI's local-mode stand-down posts `pending` on the new head, `GAIA-Audit` is a required check, and a queued merge waits indefinitely while `gh pr checks` reports the PR as having no checks at all.

Because the member-aware gates in both hooks decline until every dispatched member has cleared, whichever member clears **last** is the one that both lands the stamp and posts the status:

| Dispatched set | Outcome before this change |
|---|---|
| frontend only | frontend pushes first, status lands. Correct. |
| specialist(s) only, i.e. every maintainer-only diff | strands, deterministically |
| mixed | coin flip on which member finishes last |

## The fix

1. **The four specialized members gain the push step** `code-audit-frontend` already had, so all five push the stamp commit before posting. The handshake is now sidecar, mark, stamp, push, status in all five, with steps renumbered and cross-references updated. The self-heal prohibition is preserved and made explicit: the only commit a member ever pushes is the content-preserving trailer stamp, never a repair.

2. **`post-audit-status.sh` gains the deterministic backstop** so prose drift cannot silently re-strand it. The existing guard compares the audited **tree** against the pushed head's tree, which is blind to the stamp **by construction** because the stamp is content-preserving. It now also requires local HEAD's **commit** to be the pushed head, declining `stamp not pushed` otherwise. The tree guard and its decline reason are untouched: content divergence and an unpushed stamp are different diagnoses.

Notably this makes real a fail-safe `code-audit-frontend.md` has always promised but which never fired, since an amend preserves the tree:

> on the amend path there is nothing new to push, and the status helper declines `audited tree not on pushed head` until the operator pushes, which is the correct fail-safe

3. **The hook gains its first bats suite** (`.gaia/tests/hooks/` had none). It pins the new arm plus the two fallback paths that must not misfire: no-PR-no-upstream, where `head_sha` *is* local HEAD, and CI's detached HEAD at the PR head sha.

4. **Two wiki pages** carried statements this exposes as inaccurate. Both attributed the trailer stamp to `code-audit-frontend` when any member can land it, and `Code Audit Team` asserted a maintainer member never pushes a commit.

## Riding along

Closes #1084 (three verified secret-guard comment and name inaccuracies). It was filed to ride the next change paying a `code-audit-maintainer-shell` round rather than earn its own branch, and this branch pays that round. Its finding 2 admits only the rename, no fixture restructure, so the `difficulty:easy` grade held.

## Verification

- `post-audit-status.bats` 9/9, `audit-stamp-trailer.bats`, `pr-merge-audit-check.bats`, `block-secrets-write.bats` 126/126: zero failures across all four
- `shellcheck` clean on both modified hooks
- wiki-style audit greps clean; no absolute paths introduced under `.claude/`
- Quality Gate has nothing to check (no `.ts/.tsx/.js/.jsx/.mjs/.cjs/.css` or gate-affecting config), confirmed by pre-commit reporting no changed files for lint-staged


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.fSSMtNdJAE

PR #1098, opened 2026-07-31
fix(audit): all five members key their findings to one review base

stevensacks:
## The defect

The four specialized Code Audit Team members derived `BASE_SHA` as a bare `merge-base` against the default branch. Nothing else in the system uses that value.

| Consumer | Base derivation |
| --- | --- |
| `code-audit-frontend` | `merge-base "$(resolve-audit-base.sh)" HEAD` |
| The four specialists | `merge-base HEAD origin/<default>` |
| Merge-time findings-block reader | `merge-base "$(resolve-audit-base.sh)" HEAD` |

`gaia_audit_key` is `<base-sha>.<branch-slug>` and co-dispatched members share a branch, so the keys agree exactly when the base shas do. They diverge after any round that stamped a clean trailer.

Two consequences:

1. **Correctness.** `post-findings-block.sh` globs `"${audit_dir}/${AUDIT_KEY}".*.findings.json` under the resolver-derived key. It found the frontend's sidecar and silently missed all four specialists'. On any PR that had a clean stamped round, the consolidated findings block dropped four members' findings with no error, and the "one ledger serves the whole dispatched set" invariant had quietly become two ledgers.
2. **Cost.** The specialists re-reviewed the whole PR diff on rounds where the frontend reviewed only the delta.

Reproduced before any edit: on a scratch repo with a stamped clean round plus one non-machinery commit, the two derivations produced different keys, with the frontend reviewing 1 commit and the specialists 4.

## The fix

All five members resolve the review base through `.github/audit/resolve-audit-base.sh`.

Each specialist keeps the whole-PR fork point as `FULL_BASE` for **one** decision: the self-skip arm. Membership is resolved over the whole PR diff (`resolve-audit-members.sh`, and the "Full-PR scope (load-bearing)" note in `gate-pending-members.sh`), and a member's marker is invalid at HEAD exactly when its content digest rotated. The owned-file case is safe on the increment alone. The machinery case normally is too, because RT-006 resets to full scope, but that check fails open when the classifier libs will not load. In that arm the increment carries only the machinery file, so a member that does not own it would self-skip, write no marker, and deadlock the merge against a gate still demanding one.

Collapsing the two bases back into one value is the way this regresses.

Each member also gains a cross-file paragraph in its own domain, since a narrower increment can begin after a commit that already cleared: sourcers, `bash` callers and stdout-pinning bats suites for shell; exports, flags, emitted JSON, render templates and snapshots for node; `uses:` references and `needs.*.outputs` reads for workflows; cross-referencing instruction files for prose.

## Guards

`.gaia/scripts/check-audit-base-derivation.sh` plus its suite, in the dual-mode shape of `check-audit-key-callers.sh` and `check-main-root-derivation.sh`. Over `.claude/agents/`:

1. No review base comes from a bare `merge-base` against the default branch. `FULL_BASE` is exempt **by name**: the assignment owning a `merge-base` call is the nearest identifier to its left, so the self-skip derivation is distinguished from a drifted review derivation rather than blanket-allowed.
2. Every file naming `BASE_SHA` also names `resolve-audit-base.sh`.

`.gaia/scripts/tests/audit-base-agreement.bats` executes the five **real** derivation snippets, extracted from the definitions themselves rather than restated, against a scratch repo carrying a stamped clean round:

- base and audit-key agreement across all five, with a non-vacuity assertion that the resolver actually advanced past the whole-PR fork point
- `post-findings-block.sh` reading a sidecar written under the key a specialist itself derives
- the scope-versus-membership separation, per specialist, under a machinery-only increment with the classifier libs unloadable, asked of the real roster classifier rather than a copy of the remit globs

## Verification

Every fixture was written red-first and each guard mutated per mechanism:

| Probe | Red before | Green after |
| --- | --- | --- |
| key agreement | specialists resolved a different base | all five agree |
| findings block | `findings: declined: no sidecars` | 1 member, finding present |
| self-skip separation | increment carried the owned file | separation holds for all four |
| guard assertion 1 | 4 bare derivations | 0 |
| guard assertion 2 | 4 files | 0 |

Mutations on the **fixed** tree, each independently:

- revert one member to a bare `merge-base` → assertion 1 reds alone (assertion 2 stays 0)
- strip every mention of the resolver from one member, keeping `BASE_SHA` → assertion 2 reds alone (assertion 1 stays 0)
- remove the `FULL_BASE` separation from one member → both deadlock probes red; the key-agreement and findings-block probes stay green, which is the separation being proven by its own fixture rather than piggybacking on another

Full CI-gated sweep after the last edit: `.gaia/scripts/tests/`, `.github/audit/tests/`, `.gaia/tests/hooks/`, `.gaia/tests/lib/`, `.gaia/tests/prose-audit/` under bash 5 (2940 tests, 0 failures) plus `bash .gaia/tests/shell-lint.sh` (passed).

CI wiring confirmed rather than assumed: `.gaia/scripts/**` arms the `code:` filter, the step runs `bats .gaia/scripts/tests/` wholesale, and every source the new suite reads (`.claude/agents/code-audit-*.md`, `.github/audit/**`, `.claude/hooks/**`, `.gaia/audit-ci.yml`) is already a filter entry.

## Known cost

Agent definitions are gate machinery, so this rotates every member's digest and RT-006 resets the round to full scope. One full-price full-team round, once.

## Out of scope

Park-by-default, per-member marker scanning, and the specialists' three-dot `"${BASE_SHA}...HEAD"` versus the frontend's two-dot form. The last is a real pre-existing gap and is filing-worthy as tech debt.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.juTkRD1K1n

PR #1103, opened 2026-07-31
debt(audit): the BASE_REF exemption is a whole identifier, and its limits are stated (#1101, #1102)

### LINKED ISSUE #1101: check-audit-base-derivation.sh: the BASE_REF exemption matches a substring, so an aliased variable escapes assertion 1
## The `BASE_REF` exemption matches a substring, so an aliased variable escapes assertion 1

`.gaia/scripts/check-audit-base-derivation.sh:166`

```awk
if (right ~ /^[^|;&#)`]*BASE_REF/) {
```

Assertion 1 exempts a `merge-base` call that takes `BASE_REF` inside its own argument list, because that is the resolver-derived shape the check exists to require. The test matches `BASE_REF` as a bare substring, so it also accepts the token as the **tail of a longer identifier**.

### Failure mode

An agent definition drifting to the pre-fix bare derivation, but routing it through an aliased variable whose name ends in `BASE_REF`, is exempted and the check reports `0` at exit 0:

```bash
DEFAULT_BASE_REF="origin/${default_branch}"
BASE_SHA=$(git -C "$AUDIT_ROOT" merge-base HEAD "$DEFAULT_BASE_REF")
```

That is exactly the derivation assertion 1 rejects, spelled so the exemption fires. Two related shapes escape the same way: `${BASE_REF:+origin/main}`, which names the token and then discards it, and `BASE_REF` appearing as a redirection target or here-string body rather than as an argument.

Not live. No definition currently contains such an alias, and no realistic drift spelling contains the literal string `BASE_REF`, which is why this is filed rather than fixed in place.

### Suggested fix

Require a non-identifier character (or start-of-string) before the token:

```awk
if (right ~ /^([^|;&#)`]*[^A-Za-z0-9_])?BASE_REF/) {
```

Verified by the auditing member on a scratch copy: the aliased line reds, the real repo still reports `0`/`0` at exit 0, and every existing test in both suites stays green. Add a fixture beside `DRIFT_BASE_REF_AFTER_CLOSE` in `.gaia/scripts/tests/check-audit-base-derivation.bats` pinning the aliased form, and mutate the boundary to confirm the new fixture reds on its own.

### Context

Surfaced during the round-5 audit of PR #1098, which brought the exemption's scoping to its current form (per-call, terminated at `)` or a backtick). The member's round-5 argument for why that scoping is now structurally closed rather than merely narrower holds for call *termination*; this is the separate axis of token *identity*, which that argument does not cover.

Handler: prompt

### LINKED ISSUE #1102: check-audit-base-derivation.sh: acknowledged-limits note omits the nearest drift assertion 1 cannot catch
## The acknowledged-limits note omits the nearest drift assertion 1 does not catch

`.gaia/scripts/check-audit-base-derivation.sh:104`

The header paragraph beginning "It is not unconditional" names one shape the check cannot see, a base derived without a literal `merge-base` at all:

```
# (`BASE_SHA=$(git rev-parse "origin/${default_branch}")`) is never a
# candidate and this check does not see it. The behavioural suite next door
# is what covers that, by executing the real fences.
```

It omits a nearer and more likely one.

### Failure mode

Keep the two-step resolver shape verbatim and change only where `BASE_REF` comes from:

```bash
BASE_REF="origin/${default_branch}"
BASE_SHA="$(git -C "$AUDIT_ROOT" merge-base "${BASE_REF}" HEAD 2>/dev/null || true)"
```

Assertion 1 exempts the call, correctly by its own rule, because `BASE_REF` genuinely is an argument to it. Assertion 2 passes, because the file still names `resolve-audit-base.sh` in its prose. Yet the member has fully reverted to the pre-fix bare derivation.

This is a documentation defect, not a code one. Assertion 1's charter text at the top of the file ("The review base must come from `.github/audit/resolve-audit-base.sh`") reads as a stronger guarantee than what the assertion enforces, and a reader auditing the check's coverage would not learn this limit from the paragraph that exists to state exactly such limits.

### The backstop is real

`.gaia/scripts/tests/audit-base-agreement.bats` catches this drift by executing the real fences. The auditing member verified it rather than asserting it: applying the mutation above to `code-audit-maintainer-shell.md` on a scratch copy reds tests 4, 5 and 6 of that suite. So no coverage is missing, only its documentation.

### Suggested fix

Name that spelling in the same paragraph and delegate it to the behavioural suite, exactly as the `rev-parse` sentence beside it already does. Documentation only, no code or test change.

### Context

Surfaced during the round-5 audit of PR #1098. Filed rather than folded in because the reporting member is already marked for this content, so editing the file rotates its content digest and buys a full re-dispatch (60-110k tokens) solely to re-earn a clearance already held, per the digest-economics rule in `wiki/concepts/PR Merge Workflow.md`.

Handler: prompt

stevensacks:
Batched because both edit `.gaia/scripts/check-audit-base-derivation.sh`, and any edit there rotates `code-audit-maintainer-shell`'s content digest. Fixing one costs the same single audit round as fixing both; there is no cheaper subset.

## #1101, the exemption matched a substring

Assertion 1 exempts a `merge-base` call that takes `BASE_REF` inside its own argument list, because that is the resolver-derived shape the check requires. The test matched `BASE_REF` as a bare substring, so it was also satisfied by the token as the **tail of a longer identifier**, and a drift routed through an alias exempted itself.

The issue calls this not live. It is nearer than that: `GITHUB_BASE_REF` is an ordinary CI variable naming a branch, not an invented one, so `merge-base HEAD "origin/$GITHUB_BASE_REF"` is a spelling a drifting author could plausibly reach for. That is the fixture.

The boundary is a **two-term** class, not the single term the issue proposes. The issue's `[^A-Za-z0-9_]` closes the alias escape but is a *superset of the stop set*, so the prefix group can end **on** a stop character and hand the exemption straight back. That was caught by the round-1 audit and is fixed in the second commit; the shape matrix below is the final state:

| `right` shape | on main | issue's regex | shipped |
| --- | --- | --- | --- |
| real `"${BASE_REF}"` | exempt | exempt | exempt |
| real `"$BASE_REF"` (code span, unbraced) | exempt | exempt | exempt |
| line that ends at the token | exempt | exempt | exempt |
| every pre-existing drift fixture | red | red | red |
| `"$DEFAULT_BASE_REF"`, `"${DEFAULT_BASE_REF}"` | **exempt** | red | red |
| `"origin/$GITHUB_BASE_REF"` | **exempt** | red | red |
| `main)BASE_REF`, and the other five stop chars | red | **exempt** | red |

The mirrored **right** edge (`${BASE_REF_OLD}`) is knowingly left open, and the comment says so with an accurate reason: `BASE_REF([^A-Za-z0-9_]|$)` is plain ERE and works (verified), so the construct is available; it is unused because no drift spelling builds a name by suffixing the token, where `GITHUB_BASE_REF` makes the prefix side live.

## #1102, documentation only

The acknowledged-limits paragraph named one uncovered shape (a base derived with no `merge-base` at all) and omitted a nearer one: keep the two-step resolver shape verbatim and change only where `BASE_REF` comes from. Assertion 1 exempts that call correctly by its own rule, assertion 2 passes because the prose still names the resolver, and yet the member has fully reverted to the pre-fix bare derivation. The paragraph now names it and delegates it to the behavioural suite, exactly as the `rev-parse` sentence beside it already does.

Both halves of that new claim are verified, not asserted. Applying the drift to `code-audit-maintainer-shell.md` on a doctored tree:

- the static check reports a clean `0` / `0` at exit 0, which is the limit being documented;
- `audit-base-agreement.bats` reds tests 4, 5 and 6, which is the backstop being pointed at.

## Verification

- `check-audit-base-derivation.bats`: 26/26 (3 new).
- `audit-base-agreement.bats`: 7/7, unchanged.
- `bash .gaia/tests/shell-lint.sh`: clean.
- Quality Gate: skipped by its own rule; no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.
- Re-run in full after the last edit on each round, including prose-only edits.

**Mutation proof, per mechanism.** The boundary class carries two, so it gets two mutants and each lands on exactly one test:

| mutant | reds | stays green |
| --- | --- | --- |
| drop the identifier-char exclusion | alias test | abutting-paren, canonical, converted-file, real-repo |
| drop the stop-set subtraction | abutting-paren test | alias, canonical, converted-file, real-repo |

Three fixtures pin the boundary from every side, because a rule that closed an escape by retiring the exemption entirely would also have passed a one-sided test: the aliased escape reds, the stop-character escape reds, and the canonical `"${BASE_REF}"` / `"$BASE_REF"` spellings are asserted still exempt.

Closes #1101
Closes #1102


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.PR31n2XMRl

PR #1104, opened 2026-07-31
debt(audit): check-audit-key-callers fails closed on a root it cannot scan (#1100)

### LINKED ISSUE #1100: check-audit-key-callers.sh returns a clean 0/0 for a tree it never scanned (non-repo or subdirectory root)
## `check-audit-key-callers.sh` reports a clean verdict for a tree it never scanned

`.gaia/scripts/check-audit-key-callers.sh:91`

`gaia_check_audit_key_callers` takes `repo_root` as a required argument and goes straight to `git grep` without ever confirming the path is a git repository root:

```bash
literal_matches="$(git -C "$repo_root" grep -nIE "$GAIA_AUDIT_KEY_BAD_LITERAL_PATTERN" -- '.claude/agents/' 2>/dev/null)"
```

### Failure mode

A `git grep` that cannot run produces exactly what a `git grep` that ran and matched nothing produces: an empty string. Both assertions then print `0`, and the function returns 0, a clean bill of health for a tree it never read.

Two inputs reach it:

- **A path outside any repository.** `git grep` fails, stderr is discarded by `2>/dev/null`, the variable is empty.
- **A subdirectory inside a repository.** Worse, because it looks healthier: `git -C <subdir> grep -- '.claude/agents/'` runs successfully and matches nothing, since no `.claude/agents/` exists beneath the subdirectory. Nothing errors at all.

The blast radius is limited today and that is worth stating plainly: `git grep -n gaia_check_audit_key_callers`, excluding the script and its own test directory, returns no callers. Nothing in the repo currently passes it a bad root. This is a latent contract hole, not a live false green.

### Suggested fix

`.gaia/scripts/check-audit-base-derivation.sh` now carries the guard this one wants, so the fix is to mirror it rather than design it:

```bash
local prefix
if ! prefix="$(git -C "$repo_root" rev-parse --show-prefix 2>/dev/null)" || [ -n "$prefix" ]; then
  printf 'check-audit-key-callers: %s is not a git repository root; nothing was scanned\n' "$repo_root" >&2
  return 2
fi
```

`--show-prefix` settles both inputs in one call and needs no `cd`, which matters because `.claude/rules/shell-cwd.md` forbids one here: it fails outright outside a repository, and inside one it prints the path from the root down, empty exactly at the root. Return 2 rather than 1 so a caller can distinguish "the check says no" from "the check could not run".

Add the two sibling fixtures alongside it (non-repo, and a subdirectory of a real repo), matching `check-audit-base-derivation.bats`.

### Why this is filed rather than fixed in place

Surfaced by the `code-audit-maintainer-shell` member while auditing PR #1098, which added the same guard to the sibling script. It sits outside that PR's changed line ranges, has no live exposure, and folding it in would touch a second gate script plus its suite on the audit machinery's highest-stakes surface for no risk reduction. The member independently confirmed the deferral was correct.

Handler: prompt

stevensacks:
`gaia_check_audit_key_callers` took `repo_root` and went straight to `git grep` without confirming the path was a repository root. A `git grep` that cannot run returns the empty string, which is byte-identical to one that ran and matched nothing, so both assertions printed `0` and the function reported a clean bill of health for a tree it never read.

Three inputs reached that silence:

- a path outside any repository (`git grep` fails, stderr swallowed by `2>/dev/null`);
- a subdirectory of a real repository, which errors nowhere at all, since no `.claude/agents/` exists beneath it;
- a bare repository or a `.git` directory, both of which `--show-prefix` clears on an empty prefix.

## The fix

The guard is the one `.gaia/scripts/check-audit-base-derivation.sh` already carries, mirrored rather than redesigned. `--show-prefix` settles "is a repository" and "is its root" in one call and needs no `cd` (`.claude/rules/shell-cwd.md`, which that suite has a structural test for); `--is-inside-work-tree` settles the two shapes an empty prefix clears. Return `2`, deliberately distinct from assertion failure (`1`), so a caller can tell "the check says no" from "the check could not run". The sibling's comment block is carried across, adjusted for this function, and the function's own doc header now states the third return value.

No caller passes a bad root today (`git grep gaia_check_audit_key_callers` outside the script and its suite returns nothing), so this closes a latent contract hole, not a live false green.

## Verification

Three fixtures mirroring the sibling suite's, and each mechanism mutated separately rather than the file as a whole:

| mutation | reds |
|---|---|
| drop `--is-inside-work-tree` | the bare-repo/`.git` test alone |
| drop the empty-prefix requirement | the subdirectory test alone |
| remove the guard entirely | all three |

The third is there because neither of the first two reds the non-repo test, so without it that assertion would never have been shown to fire.

18/18 in this suite; full CI-gated sweep under bash 5: 1185 tests, 0 failures; `shell-lint` passes.

CHANGELOG entry landed under `## [Unreleased]` → `### Fixed`: the check ships to adopters (`owned` in `.gaia/manifest.json`), though its bats suite does not.

Closes #1100


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.adPO9sDjD5

Issue #1105, opened 2026-07-31
code-audit members review working-tree bytes for files their marker covers at HEAD


stevensacks:
## Aligning the changed-file LIST to HEAD does not align what the members READ

`.claude/agents/code-audit-frontend.md:683`

Every Code Audit Team member now resolves its changed-file list as `${BASE_SHA}...HEAD`, so the list names exactly the files its clearance digest covers (`git ls-tree HEAD`, `.claude/hooks/lib/audit-digest.sh`). That closes the list-side divergence.

It does not close the read-side one. A member reviews a file by `Read`ing it, and `Read` returns **working-tree** bytes. On a dirty tree the two still disagree:

```
$ git diff --name-only "$BASE...HEAD" -- '*.ts'
a.ts                       # in scope, correctly

$ cat a.ts                 # what Read returns
orig

$ git show HEAD:a.ts       # what the digest covers
changed
```

Commit a change, revert it in the working tree: the file is in scope and the member reads the reverted version, while its marker attests to a digest over the committed one. Same end state as the list-side defect, reached by a different route.

### Why it is filed rather than folded in

The list-side fix is a derivation change confined to one expression. Closing the read side is a different mechanism: it needs a dirty-tree posture at scope-resolution time, and the design question is which posture. Candidates, none obviously right:

- Refuse to resolve scope at all when `git status --porcelain` is non-empty within the member's remit. Strictest, but the default member's own self-heal deliberately dirties the tree, so the rule has to distinguish "dirty when I arrived" from "dirty because of me".
- Warn and proceed, recording the dirty paths in the findings sidecar so the divergence is at least on the record.
- Read through git (`git show HEAD:<path>`) rather than the filesystem for review purposes. Correct by construction, but it changes how every member reads every file, and the self-heal path genuinely needs working-tree access.

### Reachability

Low in practice. Every documented dispatch commits first: the merge gate runs on a pushed HEAD, membership is resolved from committed diffs (`resolve-audit-spawn.sh`, `resolve-audit-members.sh`), CI checks out clean, and the orchestrator commits between fix rounds (`wiki/concepts/PR Merge Workflow.md`). The gap needs an operator dispatching a member mid-edit. Filed because the guarantee the marker makes is unconditional while the thing backing it is not, and because the current prose now states the limit rather than leaving it implicit.

Handler: plan

PR #1106, opened 2026-07-31
debt(audit): all five members review the content their marker covers (#1099)

### LINKED ISSUE #1099: Code Audit Team members derive their changed-file list two different ways: frontend diffs the working tree, specialists diff HEAD
## The Code Audit Team's five members do not derive their changed-file list the same way

`.claude/agents/code-audit-frontend.md:655`

`code-audit-frontend` resolves the files it reviews with a bare-revision diff:

```bash
git -C "$AUDIT_ROOT" diff --name-only "$(cd "$AUDIT_ROOT" && .github/audit/resolve-audit-base.sh)" -- '*.ts' '*.tsx'
```

All four specialists use a three-dot range against a merge-base sha instead (`code-audit-github-workflows.md:59`, `code-audit-maintainer-node.md:62`, `code-audit-maintainer-prose.md:56`, `code-audit-maintainer-shell.md:72`):

```bash
BASE_SHA="$(git -C "$AUDIT_ROOT" merge-base "${BASE_REF}" HEAD 2>/dev/null || true)"
changed=$(git -C "$AUDIT_ROOT" diff --name-only "${BASE_SHA}...HEAD" 2>/dev/null || true)
```

### Failure mode

**Confirmed: the two forms disagree on uncommitted work.** `git diff <rev> -- <paths>` compares `<rev>` to the **working tree**; `git diff <rev>...HEAD` compares the merge base to the **HEAD commit**. Probed on this repo with one uncommitted edit to a file the branch does not otherwise touch:

```
$ git diff --name-only "$BASE" -- '.gaia/scripts/*.sh'          # frontend form
.gaia/scripts/append-audit-author.sh        <-- uncommitted, included
.gaia/scripts/check-audit-base-derivation.sh

$ git diff --name-only "$BASE...HEAD" -- '.gaia/scripts/*.sh'   # specialist form
.gaia/scripts/check-audit-base-derivation.sh
```

The frontend's own prose at line 655 defends the working-tree scope as "the right scope for a pre-commit/pre-merge review", so the divergence is at least partly deliberate on one side and undocumented on the other. What makes it a defect rather than a preference is the interaction with marker keying: a member's clearance digest is computed over tracked files **at HEAD** (`git ls-tree HEAD`, per `.claude/hooks/lib/audit-digest.sh`). So the frontend can review a working-tree state that HEAD does not contain and then write a marker attesting to a digest over content it never read. Concretely: commit a change in the PR, revert it in the working tree, and the frontend reviews the reverted (clean) version while its marker covers the committed one.

**Second facet, currently latent.** The frontend passes `resolve-audit-base.sh`'s output straight to `git diff` without the `merge-base` step every specialist applies. When the resolver falls back to `origin/main` (no audited ancestor yet) and `origin/main` has advanced past the fork point, the frontend's changed-file list includes files the PR never touched. Not reproducible on this branch today, `origin/main` currently sits exactly at the merge base, so this facet is unverified in practice. Note that `code-audit-frontend.md:524` already describes the base as feeding "the audit's `BASE_REF...HEAD` three-dot diff", which contradicts what line 655 actually runs.

### Suggested fix

Pick one semantics for the whole team and document why:

- **HEAD-only (`<base>...HEAD`) everywhere** aligns every member's review scope with the content its marker's digest actually covers, and closes the revert hole above. Cost: the frontend stops catching uncommitted work in a pre-commit review.
- **Working-tree everywhere** preserves the pre-commit review scope but leaves every member's marker attesting to a digest over content that may differ from what was reviewed.

If the working-tree scope is kept for the frontend, say so explicitly at both line 524 and line 655 (they currently disagree), and state that specialists deliberately differ. Independently of which arm is chosen, add the `merge-base` step to the frontend so the fallback-to-`origin/main` path stops widening the list when main advances.

Handler: plan

stevensacks:
## The defect

The Code Audit Team's five members derived their changed-file list two different ways. The four specialists resolve `${BASE_SHA}...HEAD`, comparing the review base to the **HEAD commit**. `code-audit-frontend` used a bare-revision two-dot diff, comparing the base to the **working tree**, in two places: the review's own file list, and the in-flight-fix scope filter that decides whether a finding is in scope at all. The issue cites only the first.

A member's clearance digest is computed over `git ls-tree HEAD` (`.claude/hooks/lib/audit-digest.sh`), so the two-dot form let the frontend certify a digest over content it never read. Three divergences, all dirty-tree-only, reproduced on scratch repos:

| | two-dot (before) | three-dot (after) |
|---|---|---|
| committed then reverted in the working tree | *(absent from review; marker still covers it)* | `a.ts` |
| never committed | `scratch.ts` *(reviewed, no marker covers it)* | *(absent)* |
| base is a ref whose tip advanced past the fork point | `a.ts unrelated.ts` | `a.ts` |

The third was filed as latent and unverified; it reproduces.

## Which arm, and why

**HEAD-only.** The frontend's own prose already described the range as three-dot in three places (`:83`, `:524`, `:653`) and `wiki/concepts/Code Review Audit Agent.md` in a fourth — only the executed command dissented. Its self-skip already reads a committed-only list from `resolve-audit-spawn.sh`, so the member decided *whether* to run from HEAD and then reviewed something else. And no documented flow dispatches a member against a dirty tree: the merge gate runs on a pushed HEAD, membership resolves from committed diffs, CI checks out clean, and the orchestrator commits between fix rounds.

`merge-base` stays, but for the **key** rather than the diff. `git diff <rev>...HEAD` resolves its own merge base, so three-dot narrows correctly whether the resolver returned a sha or a ref; `gaia_audit_key` still needs a stable sha.

## Guards

- **`check-audit-base-derivation.sh` assertion 3**: no `diff --name-only` naming any spelling of the review base may omit a `...` range. Deliberately *not* keyed to the pre-merge-base spellings alone — `"${BASE_SHA}"` performs the same two-dot comparison as `"$BASE_REF"`, so a rule that policed the variable would miss the likelier drift. (It did, during authoring: a stray whole-file `perl -0p` substitution reverted the `:281` restatement to two-dot and the narrower rule passed it. That is now a regression fixture.)
- **`audit-base-agreement.bats`**: two probes executing the frontend's real derivation snippet against scratch repos carrying a dirty tree and an advanced default branch — the multi-line shape a line-scoped static check cannot see.

Mutation-proved per mechanism: two-dot on `BASE_SHA` → behavioural probe + real-repo pin red; ref-tip anchor → assertion 3 red; the retired resolver-direct shape → both probes + assertion 3 red. Each of assertion 3's own terms (four base tokens, the range requirement, the `#` wall, the backtick wall) individually mutated and confirmed to red a distinct fixture.

## Not closed here

`Read` returns working-tree bytes, so on a dirty tree a file named in `changed` can still hold content HEAD does not. Aligning the file **list** does not align what is **read**. Filed as #1105 rather than folded in: closing it needs a dirty-tree posture at scope-resolution time, which is a different mechanism with a real design choice behind it. The new prose states the limit rather than implying the alignment is total.

## Verification

- `bats5 .gaia/scripts/tests/*.bats` — 1200 tests, exit 0
- Quality Gate skipped by its own skip logic: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config

Closes #1099


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.NvTXVwqJE1

PR #1108, opened 2026-07-31
debt(wiki): the update class table states what the merge walk does (#1050)

### LINKED ISSUE #1050: debt(wiki): two wiki pages contradict the code they document (update-gaia class table, gaia-lint 2.0.0)
Two wiki pages document behavior that no longer matches the code they describe. Filed together because both were surfaced by the same change (#1049) and both are doc-vs-reality drift in the same subsystem. The dedup key anchors on the first, which is the consequential one.

## 1. `wiki/concepts/Update Workflow.md:30` contradicts the update skill

The file-class table describes drifted `owned` files as:

> Pristine → overwrite silently. Drifted → prompt: skip / overwrite / backup+overwrite.

`.claude/skills/update-gaia/SKILL.md`, the executable spec, says otherwise in its own class table:

| `owned` | `A` ≠ `B` and `A` ≠ `L` | `diff -u "$A" "$L" > .gaia-merge/<path>.patch` | `conflicts[]` |

There is no prompt. A drifted `owned` file gets a patch written for the adopter to resolve by hand, and the update records it in `conflicts[]`.

**Failure mode, and it already happened.** During #1049 this page was the basis for deciding whether the `@gaia-react/lint` 2.0.0 CHANGELOG entry needed an **Action required:** anchor. Reading line 30, the conclusion was that adopters receive the new `...lint.reactRouter` spread automatically, so no anchor was warranted, and the entry was authored and committed without one. The pre-merge audit caught it and the anchor was added before merge.

Had it not been caught: an adopter who customized `eslint.config.mjs` (the pattern `@gaia-react/lint`'s own README invites, "adjust the override block at the bottom for your project") resolves a conflict patch and can land on 2.0.0 without the new spread. Their route modules written as `({}: Route.ComponentProps)` then fail `no-empty-pattern`, `pnpm lint --max-warnings=0` fails, the pre-commit gate blocks their commits, and because no **Action required:** anchor exists, `/update-gaia` records nothing in `adopterActions[]` and tells them nothing. A wrong wiki page produced a wrong decision that only a review caught.

**Suggested fix.** Replace the `owned` row's drift-handling cell with the skill's actual behavior (conflict patch to `.gaia-merge/<path>.patch`, recorded in `conflicts[]`). While there, verify the `shared` / `wiki-owned` rows against the same skill table; they appear to agree, but that was not checked as part of this finding.

## 2. `wiki/dependencies/gaia-lint.md` is stale for 2.0.0

- **`wiki/dependencies/gaia-lint.md:43`** — the canonical `eslint.config.mjs` example spreads `...lint.react,` with no `...lint.reactRouter,` after it. The repo's actual `eslint.config.mjs` now carries that line, so the documented config no longer reproduces the shipped one.
- **`wiki/dependencies/gaia-lint.md:61`-`71`** — the "Active rule groups" table has rows for `base`, `react`, `testing`, `storybook`, `playwright`, `guardrails`, `styleHygiene`, `betterTailwind`, and `prettier`, but none for `reactRouter`. The group is entirely undocumented.

**Failure mode.** Anyone copying the documented config gets one that silently differs from the shipped one: `no-empty-pattern` stays enabled for route modules, so `({}: Route.ComponentProps)` fails lint with nothing explaining why the documented and real configs disagree. Same class of error as item 1: a doc that reads as authoritative while describing superseded behavior.

**Suggested fix.** Add `...lint.reactRouter,` after `...lint.react,` in the example, and add a `reactRouter` row to the rule-group table noting it is React-Router-only and *subtractive* (it disables `no-empty-pattern` for `**/routes/**/*.tsx`, and a project on another file-based router should omit it).

Handler: prompt

stevensacks:
## What

`wiki/concepts/Update Workflow.md` documented a prompt that `/update-gaia` does not have. Its file-class table said a drifted `owned` file offers **skip / overwrite / backup+overwrite**. The executable spec (`.claude/skills/update-gaia/SKILL.md`, Step 7) writes a `.gaia-merge/<path>.patch` and records the path in `conflicts[]`. No class prompts on drift.

The false claim appeared in **three** places on the page, not one:

- the file-class table's `owned` drift cell (the issue's named target)
- the decision table's `Adopter drifted, latest changed, owned` row
- the `Never auto-clobber drift` safety invariant

and a fourth time in `SKILL.md`'s own preamble (`prompts if drifted`), which contradicted the Step 7 table further down the same file. Fixing only the wiki would have left the skill's summary asserting the same wrong thing, so it is corrected too.

## Why it matters

This page was the basis for deciding whether the `@gaia-react/lint` 2.0.0 CHANGELOG entry needed an **Action required:** anchor. Reading the drift row, the conclusion was that adopters pick up the new `...lint.reactRouter` spread automatically, so no anchor was written. The pre-merge audit caught it. Uncaught, an adopter resolving a conflict patch could land on 2.0.0 without the spread, fail `no-empty-pattern` on their route modules, and get nothing in `adopterActions[]` explaining it.

## Corrected alongside

Same defect class (documenting a prompt or default the walk does not have), in the tables already being rewritten. Called out separately so it is easy to push back on:

- `New file: add (default yes)` implied a confirmation; new files are added and reported as a count.
- The upstream-deletion table invented a drift-dependent split with opposite defaults. The skill has one rule: confirm before removing, never auto-delete, drifted or not.
- The `.gaia-backup/` row described backups of "any file the adopter agreed to overwrite". Backups are taken on the silent-overwrite path, which involves no agreement.
- Decision-table row order now matches the skill's declared order, which is load-bearing because the skill matches first-row-wins. Added the `owned` + baseline-missing row, which is what actually distinguishes `owned` from `shared` now that the drift column reads the same for all three classes.

## Issue scope notes

- **Item 2 was already fixed.** `wiki/dependencies/gaia-lint.md` already carries `...lint.reactRouter,` in the config example and a `reactRouter` row in the rule-group table. No change needed.
- **The `shared` / `wiki-owned` rows the issue asked to verify agree with the skill** and are unchanged.

## Verification

Markdown-only, so the Quality Gate skips by its own rule. Ran what covers these paths:

- `.gaia/scripts/tests/main-only-lib.bats` (references `SKILL.md`): 20/20 pass
- `.gaia/tests/hooks/audit-scope-routing-parity.bats` (its fixture names both files): pass
- `bash .gaia/tests/shell-lint.sh`: pass
- wiki-style + repo-relative-path audit greps: clean
- Repo-wide sweep for the false claim on any other surface: no other occurrence

**Pre-existing failure, not from this branch:** `.gaia/tests/distribution/17-gaia-update-merge-region.sh` fails at build-staging on two `maintainer-paths` leaks (`.gaia/scripts/check-audit-base-derivation.sh:16`, `wiki/concepts/Code Review Audit Agent.md:55`). Reproduced identically on clean `main` at the same commit (`01b0be97`). Untouched by this PR.

Closes #1050


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.eXiwyoIkRW

PR #1120, opened 2026-07-31
debt(audit): the per-call window closes at ';', and the suite path ships wrapped

### LINKED ISSUE #1107: check-audit-base-derivation: assertion 3's per-call window is not walled at ';', and its positive pin misses quoted command substitution
Two related gaps in the assertion-3 machinery, raised as accepted-and-noted Suggestions by the round-2 `code-audit-maintainer-shell` audit on #1106 and deliberately not folded in there (the member was already marked and nothing else was rotating its digest). They share a file pair and a fix session, so they are filed together.

Both are latent: neither is reachable against the roster as it stands today. They are gaps in the guard's own logic, not live failures.

## 1. The per-call window is not walled at `;`

`.gaia/scripts/check-audit-base-derivation.sh:335-337`

`_gaia_keep_unanchored_diff_matches` walks every `diff --name-only` occurrence on a line and cuts a window after it at three walls: the next `diff --name-only`, a `#`, and a backtick. A call is reported as a violation when its window names a review base (`BASE_SHA` / `BASE_REF` / `FULL_BASE` / `resolve-audit-base.sh`) but carries no `...` range.

**Failure mode.** `;` is not a wall, so on a `;`-joined line the window runs past the end of the first command and into the next one. Given a definition writing:

```bash
changed=$(git -C "$AUDIT_ROOT" diff --name-only "$BASE_SHA") ; echo "${BASE_REF}...HEAD"
```

the first call is a two-dot comparison, exactly the drift assertion 3 exists to catch. Its window is cut by none of the three walls, so it extends to end of line and picks up the `...` belonging to `echo`. The `index(window, "...")` test succeeds on a range that is not this call's, the call is not printed, and the check reports `review diffs consuming a base that never reached the fork point: 0`. A guard reports green in precisely the case it exists to catch.

The sibling walk in the same file, `_gaia_drop_full_base_matches` (`:275`), already treats `;` as a wall through its stop class ``[^|;&#)`]``. The two per-call walks currently disagree about what closes a call.

**Suggested fix.** Add a fourth wall mirroring the sibling:

```awk
if ((w = index(window, ";")) > 0)   window = substr(window, 1, w - 1)
```

The comment at `:305-310` reasons about `|` and `;` together as deliberately-not-walls. That reasoning is sound for `|` and does not carry to `;`: it rests on `2>/dev/null || true` closing almost every real call, which contains no `;`. Nothing in a real call's argument list does. Update that comment to keep `|` excluded for the stated reason and drop `;` from the exemption.

**Fixture to add.** A `;`-joined line whose trailing command carries a `...` while the leading `diff --name-only` call is two-dot. It must red before the fix and green after. Per the mutation discipline in `.claude/rules/pr-merge.md`, also confirm the new wall can fail on its own: remove the `;` wall and check the fixture goes red, rather than only checking the suite is green with it in place.

**Currently unreachable.** The only roster line matching `diff --name-only.*;` is prose in `.claude/agents/code-audit-frontend.md:281`, where the `;` falls after a closing backtick and is already walled.

## 2. The real-repo positive pin matches only unquoted command substitution

`.gaia/scripts/tests/check-audit-base-derivation.bats:658`

The test `real repo: every changed-file diff in the roster is a three-dot range` pins the positive form of the invariant that assertion 3 states negatively:

```bash
run git -C "$REPO_ROOT" grep -hIE '^[a-z_]+=\$\(git .*diff --name-only' -- '.claude/agents/'
```

**Failure mode.** `=\$\(` requires the `$(` to sit immediately after the `=`, so the pattern matches an unquoted command substitution only. Normalizing a definition to the quoted form, `changed="$(git -C "$AUDIT_ROOT" diff --name-only ...)"`, silently drops that line out of the net: a `"` now sits between `=` and `$(`. The test still passes, because its only breadth assertion is `[ -n "$output" ]`, which is satisfied by whatever other lines still match. The net shrinks and nothing reports it.

This costs a redundant second net rather than the guarantee itself, since assertion 3 preserves the invariant either way. It is still a check that quietly stops checking what it names.

**Suggested fix.** Widen the pattern to accept the optional quote and pin the expected breadth so the net cannot silently shrink:

```bash
run git -C "$REPO_ROOT" grep -hIE '^[a-z_]+="?\$\(git .*diff --name-only' -- '.claude/agents/'
```

plus an assertion on the expected match count. Both patterns return the same 9 lines against the roster as it stands, so the count to pin is 9 and the widening is a no-op today. Pin the count against the widened pattern, and treat a future change in the roster's definition count as a deliberate update to this number rather than something to relax.

## Provenance

Raised on #1106 (which closed #1099) by `code-audit-maintainer-shell`, round 2, both graded Suggestion and neither blocking. Not folded into that PR because the member had already earned its marker and no other change was rotating its content digest, so applying either would have bought a full re-dispatch (~60-110k tokens) for a latent nit. See the digest-economics guidance in `wiki/concepts/PR Merge Workflow.md`.

Handler: prompt

### LINKED ISSUE #1110: Two maintainer-path leaks fail build-staging, which fails every distribution scenario
**Location:** `.gaia/scripts/check-audit-base-derivation.sh:16` and `wiki/concepts/Code Review Audit Agent.md:55`

## Failure mode

The release scrub's leak detector fails with two `maintainer-paths` leaks, which fails `build-staging.sh`, which fails **every** distribution-suite scenario (all of them build a staging tree first).

Reproduced on `main` at HEAD (`01b0be97`) with a clean working tree:

```
$ bash .gaia/tests/distribution/lib/build-staging.sh "$(mktemp -d)"
release scrub: stripped 69 marker block(s) across 36 file(s)

leaks (2):
  [maintainer-paths] .gaia/scripts/check-audit-base-derivation.sh:16  .gaia/scripts/tests/
  [maintainer-paths] wiki/concepts/Code Review Audit Agent.md:55  .gaia/scripts/tests/
```

Both are prose references naming `.gaia/scripts/tests/audit-base-agreement.bats`, a release-excluded maintainer-only path, from a file that ships to adopters. On an adopter clone those references name a path that does not exist.

Introduced by the recent audit work (#1099 / #1103), which added the behavioural-suite cross-references.

**Why it has stayed latent, and why that ends.** The `CLI Tests` workflow's distribution job is path-filtered on `.gaia/cli/**`, `.gaia/release-exclude`, `.gaia/release-scrub.yml`, `.gaia/tests/distribution/**`, and its own YAML. The last four pushes to `main` touched none of those, so the job reported `skipped` each time and nothing ran the detector. The next PR that touches `.gaia/cli/**` turns the required check red for a defect it did not introduce.

## Suggested fix

The two halves need **different** fixes, which is the substance of this issue:

- `wiki/concepts/Code Review Audit Agent.md:55` wraps in `<!-- gaia:maintainer-only:start -->` / `<!-- gaia:maintainer-only:end -->` markers. The bundle-time scrub strips marker-delimited blocks from markdown under `wiki/`, `.claude/`, and `.specify/extensions/gaia/`, so this one is the ordinary path. Keep the general statement in the visible body and confine the suite's path to the wrapped block.
- `.gaia/scripts/check-audit-base-derivation.sh:16` **cannot** use markers: the scrub only applies them to markdown under those three roots, so a marker block in a `.sh` file is inert and the reference still ships. That line needs either a rephrase that does not name the path (the surrounding comment already says "the behavioural suite next door", which carries the meaning without the path) or an explicit leak-detector allowlist entry.

Prefer the rephrase over an allowlist entry unless the path is genuinely load-bearing for a maintainer reading that comment: an allowlist entry weakens the detector permanently for that path, while the rephrase costs nothing.

Handler: plan

stevensacks:
Batch fix for two tech-debt issues sharing `.gaia/scripts/check-audit-base-derivation.sh`.

## #1110 — two maintainer-path leaks fail build-staging

`.gaia/scripts/tests/audit-base-agreement.bats` is release-excluded, and two shipped surfaces named it unwrapped. The scrub's leak detector failed, which failed `build-staging.sh`, which fails **every** distribution scenario since they all stage a tree first. It stayed latent only because the `CLI Tests` distribution job is path-filtered and the last four pushes to `main` matched none of its paths; the next PR touching `.gaia/cli/**` would have turned a required check red for a defect it did not introduce.

Both halves now wrap the path in maintainer-only markers.

**The issue's suggested fix for the `.sh` half rested on a wrong premise and is not what landed.** It said markers are inert in a `.sh` file and recommended a rephrase or a leak-detector allowlist entry. There is in fact a third `marker-strip` transform (`.gaia/release-scrub.yml:63-69`) covering `**/*.sh` with the `# gaia:maintainer-only:` pair, and `.gaia/scripts/lint-hook-array-guard.sh` and `.gaia/scripts/summary-verify.sh` already use it for exactly this purpose — the scrub config's own comment (`:198-202`) documents the convention. Wrapping keeps the path for maintainers, matches the two sibling scripts, and avoids permanently weakening the detector, so no allowlist entry was added.

Each visible half is self-contained once the block is stripped: neither leaves a dangling "that suite" reference on an adopter clone.

## #1107 — two gaps in the assertion-3 machinery

**The `;` wall.** `_gaia_keep_unanchored_diff_matches` cut each call's window at three walls (next call, `#`, backtick). `;` was not one, so on a `;`-joined line the window ran past the end of the first command: a two-dot `diff --name-only` followed by any later command carrying a `...` was vouched for by dots that were not its own, and the check reported `0` in precisely the case it exists to catch. The sibling `_gaia_drop_full_base_matches` already walled at `;` through its stop class; the two per-call walks agree again. `|` stays unwalled for the stated reason (a pathspec or redirect routinely follows the revision argument).

**The positive pin.** `^[a-z_]+=\$\(git ...` required `$(` immediately after the `=`, matching the unquoted assignment only. Normalizing a definition to `changed="$(git ...)"` would silently drop it from the net while `[ -n "$output" ]` stayed satisfied by the remaining lines. The pattern now accepts `"?` and the breadth is pinned at 9 (both patterns return the same 9 lines against today's roster, so the widening is a no-op now and a guard later).

## Verification

Per `.claude/rules/pr-merge.md`, each mechanism was mutated separately rather than only confirmed green:

- **New fixture** `DIFF_SEMICOLON_LATER_RANGE` — a `;`-joined line whose trailing command is not a diff call at all, so no other wall can close it. Confirmed RED before the fix (the check exited clean on a call it should catch) and green after.
- **`;` wall mutation** — removing the wall line from the guard's own formula reds exactly that one test and nothing else.
- **Count-pin mutation** — changing the pin from 9 to 8 reds the real-repo test, so the assertion fires.
- **Widening mutation** — doctoring one roster definition to the quoted form: the widened net still sees 9, the pre-fix narrow net sees 8, and with the pattern narrowed back the count pin reds on exactly that shrink.
- Full suite green (42 tests, bash 5 via `.gaia/scripts/bats5.sh`), sibling `audit-base-agreement.bats` green, `.gaia/tests/shell-lint.sh` passed.
- `build-staging.sh` reports `leaks: none`; distribution scenarios `01`, `02`, `03`, and `16` all PASS.
- Wiki-style, repo-relative, and marker-balance audits clean.

## CHANGELOG

No `## [Unreleased]` entry. Both defects live in unreleased code: `check-audit-base-derivation.sh` does not exist at `v1.6.1`, and the leaked wiki line came from `01b0be97`, whose entry already sits in `## [Unreleased]`. That entry describes the shipped behavior and stays accurate; a separate entry would narrate a fix to something no adopter has received.

Closes #1110
Closes #1107


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.gTGFJowFHT

PR #1127, opened 2026-07-31
debt(audit): the marker block owns its separator, and the roster pin reports the count it saw (#1121, #1122, #1123)

### LINKED ISSUE #1121: check-audit-base-derivation: the wrapped suite reference leaves two bare '#' separators in the stripped adopter copy
**Location:** `.gaia/scripts/check-audit-base-derivation.sh:26`

## Failure mode

The maintainer-only block at `:22-25` sits between a bare `#` separator at `:21` and another at `:26`. The scrub's `**/*.sh` marker-strip removes the block itself but neither surrounding separator, so the shipped adopter copy carries two consecutive bare `#` comment lines where the block used to be.

Reproduce against a staging tree:

```
$ bash .gaia/tests/distribution/lib/build-staging.sh "$(mktemp -d)"
$ sed -n '18,24p' <staging>/.gaia/scripts/check-audit-base-derivation.sh
```

The doubled separator appears immediately after `# key itself.`.

Cosmetic only: the file stays valid bash and no check fails on it. The cost is that every adopter's copy of this header carries a small artifact of a block they cannot see.

**This is a shared convention, not a defect unique to this file.** `.gaia/scripts/lint-hook-array-guard.sh` wraps its own sibling-suite reference the same way (`:8-15`, bare `#` on both sides) and produces the identical doubled separator, so a fix that only touches this file leaves the convention inconsistent. `.claude/hooks/lib/audit-selfheal-paths.sh:40` places its block mid-list and leaves no orphan, which is the shape that already avoids this.

## Suggested fix

Pull one separator inside the block so exactly one survives the strip: make the block start after `:21`'s `#` and end with a `#` that is itself inside the markers, or drop `:21`. Decide whether to apply the same shape to `lint-hook-array-guard.sh` (and any other shipped script using this wrap) so the convention is uniform rather than per-file.

Handler: prompt

### LINKED ISSUE #1122: check-audit-base-derivation: two comments still promise a 'behavioural suite next door' absent from every adopter clone
**Location:** `.gaia/scripts/check-audit-base-derivation.sh:123` and `:186`

## Failure mode

Two comment blocks in this shipped script promise a companion suite that does not exist on an adopter clone:

- `:123` — "The behavioural suite next door executes the real fences and compares the resulting file lists, which is what covers that shape."
- `:186` — "The behavioural suite next door covers both, by executing the real fences."

The suite they mean is `.gaia/scripts/tests/audit-base-agreement.bats`, which lives under `.gaia/scripts/tests/` and is release-excluded, so it is absent from every adopter tree. Both sentences tell an adopter reading this header that a stated coverage gap is closed by something they can neither find nor run, and each is load-bearing prose: it is the stated reason the adjacent assertion is allowed to be line-scoped.

This is **not** a leak-detector failure and does not fail `build-staging.sh`: the maintainer-paths pattern matches paths, and neither sentence names one. It is prose accuracy on a shipped surface, which is why it survived the fix that wrapped the one *named* reference at `:23`.

## Suggested fix

Two reasonable options, and the surrounding conventions pick between them:

1. Wrap both sentences in the same `# gaia:maintainer-only:start` / `# gaia:maintainer-only:end` pair the named reference at `:22-25` already uses. The scrub's third `marker-strip` transform covers `**/*.sh`, so they are stripped from adopter bundles. This keeps the rationale intact for maintainers.
2. Reword each to state the uncovered gap without promising a companion, e.g. "this assertion is line-scoped; the multi-line form is out of its reach."

Prefer (1) where the sentence explains *why* the assertion may be line-scoped, since that rationale is worth keeping for the maintainer audience, and (2) where the sentence only gestures at coverage. Whichever is chosen, apply it to both sites so they do not diverge.

Handler: prompt

### LINKED ISSUE #1123: check-audit-base-derivation.bats: the roster breadth pin discards the count it observed, so a shrink and a roster change fail identically
**Location:** `.gaia/scripts/tests/check-audit-base-derivation.bats:694`

## Failure mode

The breadth pin in `real repo: every changed-file diff in the roster is a three-dot range` is:

```bash
[ "$(grep -c . <<<"$output")" -eq 9 ]
```

When it fails, bats reports only that the `[ ... ]` returned non-zero. The observed count is computed inside the command substitution and discarded, so the failure output never says whether the net matched 8 lines or 40.

That distinction is exactly the one a maintainer needs, and the two causes point opposite ways:

- **The net shrank** (a definition normalized to a spelling the pattern no longer matches) — a real regression in the assertion's coverage, which is the case this pin was added to catch.
- **The roster grew** (a sixth Code Audit Team member landed, or a member gained a second diff line) — a legitimate change, and the number is simply due an update.

With no count in the output, telling them apart means re-running the `git grep` by hand. The `while` loop immediately below already prints its offending line on failure, so the surrounding test sets the better precedent.

## Suggested fix

Assign the count first and print it on the failing branch:

```bash
net_lines="$(grep -c . <<<"$output")"
[ "$net_lines" -eq 9 ] || {
  printf 'roster diff-line net covered %s lines, expected 9\n' "$net_lines"
  return 1
}
```

Per `.claude/rules/bats-assertions.md` this keeps the assertion failing correctly on bash 3.2, since the failing branch ends in an explicit `return 1` rather than relying on a bare `[[ ]]` or a `!`-negation.

Handler: prompt

stevensacks:
Three advisory Suggestions from the same auditor on the same file, fixed as one unit because `code-audit-maintainer-shell` is the sole auditor for all three files.

Closes #1121
Closes #1122
Closes #1123

## #1121 — the doubled separator, and its scope

The scope question first, since that is the issue's real content. Answered empirically: built a staging tree and scanned every shipped `.sh` for a strip-introduced doubled bare `#`.

- **staging:** `check-audit-base-derivation.sh:22`, `lint-hook-array-guard.sh:8` — those two, no others
- **source:** zero, so the strip was the sole cause

The other wrap sites already avoid it: `check-audit-key-callers.sh:45` sits mid-paragraph, `resolve-audit-spawn.sh:120` and `summary-verify.sh:25` sit at end-of-header, `audit-selfheal-paths.sh:40` sits mid-list. The artifact appears only when a block sits between *two* bare separators.

**Convention: the block owns the separator that introduces it** — start marker first, bare `#` on the next line inside the markers. Exactly one survives, and deleting the block leaves the surrounding prose correctly separated, which is what the strip does. Recorded in `wiki/decisions/Bundle-time Scrub.md` under Marker discipline.

Applied to all three shipped scripts that departed from it. `summary-verify.sh` came in during review: it carried the milder sibling shape (separator *before* the block), which strips to a dangling `#` above the first line of code. `resolve-audit-spawn.sh` matches too but is release-excluded, so it reaches no adopter and is left alone.

A convention nobody can see in source is a convention that drifts back, so `03-marker-strip.sh` gains a fourth assertion: **no staged `#`-marker file carries a doubled bare `#` separator**, whichever produced it — an adopter reads the same artifact either way. The report names both the source and staged counts, so which cause it was stays readable. It reuses the staging tree the scenario already builds, so it costs no extra work.

Timing, stated plainly: the PR lane for the distribution scenarios (`cli-tests.yml`) filters on bundle inputs, so a PR that adds a wrap under `.gaia/scripts/**` does **not** trip this. It fires from `release.yml` before the tarball is built, which is the boundary the artifact actually concerns. Filed as #1129 rather than fixed here, because closing it is a real CI-cost tradeoff (widen the filter vs. add a narrow job), not a one-liner.

## #1122 — the absent companion suite

Both sentences are load-bearing (each is the stated reason its assertion may be line-scoped), so both take option (1): wrapped, not reworded, applied identically to `:123` and `:186`. Wording unchanged; only the wrapping and line re-flow.

Both wraps use #1121's settled convention rather than the current one, so neither creates a new instance of the artifact. `:186` is the clearer demonstration — the block owns the separator, so the stripped copy runs prose straight into the code line with no orphan separator at all.

That site also gained an adopter-visible **"Neither gap is closed statically."** Without it the stripped copy ended on an acknowledged hole whose resolution had just been stripped away. The sentence promises no companion, which is the defect being fixed — naming one, even obliquely, would have reintroduced it.

`grep -rn "suite next door"` over the staging tree returns nothing.

## #1123 — the discarded count

Applied the suggested fix, with one change: the issue's snippet has `-eq 9` and `expected 9` as two independent literals, so an update to one could leave the message claiming an expectation the assertion does not hold. Named once as `expected_lines` instead.

## Verification

- `build-staging.sh` → `leaks: none`; exactly one separator survives at each fixed site; doubled-separator sweep over all staged `.sh` → none; `"suite next door"` absent from staging
- `check-audit-base-derivation.bats` 42/42, `audit-base-agreement.bats` 9/9, `summary-verify.bats` 9/9, all under bash 5 via `bats5.sh`
- `shell-lint.sh` → passed; `03-marker-strip.sh` → PASS

**Every new mechanism mutated separately, not just its subject:**

| mutation | expected | observed |
|---|---|---|
| #1123 pin `9`→`8` (mechanism) | red, prints observed count | `not ok` + `covered 9 lines, expected 8` |
| #1123 net shrunk to 8 (subject) | red, prints the *different* count | `not ok` + `covered 8 lines, expected 9` |
| #1121 wrap reverted (subject) | guard red, names the file | `FAIL … src=0 staged=1` |
| #1121 run-counter broken (mechanism) | greens despite the artifact | PASS ⇒ counter load-bearing |
| #1121 comparison weakened (mechanism) | greens despite the artifact | PASS ⇒ comparison load-bearing |
| #1121 trailing-`ws` term dropped, `'#  '` fixture (mechanism) | misses the artifact | PASS ⇒ term load-bearing |
| #1121 authored double, no marker involved (subject) | red, attributed to the header | `FAIL … src=1 staged=1` |

The two #1123 rows are the point of that issue: a shrink and a roster change now print different numbers. The old form printed neither.

All checks re-run after the last edit, including prose-only ones and after the rebase. Every mutation restored; working tree clean.

## Review rounds

Three audit rounds, all clean passes. Rounds 1 and 2 each found real defects in the round before, both folded in:

- **Round 1** → 5 Suggestions, all applied. Two mattered: the guard counted a separator only when the line was byte-exactly `#`, so one with trailing whitespace evaded the check it exists to enforce; and it gated on a staged-vs-source *delta*, which nets a strip-introduced double against an authored one removed elsewhere in the same file. Both fixed, both now mutation-proved above.
- **Round 2** → 1 Suggestion + 1 cross-remit, both applied. The delta→absolute change had left three pieces of prose describing the gate it replaced, including the wiki page that is the convention's source of truth. A maintainer reading it would have concluded an authored double was safe.
- **Round 3** → clean, one marginal diagnostic-hint Suggestion. Accept-and-noted per digest economics (PR clean, member marked, folding buys a full re-dispatch to re-earn the marker) and filed as #1133; the auditor concurred it did not warrant a fourth round.

## Gates

- **Quality Gate:** skips. The staged-files grep from `wiki/decisions/Quality Gate.md` returns nothing — three `.sh`, one `.bats`, one test `.sh`, one wiki `.md`, one harness `.md`.
- **CHANGELOG:** no `## [Unreleased]` entry, re-derived after the scope grew. Every script touched is absent at `v1.6.1` (unreleased code), and the two files that do exist there — `03-marker-strip.sh` and `Bundle-time Scrub.md`, plus the harness `README.md` — are release-excluded and confirmed absent from the staging tree. Nothing here is adopter-visible behavior.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.xCExPvgkzc

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.HJ2F3ifKYy

PR #1137, opened 2026-08-01
feat(audit): widen the machinery waive to any file the PR already touches

stevensacks:
## What this changes

`machinery_waived` records a non-security out-of-scope audit finding in the pull request body instead of opening a durable tech-debt issue. Its eligibility was a fixed list of gate-machinery paths, and the only member with a waive path at all was the default one. Both facts are why draining the tech-debt backlog files roughly as many issues as it closes: the drain works on surfaces the machinery list never covered, owned by members that file nothing themselves.

Eligibility becomes the **union** of the gate-machinery path set and the pull request's own changed-file set. The rule is stated at the orchestrator's own disposition contract, where all five Code Audit Team members' out-of-scope findings converge, and both merge gates verify it through one shared implementation.

## The shape of it

- **Verify side.** `disposition_offenders` gains an optional acting-root parameter and re-derives the changed-file set itself: whole-PR fork point, unfiltered by file type, exact whole-string equality. Never the TS/TSX-filtered review scope, never a prefix or basename test.
- **Write side.** The default member's agent file carries the one executable copy of the same derivation, named `FULL_BASE`, kept honest by the base-derivation guard and by a suite that asserts the two agree.
- **Unresolved base is not an empty answer.** git resolves an empty left side to HEAD, so an unresolved base and a resolved base with no differences yield identical empty diffs. The base is what gets tested; an unresolved one leaves the gate-machinery term deciding alone.
- **Attribution.** The sidecar is keyed by a content digest that does not rotate for every diff, so one file can be read while judging several consecutive pull requests. An entry's changed-files term is evaluated only when the sidecar's recorded sha is absent, an ancestor of HEAD, or orphaned. The orphan probe keeps an amend, rebase, or force-push on the same branch judged rather than silently set aside.
- **Operator visibility.** `disposition_notes` reports the three states that are not "verified clean" (classifier unavailable, sidecar not attributable, diff base unresolved), so an unrun check is distinguishable from a passed one.
- **Label.** The offender label becomes `machinery-waived-not-eligible`; the predicate no longer makes the machinery judgment the old spelling claimed. The sidecar enum value `machinery_waived` is unchanged.

## Verification

Per-phase quality gate plus `shell-lint.sh`, `check-audit-base-derivation.sh`, and the three bats trees under the bash-5 guard. An orchestrator-owned mutation-proof matrix runs after the suites land: one mutation and one victim per mechanism, each confirmed red under mutation and green on restore.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.DLbaH7Oonb

PR #1139, opened 2026-08-02
feat(audit): measure the audit re-spawn before re-keying the marker

stevensacks:
## What

The Code Audit Team's spawn oracle re-dispatches a member whenever that member's content digest rotates, and one cause of a rotation is not a content change at all: a peer's merge landing under an already-audited branch. Nobody knows how often that actually happens.

This builds the instrument rather than the cure. The oracle appends one JSON-Lines breadcrumb per considered member at the one point that already holds both the digests and the marker store, a session-start sweep bounds the resulting ledger, and a documented query turns it into a count of peer-merge-attributable re-spawns.

## How

- `.gaia/scripts/audit-respawn-lib.sh`, sourced-only shared library: ledger path resolver, fail-open silent record writer, and two retention knobs with default-then-floor semantics.
- `.gaia/scripts/resolve-audit-spawn.sh` writes one breadcrumb per considered member, inside the digest-marker filter's active branch only. Its stdout and exit status stay byte-identical on every path.
- `.gaia/state-registry.json` registers the ledger with a real reaper. The retention window outranks the record cap: the cap may only trim records already outside the window.
- `.gaia/scripts/audit-respawn-prune.sh` reaps, delegated from the session-start hook. `.gaia/scripts/audit-respawn-report.sh` runs the documented attribution query.

## Notes

Exactly one changed path is audit machinery, `.gaia/scripts/resolve-audit-spawn.sh`; every other changed path returns flag `0`. No byte of the digest recipe or its consumers changes. Editing the oracle rotates all five member digests once, so this lands while no other pull request is open.

### DIFF
--- a/.claude/hooks/wiki-session-start.sh
+++ b/.claude/hooks/wiki-session-start.sh
@@ -13,4 +13,10 @@ git rev-parse HEAD > "$GIT_DIR/claude-session-start" 2>/dev/null || true
 # blocks the session. See local-janitor.sh for the provable-death contract.
 [ -f .claude/hooks/local-janitor.sh ] && bash .claude/hooks/local-janitor.sh || true
 
+# Bounded prune of the Code Audit Team re-spawn breadcrumb ledger: age-drops
+# records past the retention window and caps the file's line count. Side-effect
+# only; never blocks the session. Absent on a clone that does not carry the
+# spawn oracle, where the guard makes it a no-op.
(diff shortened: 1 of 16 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.96LBfX5ocL

PR #1140, opened 2026-08-02
docs(audit): the bats rule carries the trap its own advice creates

stevensacks:
Closes two of the three follow-ups left open when the machinery-waive widening merged. The third is declined; see below.

## What changed

**`.claude/rules/bats-assertions.md`** — the `!`-negation section prescribes writing an absence check as `<positive-condition-for-the-bad-case> && return 1` and stops there. As a test's *final* statement that form inverts its own result: an absent needle leaves grep's non-zero status as the AND-list's status, which becomes the test's return value, so the test fails in exactly the case it exists to pass. Three lines now state the trap and the fix (an explicit trailing `true`), lifting phrasing already audit-proven in `.gaia/tests/lib/doc-machinery-waive-prose.bats`'s header rather than inventing new.

Placed in the version-independent `!`-negation section, not the bash-3.2 one: reproduced on bash 3.2.57 and 5.3.15, both fail. `cmd && return 1` is an AND-list whose status is `cmd`'s own, and `set -e` is ignored for a non-final member of an AND-OR list. The section that prescribes the form is where the trap belongs.

**`.gaia/scripts/check-audit-base-derivation.sh`** — one clause on `_gaia_drop_full_base_matches`'s awk docstring. It described `FULL_BASE` as the self-skip base only; the exemption also covers the default member's waive-eligibility derivation, which the assertion-1 charter comment directly above it already states. The two comments now agree. Comment only, no behavior change.

## Declined, recorded so it is not re-raised

The macOS BWK awk backslash-stripping gotcha (`-v`-assigned regex loses `\(` `\)`) is **not** added to any rule. It is awk portability in a per-suite copy-pasted helper rather than assertion hygiene; it produced no false pass, because `extract_section_or_fail`'s vacuous-match guard caught it loudly; and it is already documented at the point it bites, in the header of the suite that hit it. A rule is the wrong altitude for something that fired once.

## Verification

- `bash .gaia/tests/shell-lint.sh` — passes (193 scripts, 181 suites, hook-array guard clean)
- `.gaia/scripts/tests/check-audit-base-derivation.bats` under the bash 5 guard — 42/42, exit 0
- Quality Gate — skipped, confirmed against `wiki/decisions/Quality Gate.md`: no staged file matches the source-extension or gate-affecting-config lists, and line 46 names `.claude/**` and pure-markdown commits explicitly.

No mutation testing: both edits are prose and a comment with no executable behavior.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.xt2FdCSRsS

Issue #1059, opened 2026-07-26
wiki sync's recovery baseline is resolved by commit date, so a commit that lands after its own date is silently skipped


stevensacks:
`/gaia-wiki sync` records where it stopped as a commit SHA in `wiki/.state.json`. When a sync runs on a feature branch, that SHA is a branch commit, and a squash merge destroys it. The recorded marker is then unreachable from `main`, and the CLI falls back to locating the baseline **by time**:

```
git rev-list -1 --before=<last_evaluated_at> HEAD
```

`.gaia/cli/src/wiki/util/git.ts:79-88` (`ancestorBefore`), consumed at `.gaia/cli/src/wiki/state.ts:245-246`, driven by `.claude/skills/gaia/references/wiki/sync.md:20`.

That resolves `argmax{committer_date <= last_evaluated_at}` over HEAD's ancestors, and the sync then evaluates `suggested_base..HEAD`. **A commit is skipped iff it is an ancestor of `suggested_base`.**

So the precondition is:

> a commit whose committer date is **older** than `last_evaluated_at`, which becomes **reachable** from `main` only **after** it.

A merge commit produces exactly that: the merged branch's commits keep their original committer dates and arrive on `main` on the merge date. The skip is silent. No conflict, no warning, and `commits_ahead` reads 0 on the recovery path so nothing looks wrong.

## Reproduction

Fails on `main` today. Save and run; it builds a throwaway repo and needs no fixtures.

```bash
#!/usr/bin/env bash
set -euo pipefail
R="${1:?scratch dir required}/f73-repro"
rm -rf "$R"; mkdir -p "$R"; cd "$R"
git init -q -b main .; git config user.email t@t; git config user.name t
mk(){ export GIT_COMMITTER_DATE="$1" GIT_AUTHOR_DATE="$1"; echo "$2" > "$3"; git add -A; git -c commit.gpgsign=false commit -q -m "$2"; }
mkdir -p wiki
mk "2026-01-01T00:00:00Z" C0 wiki/seed.md
mk "2026-01-02T00:00:00Z" C1 f1.txt
git branch B; git branch A
# Branch B: work COMMITTED before the sync, LANDS after it.
git checkout -q B; mk "2026-01-03T00:00:00Z" B1-work-needing-evaluation f2.txt
# Branch A: runs the sync at T = Jan 5, recording its own tip as the marker.
git checkout -q A; mk "2026-01-04T00:00:00Z" A1 f3.txt
[...]

PR #1151, opened 2026-08-02
fix(cli): propagate git failures out of commitDetails (#1135)

### LINKED ISSUE #1135: debt(cli): commit-classify reports an empty range when git fails, bypassing its own git_failed path
`gaia wiki commit-classify` reports an empty commit range when git fails, instead of the structured error its caller is already written to emit.

## Location

`.gaia/cli/src/wiki/util/git.ts:283`, inside `commitDetails`:

```ts
const raw = tryRunGit([...], {cwd});

if (raw === null) return [];
```

`tryRunGit` (`.gaia/cli/src/wiki/util/git.ts:30`) is a bare `catch { return null }`, so every git failure reaching `commitDetails` becomes `[]`, which is byte-for-byte what a genuinely empty `<since>..HEAD` range returns.

## Failure mode

The single caller, `.gaia/cli/src/wiki/commit-classify.ts:601`, already wraps the call in the error path this deserves:

```ts
try {
  details = commitDetails(parsed.flags.since, repoRoot);
} catch (error) {
  structuredError({code: 'git_failed', ...});
  return EXIT_CODES.UNKNOWN_SUBCOMMAND;
}
```

Because `commitDetails` swallows the failure and returns a value rather than throwing, **that `git_failed` branch is unreachable for the failure it was written for**. The command instead exits 0 and reports `evaluated: 0`, i.e. "there is nothing to review", which is a wrong answer rather than a loud one. Anything downstream of `/gaia-wiki sync` that trusts a zero count treats a transient git failure as a clean, up-to-date wiki.

Observed input/state → outcome: a `git log --no-merges --reverse --shortstat --format=... <sha>..HEAD` invocation that fails to run (spawn/exec failure under load is the observed trigger; any non-zero git exit does the same) → `commitDetails` returns `[]` → `health.evaluated === 0` → exit 0, no error on stderr.

## Evidence it is live, not theoretical

It surfaced as a CI flake on PR #1124. The **identical tree** `7ac1f2aa2fa591598ef0a80756c8a211e1cec00d` passed `cli-tests.yml` at 15:46 (sha `002ab22e`) and failed at 16:07 (sha `0ce76868`, an empty audit-stamp commit over the same tree):

```
FAIL src/wiki/commit-classify.test.ts > wiki commit-classify > health signal
     > a large healthy sample reports neither inert nor a warning
AssertionError: expected +0 to be 25
```

That is `.gaia/cli/src/wiki/commit-classify.test.ts:646`, which commits 25 commits to a sandbox repo and asserts `health.evaluated` is 25. A re-run of the same job passed clean. 129 test files spawn git subprocesses in parallel vitest workers, so a transient spawn failure under CI load is the plausible trigger; the swallow is what converts it into a silent wrong answer instead of a diagnosable one.

Every sibling test built on the same `commitManyTo` / `classify` helpers shares the exposure, so the flake is not confined to the one test that happened to fire.

## Suggested fix

Propagate the failure instead of swallowing it, so the caller's existing `git_failed` handler fires: have `commitDetails` call `runGit` directly rather than `tryRunGit`, letting the exception reach `.gaia/cli/src/wiki/commit-classify.ts:603`. That handler already emits the right structured error and exit code, so no new error surface is needed.

Do **not** change `tryRunGit` itself: `shortSha` (`.gaia/cli/src/wiki/util/git.ts:47`) depends on its null-on-failure contract as a legitimate fallback.

A regression test should assert that a failing git invocation produces the `git_failed` structured error and a non-zero exit, rather than an empty, successful classification.

Handler: prompt

stevensacks:
## What

`commitDetails` called `tryRunGit`, whose bare `catch { return null }` it turned into `[]`. That is byte-for-byte what a genuinely empty `<since>..HEAD` range returns, so `gaia wiki commit-classify` answered a transient git failure with `evaluated: 0` and exit 0, i.e. "there is nothing to review". Its single caller already wraps the call in a `git_failed` structured-error handler, which was therefore unreachable for the failure it was written for.

The fix is the issue's own suggested one: call `runGit` so the exception reaches that handler. `tryRunGit` is deliberately unchanged, because `shortSha` depends on its null-on-failure contract as a legitimate fallback.

## Honest limit

This **renames** the CI flake rather than removing it. A transient spawn failure under parallel-vitest load now reports `git_failed` and a non-zero exit instead of a silent wrong answer. That is the value during a long drain: telling "CI flake" from "your change broke it" at a glance.

## Verification

The new regression test was **proven able to fail** before being relied on: RED before the fix with the issue's exact symptom (exit 0 and `{"commits":[],"health":{...,"evaluated":0,...}}` on stdout), GREEN after. It triggers the failure with a well-formed 40-hex sha absent from the sandbox repo, so no mocking of `execFileSync` is involved.

Quality Gate run as written from `wiki/decisions/Quality Gate.md`:

| Step | Result |
| --- | --- |
| Simplify | 4 angles; 1 endorsed change applied |
| Localization | n/a, no user-facing strings |
| Type checking | `pnpm typecheck` 0; `pnpm -C .gaia/cli typecheck` 0 |
| Linting | `pnpm lint` 0; `pnpm -C .gaia/cli lint` 0 |
| Unit tests | CLI 129 files / 1865 passed; app 29 files / 155 passed; zero console warnings |
| E2E tests | 8 passed |
| Dev server | HTTP 200 |
| Build | passed |

`bash .gaia/tests/shell-lint.sh` passed. No bats suite exercises the touched paths (the only `.gaia/tests/` references to `commit-classify` are a README and a TSV fixture), so the CLI vitest suite is the guard here.

## Waived findings

One out-of-scope finding, surfaced by this PR's own simplify review and verified by hand. It is non-security and its path is a file this pull request already changes, so per `wiki/concepts/PR Merge Workflow.md` (`#### Cross-remit findings`) it is recorded here rather than filed.

**`.gaia/cli/src/wiki/util/git.ts` has no stated swallow-vs-throw policy, so two more helpers turn a git failure into a value indistinguishable from a legitimate result.**

- **`commitsAhead` (`git.ts:94`) returns `0` on git failure**, and `state.ts:240` maps `commitCount <= 0` to `drift_severity: 'none'`. So a failed `git rev-list --count` reports `commits_ahead: 0, drift_severity: 'none'`, a clean and up-to-date wiki, which is the signal the session-start hook and `/gaia-wiki sync` gate on. Sharper than it first looks: `state.ts:236-237` only calls `commitsAhead` when `reachable === true`, so its documented "Returns 0 if `sha` is unreachable" fallback is unreachable by construction from that caller, and the `0` fires **only** on git failure.
- **`fileListForCommit` (`git.ts:187`) returns `[]` on git failure**, once per commit inside `commitDetails`' map. A commit with `files: []` cannot match `classifyTouchedDomains`, so a genuinely WORTHY commit can be classified SKIP. The dispatched `code-audit-maintainer-node` member raised this independently and sharpened it: rules 3/4 fall through on the vacuous `touchesAny`, rule 5 returns a confident `SKIP` with `deferred` unset, so the dropped commit never registers in the `deferral_rate` / `inert` health signal either, and the sibling `touchesOnlyTests` (`commit-classify.ts:213`) already carries an explicit `files.length > 0 &&` guard against exactly this vacuous-truth hazard. It also named a third remedy beyond throw-vs-swallow: mark the resulting decision `deferred: true` so a swallowed failure at least reaches the `inert` signal. (`recentCommits` has the same shape but is display-only, so it is milder.)
- **The module docblock says "Errors are surfaced as thrown `Error`s"** while most exported helpers swallow, so it does not record the actual two-tier contract. It also says "returns trimmed stdout" while `runGit` does not trim. Both predate this PR.

Not folded into this diff: making `fileListForCommit` throw aborts a whole classification run over one unreadable commit, and with the member's `deferred: true` alternative there are now three defensible answers, so this is a real design fork rather than a mechanical fix. The drain plan's Invariant 8 forbids batching a design decision with a mechanical one. Adjacent but distinct: #1059 already covers `ancestorBefore` at `git.ts:79-88`, for its date-based baseline rather than its swallow.

## Note, not a finding

The committed bundle `.gaia/cli/gaia` still carries the pre-fix `tryRunGit` code, so `/gaia-wiki sync`, which runs the bundle rather than `tsx src`, keeps the old behavior until the next rebundle. That state is inherited from `main` and matches the established cadence here (src-only commits land, a catch-up rebundle follows), and `.github/workflows/release.yml:74` hard-fails a release whose committed binary differs from a fresh `pnpm -C .gaia/cli bundle`. Recorded rather than acted on: rebundling here is out of this PR's scope and the release gate already prevents it from shipping stale.

Closes #1135


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.UmiPtYKzzw

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.48CxUTZ9J6

Issue #1153, opened 2026-08-02
block-eslint-config-edit denies every eslint.config edit, including the reactRouter migration GAIA tells adopters to make


stevensacks:
`.claude/hooks/block-eslint-config-edit.sh:11` matches on filename alone, so it denies **every** edit to an `eslint.config.*` file. Its message states a narrower rule than it enforces: "Do not modify eslint.config.* to fix lint errors." It has no condition on lint state, on what the edit contains, or on whether the edit is a comment.

**Failure mode.** GAIA ships an action-required instruction that this hook stops Claude from carrying out. `CHANGELOG.md`'s `## [Unreleased]` block (#1049) tells adopters: "confirm your `eslint.config.mjs` spreads `...lint.reactRouter` after `...lint.react`, then run `pnpm lint`", and warns that a config which missed the new line starts failing `no-empty-pattern` on any route module that destructures nothing. An adopter who asks Claude to make that edit gets `BLOCKED` with a reason that does not describe their situation: they are not silencing a lint error, they are performing a documented migration. The hook is `manifest: owned` and registered in `.claude/settings.json:178`, so this reaches every adopter clone on the release that carries the note.

Encountered live while draining #1051: the `@gaia-react/lint` 2.0.0 bump left `.gaia/cli/eslint.config.mjs:8` enumerating the presets that config omits ("storybook, playwright, betterTailwind") without the fourth one 2.0.0 adds, `reactRouter`. The comment-only edit to complete that list was blocked, so the CLI config still carries the incomplete enumeration.

**Not a breakage, and the grade reflects that.** The human is not blocked, only the agent, and the workaround (edit by hand) is immediate. Graded `severity:suggestion` on that basis. What it costs is a confusing stop in the middle of a documented migration, plus doc drift like the instance above that never gets repaired because repairing it is blocked.

**Suggested fix.** Narrow the deny to what the message actually claims, or widen the message to match the deny. The design question the surrounding code does not settle is which discriminator to use, and the candidates are not equivalent:

- allow an edit whose diff touches only comments (repairs the doc-drift case; does nothing for the `reactRouter` migration)
- allow additive spreads of a `@gaia-react/lint` group while still denying rule overrides and `-- off` disables (covers the migration; needs the hook to parse the edit)
- gate on lint state, denying only while `pnpm lint` is red (closest to the stated intent; costs a lint run inside a PreToolUse hook, which is likely too slow)
[...]

Issue #1152, opened 2026-08-02
The .gaia/cli pnpm workspace does not inherit the root workspace's supply-chain hardening


stevensacks:
`.gaia/cli/pnpm-workspace.yaml:20` ends the file's settings without the two supply-chain controls the root workspace sets, so installs into the CLI workspace run unhardened.

**Verified live.**

```
$ pnpm -C .gaia/cli config list | grep -iE 'minimumReleaseAge|trustPolicy'
(no output)

$ pnpm config list | grep -iE 'minimumReleaseAge|trustPolicy'
  "minimumReleaseAge": 10080,
  "minimumReleaseAgeExclude": [
  "trustPolicy": "no-downgrade",
  "trustPolicyExclude": [
```

**Failure mode.** pnpm resolves the nearest `pnpm-workspace.yaml`, and `.gaia/cli` deliberately carries its own to stay a standalone workspace root, so it inherits nothing from the repo root. Root sets `minimumReleaseAge: 10080` (a 7-day vetting delay, commented there as supply-chain hardening) and `trustPolicy: no-downgrade` (fail the install when a package's trust level drops versus prior releases). Neither is in effect for `pnpm -C .gaia/cli install`, so that install can resolve a version published minutes ago, and a provenance downgrade in the CLI's ~700-package dev closure does not fail it. The exposure is at lockfile-write time, when a maintainer bumps a CLI dependency; CI's `--frozen-lockfile` install does not re-resolve and so inherits whatever the unhardened resolve committed.

Not theoretical: bumping one pin (`@gaia-react/lint` 1.9.0 -> 2.0.0, #1051) moved **158 distinct packages** in `.gaia/cli/pnpm-lock.yaml` through exactly this path.

**Why this reads as an oversight rather than a decision.** The file reasons explicitly about root parity for the setting immediately above the citation: `strictPeerDependencies: false` is justified as "for parity with the root workspace's explicit setting". The two supply-chain settings are omitted with no comment either way.

[...]

PR #1154, opened 2026-08-02
chore(cli): align the .gaia/cli lint pin with the root workspace

### LINKED ISSUE #1051: debt(cli): .gaia/cli lint pin drifts from root workspace (1.9.0 vs 2.0.0)
`.gaia/cli/package.json:22` pins `"@gaia-react/lint": "1.9.0"` while the root workspace is at `"2.0.0"` (`package.json:94`). The CLI is a separate pnpm workspace with its own lockfile, so nothing forces the two to agree and nothing reports when they don't.

**Failure mode.** The CLI's lint rules silently diverge from the rest of the project. `.gaia/cli/eslint.config.mjs` consumes the same presets (`base`, `react`, `testing`, `styleHygiene`, `guardrails`, `prettier`), so every preset change since 1.9.0 applies to `app/` but not to `.gaia/cli/src/`. CLI source is currently linted by a rule set two minors and one major behind: it never received the 1.10.0 or 1.11.0 rule changes, and will not receive any 2.x change either.

The drift is invisible and self-widening. Both workspaces lint green against their own pinned version, so CI reports nothing, and the gap grows on every root bump because no check compares the two pins. It has already grown once unnoticed: the skew predates this finding (1.9.0 vs 1.11.0 before #1049) and widened to a major when the root moved to 2.0.0.

**This is not a 2.0.0 blocker.** The 2.0.0 major is a no-op for this workspace. Its breaking change is the extracted `reactRouter` group, and the CLI has no route modules (`find .gaia/cli -path "*/routes/*" -name "*.tsx" -not -path "*/node_modules/*"` returns nothing), so it needs no `...lint.reactRouter` spread and would see no new errors from the extraction itself. The debt is the unmanaged drift, not this particular version gap.

**Suggested fix.** Bump `.gaia/cli/package.json` to `2.0.0`, run `pnpm -C .gaia/cli lint`, and fix whatever the two intervening minors surface. Then close the loop so it does not reopen on the next root bump: either bump the CLI pin alongside the root one in `/update-deps`, or add a check that fails when the two pins disagree. The bump alone clears today's gap and nothing more.

Handler: prompt

stevensacks:
Closes #1051

`.gaia/cli` is its own pnpm workspace root with its own lockfile, so nothing forced its `@gaia-react/lint` pin to agree with the root workspace's and nothing reported when it did not. The gap had reached two minors and a major (1.9.0 vs 2.0.0): both workspaces linted green against their own pin, so CI said nothing while `.gaia/cli/src` was checked by a rule set the rest of the repo had left behind.

## What changed

- **`.gaia/cli/package.json`**: `@gaia-react/lint` 1.9.0 → 2.0.0, plus the lockfile the install wrote.
- **`.gaia/cli/src/lint-pin-parity.test.ts`** (new): asserts both workspaces declare `@gaia-react/lint` in `devDependencies` and pin the same version.
- **`.github/workflows/cli-tests.yml`**: root `package.json` joins the `code` paths filter.

## The bump is a no-op for lint output, verified rather than assumed

The issue predicted this and it holds for all three versions in the gap. 1.10.0 bans a bare `@conform-to/zod` import and the CLI has no conform; 1.11.0 stops ignoring `.playwright/` and `.storybook/` and the CLI has neither; 2.0.0's breaking change extracts an opt-in `reactRouter` group the CLI does not spread, correctly, because it has no route modules (`find .gaia/cli -path "*/routes/*" -name "*.tsx"` returns nothing, so the group would relax a rule over an empty set).

Verified with `eslint --no-cache` over 277 files: **zero errors, zero warnings**. The `--no-cache` matters. The first run reused `node_modules/.cache/eslint` and returned green in a way that would have proved nothing.

## Why a deterministic guard rather than `/update-deps` prose

The issue offers either. Three things in the code pick the guard:

1. `/update-deps`'s SKILL.md **ships to adopters**, and adopters have no `.gaia/cli/package.json` (release-excluded, `.gaia/release-exclude:123`). The instruction would be maintainer-only noise in an adopter-shipped skill, and it is model-executed rather than enforced.
2. `.gaia/cli/src` and `.gaia/cli/package.json` are both release-excluded, so a test in the CLI vitest suite is maintainer-only by construction and costs an adopter nothing. `Vitest (.gaia/cli)` is a declared-required context, so a failure blocks the merge.
3. **The guard has to fire on the PR that causes the drift, and nothing would have.** A root-only bump is what breaks parity, `tests.yml:73-84` skips its entire suite for the `chore(deps):` title `/update-deps` opens, and `cli-tests.yml`'s filter covered `.gaia/cli/**` but not root `package.json`. Hence the filter entry. Without it the guard would first fire on some later, unrelated `.gaia/cli/**` PR, contaminating that PR with an unrelated pin bump and its lint fallout.

The two presence tests exist so the parity assertion cannot pass vacuously by comparing `undefined` to `undefined`.

## Guard proofs

Every new assertion was proven able to fail before being relied on, and re-proven after the Quality Gate's Simplify step changed the implementation:

| Fixture | Result |
|---|---|
| CLI pin set back to `1.9.0` | RED, `AssertionError: expected '1.9.0' to be '2.0.0'` |
| key deleted from `.gaia/cli/package.json` | RED, 2 failed |
| key deleted from root `package.json` | RED, 2 failed |
| pin moved to `dependencies` | RED, 2 failed |

Each fixture was reverted and the restored tree re-verified. `bash .gaia/tests/shell-lint.sh` passes; no bats suite exercises the touched paths.

**One fixture bit back, worth knowing.** The move-to-`dependencies` fixture was run through `pnpm exec vitest`, and pnpm 11 auto-synced `pnpm-lock.yaml` to the mutated manifest, relocating the `@gaia-react/lint` importer entry and moving an unrelated `@napi-rs/wasm-runtime` line. `git checkout -- .gaia/cli/package.json` restored the manifest but not the lockfile, so the residue survived the revert. Caught by re-reading `git status` rather than trusting the revert, and confirmed clean afterwards with `pnpm -C .gaia/cli install --frozen-lockfile`, which is what CI runs.

## Quality Gate

Run from `wiki/decisions/Quality Gate.md`, all eight steps.

| Step | Result |
|---|---|
| Simplify | ran; one change applied (`readPin` narrowed to `devDependencies`, reads hoisted) |
| Localization | n/a, no user-facing strings |
| Typecheck | root 0 errors; `.gaia/cli` 0 errors |
| Lint | root 0/0; `pnpm lint:cli` 0/0 (step 4's `.gaia/cli/**` clause) |
| Unit tests | root 29 files / 155 passed, **0 console warnings**; CLI 130 files / 1868 passed |
| E2E | 8 passed |
| Dev smoke | HTTP 200 |
| Build | passed, no warnings |

## CHANGELOG: no entry, and the bundle proves it

Every changed path is release-excluded, checked against `.gaia/manifest.json` rather than assumed: `.gaia/cli/package.json`, `.gaia/cli/pnpm-lock.yaml`, `.gaia/cli/src/**`, and `.github/workflows/cli-tests.yml` are all absent from the manifest. The one path in this area that does ship, `.gaia/cli/gaia`, is **byte-identical** after `pnpm -C .gaia/cli bundle`, as is `gaia-maintainer` and the `templates/` tree. Nothing an adopter can observe changes, so no `## [Unreleased]` entry is owed.

## Honest limits

**The lockfile churn is wide: 158 distinct packages moved**, 411 insertions / 394 deletions. All of it is transitive dev-only deps of the lint closure re-resolving (typescript-eslint 8.62.1 → 8.65.0, storybook 10.4.6 → 10.5.5, oxc-resolver bindings, babel patches). The CLI's own direct pins are unchanged apart from the lint one, the runtime deps (`js-yaml`, `zod`) did not move, and the byte-identical bundle is the evidence that none of it reaches a shipped artifact.

**`.gaia/cli/eslint.config.mjs:8` still enumerates the omitted presets without `reactRouter`**, which 2.0.0 adds. The comment-only edit to complete that list was blocked by `.claude/hooks/block-eslint-config-edit.sh`, which denies every edit to an `eslint.config.*` file by filename. Filed as #1153 rather than fought.

## Out-of-scope findings

**Filed:**

- **#1152** (`severity:suggestion`, `difficulty:medium`) — `.gaia/cli/pnpm-workspace.yaml` does not inherit the root's supply-chain hardening. Root sets `minimumReleaseAge: 10080` and `trustPolicy: no-downgrade`; the CLI workspace sets neither, verified with `pnpm -C .gaia/cli config list`. This PR is the live instance: it moved 158 transitive packages through that unhardened path. Untouched by this PR, so not waive-eligible.
- **#1153** (`severity:suggestion`, `difficulty:hard`) — the eslint-config hook above. It ships (`manifest: owned`) and is registered at `.claude/settings.json:178`, and the `## [Unreleased]` block already carries an **Action required** telling adopters to spread `...lint.reactRouter` in `eslint.config.mjs`, which is precisely the edit the hook denies.

**Waived under `wiki/concepts/PR Merge Workflow.md` `#### Cross-remit findings`**, because the path is one this PR already changes:

- `.gaia/cli/package.json:24` and below: five more shared pins drift from root, `@types/node` (25.6.1 vs 26.1.1), `eslint` (9.39.4 vs 9.39.5), `prettier` (3.9.1 vs 3.9.6), `tsx` (4.22.4 vs 4.23.1), `vitest` (4.1.5 vs 4.1.10). Deliberately left, and the parity test deliberately does not assert them: `@gaia-react/lint` names the **rule set**, which is #1051's stated failure mode, while these are tool versions that drift without changing what either workspace considers an error. Asserting all six would fail red today and force a much larger bump than this step. Recorded here rather than filed because this PR changes the file.

## Audit rounds

**Round 1** (tree `f30f58eb`) dispatched `code-audit-github-workflows` and `code-audit-maintainer-node`, both synchronously, neither with a `model` override so their frontmatter pin held. The set was non-empty and the filtered and `--no-carry-forward` outputs agreed, so it was not `#1055`'s fail-open shape. Both returned clean and both classified `real` under `audit-noop-detect.sh` rather than being trusted on sight.

`code-audit-maintainer-node` raised one Suggestion: this PR's own guard overclaimed its coverage. **Verified before applying**, per the merge workflow's warning that a Suggestion can assert an inferred mechanism. It holds exactly: `eslint-config-airbnb-extended@3.1.0` declares `typescript-eslint: ^8.58.0`, `@gaia-react/lint` pins airbnb-extended but not `typescript-eslint`, neither workspace declares it directly, and the two lockfiles resolve **8.59.2 (root) vs 8.65.0 (CLI)** today. So the guard narrows the drift class rather than closing it, and the docblock said otherwise.

Digest economics say accept-and-note once a member is already marked, since folding a fix in buys a full re-dispatch. Paid it anyway: a false "this class is closed" claim inside the guard's own docblock is the kind a later reader acts on. Fixed in `8fe4b44f`, which rotated only `code-audit-maintainer-node`'s digest.

**Round 2** (tree `54ddf1a8`) re-dispatched `code-audit-maintainer-node` alone. `code-audit-github-workflows` kept digest `919f7290…` and its round-1 marker stayed valid, exactly as per-member digest keying predicts, so re-dispatching it would have been waste. Round 2 returned clean and independently re-verified every factual claim the rewritten docblock makes.

Its one Suggestion is **filed as #1155**: the transitive `typescript-eslint` divergence is real, pre-dates this branch (root `8.59.2` / CLI `8.62.1` before the install floated the caret to `8.65.0`), and nothing but a docblock tracks it. Filed rather than waived because its anchor is root `pnpm-lock.yaml`, which this PR does not touch, and because the finding is explicitly not fixable here, so a PR-body waive would have evaporated with the branch.

### DIFF
--- a/.gaia/cli/package.json
+++ b/.gaia/cli/package.json
@@ -19,7 +19,7 @@
     "zod": "4.4.3"
   },
   "devDependencies": {
-    "@gaia-react/lint": "1.9.0",
+    "@gaia-react/lint": "2.0.0",
     "@types/js-yaml": "4.0.9",
     "@types/node": "25.6.1",
     "esbuild": "0.28.1",
--- a/.gaia/cli/pnpm-lock.yaml
+++ b/.gaia/cli/pnpm-lock.yaml
@@ -16,8 +16,8 @@ importers:
         version: 4.4.3
     devDependencies:
       '@gaia-react/lint':
-        specifier: 1.9.0
-        version: 1.9.0(@testing-library/dom@10.4.1)(@typescript-eslint/eslint-plugin@8.62.1(@typescript-eslint/parser@8.62.1(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(@typescript-eslint/parser@8.62.1(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(@typescript-eslint/utils@8.62.1(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(eslint-import-resolver-node@0.3.10)(eslint-plugin-import-x@4.17.1(@typescript-eslint/utils@8.62.1(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(eslint-import-resolver-node@0.3.10)(eslint@9.39.4(jiti@2.7.0)))(eslint-plugin-import@2.32.0)(eslint@9.39.4(jiti@2.7.0))(prettier@3.9.1)(storybook@10.4.6(@testing-library/dom@10.4.1)(prettier@3.9.1)(react@19.2.7))(stylelint@17.14.0(typescript@6.0.3))(tailwindcss@4.3.2)(typescript@6.0.3)(vitest@4.1.5(@types/node@25.6.1)(vite@8.0.16(@types/node@25.6.1)(esbuild@0.28.1)(jiti@2.7.0)(tsx@4.22.4)(yaml@2.9.0)))
+        specifier: 2.0.0
+        version: 2.0.0(@testing-library/dom@10.4.1)(@typescript-eslint/eslint-plugin@8.65.0(@typescript-eslint/parser@8.65.0(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(@typescript-eslint/parser@8.65.0(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(@typescript-eslint/utils@8.65.0(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(eslint-import-resolver-node@0.3.10)(eslint-plugin-import-x@4.17.1(@typescript-eslint/utils@8.65.0(eslint@9.39.4(jiti@2.7.0))(typescript@6.0.3))(eslint-import-resolver-node@0.3.10)(eslint@9.39.4(jiti@2.7.0)))(eslint-plugin-import@2.32.0)(eslint@9.39.4(jiti@2.7.0))(prettier@3.9.1)(storybook@10.5.5(prettier@3.9.1)(react@19.2.8))(stylelint@17.14.1(typescript@6.0.3))(tailwindcss@4.3.3)(typescript@6.0.3)(vitest@4.1.5(@types/node@25.6.1)(vite@8.0.16(@types/node@25.6.1)(esbuild@0.28.1)(jiti@2.7.0)(tsx@4.22.4)(yaml@2.9.0)))
       '@types/js-yaml':
         specifier: 4.0.9
         version: 4.0.9
@@ -66,8 +66,8 @@ packages:
     resolution: {integrity: sha512-RgHBCvtjbOK2gXSNBNIkNoEc9qoVEtau3hj8gEqKQuL3HZAibKarWFEI3Lfm6EYKkLalOh8eSrj9b+ch9H/VBA==}
     engines: {node: '>=6.9.0'}
 
-  '@babel/generator@7.29.7':
-    resolution: {integrity: sha512-DkXD5OJQaAQIdZ1bt3UZdEnHAn9Imd3IVBdX03UFe+ony9Ojw5pzr9YVKGDY1jt+Gcn/FnGkNf8r+Vj5NOJWtQ==}
+  '@babel/generator@7.29.8':
+    resolution: {integrity: sha512-gZbepsdh3WDtgZKWL+vTPh71LSBrm/Y4/QDZBVCcYfmeTEEuoOYwlSy+G1StfJg+/Zy550u/3TATbm7qDbbMtg==}
     engines: {node: '>=6.9.0'}
 
   '@babel/helper-compilation-targets@7.29.7':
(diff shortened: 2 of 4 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.6nsOqL4xeY

Issue #1157, opened 2026-08-02
release bump dies with ENOBUFS: bump.ts's defaultRunner sets no maxBuffer either, so the release is blocked one step before #1147


stevensacks:
**Location:** `.gaia/cli/src/release/bump.ts:65` (`defaultRunner`)

## Failure mode

`gaia-maintainer release bump` fails outright on this repository today. Reproduced live on the current tree:

```
$ .gaia/cli/gaia-maintainer release bump
bump: git log failed: spawnSync git ENOBUFS
```

This is #1147's defect in the second file that carries it. `bump.ts`'s own `defaultRunner` (`:65-70`) calls `spawnSync` with no `maxBuffer`, so Node's 1 MiB default applies:

```ts
export const defaultRunner: CommandRunner = (command, args, options) =>
  spawnSync(command, args as string[], {
    cwd: options.cwd,
    encoding: 'utf8',
    stdio: ['ignore', 'pipe', 'pipe'],
  });
```

`collectCommits` (`bump.ts:235`) runs the identical `git log --no-merges --format=%s%n%b%n---END-COMMIT--- <lastTag>..HEAD` that `changelog.ts` does. The body (`%b`) is included, so output scales with commit-message length rather than commit count. Measured for `v1.6.1..HEAD`: 1,248,452 bytes, past the 1 MiB limit.

The trigger is distance from the last tag, so it worsens until the next release cuts a tag, and it is worst exactly when a release is due.

It fails **closed**: `expectSuccess` throws and `run` reports a structured error, so the consequence is a blocked release rather than a wrong version proposal. Both `release bump` forms reach `collectCommits` before they diverge, so `--auto` is affected identically.

**Why this is separate from #1147.** #1147 fixes `changelog.ts:58`, which is release runbook step 7. This is step 5. Fixing only #1147 leaves the runbook blocked one step earlier, so the two must both land for a release to proceed.

## Suggested fix

The same shape #1147 uses: a named `maxBuffer` constant on this file's own `defaultRunner`, at or above 32 MiB, matching the bound the rest of the CLI already applies to git spawns.

- `.gaia/cli/src/release/changelog.ts` — `MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024`
- `.gaia/cli/src/wiki/util/git.ts:16` — `MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024`
[...]

PR #1158, opened 2026-08-02
fix(cli): bound the release changelog git spawn (#1147)

### LINKED ISSUE #1147: release changelog dies with ENOBUFS: defaultRunner sets no maxBuffer and the git log since the last tag is 1.2 MB
**Location:** `.gaia/cli/src/release/changelog.ts:58` (`defaultRunner`)

## Failure mode

`gaia-maintainer release changelog` fails outright on this repository today. Reproduced live on `main` at `e81232e0`:

```
$ .gaia/cli/gaia-maintainer release changelog --draft
changelog: git log failed: spawnSync git ENOBUFS
```

`defaultRunner` (`changelog.ts:58-63`) calls `spawnSync` with no `maxBuffer`, so Node's 1 MiB default applies:

```ts
export const defaultRunner: CommandRunner = (command, args, options) =>
  spawnSync(command, args as string[], {
    cwd: options.cwd,
    encoding: 'utf8',
    stdio: ['ignore', 'pipe', 'pipe'],
  });
```

`collectCommits` (`changelog.ts:206`) runs `git log --no-merges --format=%s%n%b%n---END-COMMIT--- <lastTag>..HEAD`. The body (`%b`) is included, so the output scales with commit-message length, not just commit count. Measured for `v1.6.1..HEAD`: **461 commits, 1,242,652 bytes**, past the 1 MiB limit.

The trigger is distance from the last tag, so this gets worse until the next release cuts a tag, and it is worst exactly when a release is due. Both `release changelog` paths are affected: `--draft` and the graduating form both reach `collectCommits` before they diverge.

It fails **closed**, not open. `collectCommits` checks `result.error` and throws, and `run` reports a structured error. So the consequence is a blocked release with an accurate message, not a silently truncated changelog.

Same defect class as #1126 (`regen-regions` spawn bounds), different file. `lastTag` and any other `defaultRunner` caller share the same unbounded runner, though only `collectCommits` produces output at this scale today.

## Suggested fix

Give `defaultRunner` an explicit `maxBuffer`, matching what the rest of the CLI already does for git spawns:

- `.gaia/cli/src/update/regen-regions.ts:933` — `SPAWN_MAX_BUFFER_BYTES = 32 * 1024 * 1024`, passed at `:949`
- `.gaia/cli/src/util/git-env.ts:40` — `maxBuffer: 64 * 1024 * 1024`
- `.gaia/cli/src/wiki/util/git.ts:23` — `maxBuffer: MAX_GIT_BUFFER_BYTES`

A named constant in the same shape as those, at or above 32 MiB, clears the current 1.2 MB by a wide margin.

Worth pairing with a regression test that drives `collectCommits` through a runner returning an `ENOBUFS` error, asserting the thrown message, since nothing currently exercises the failure arm.

Handler: prompt

stevensacks:
Closes #1147

## What

`release changelog` reads every commit subject **and body** since the last tag, so its `git log` output grows with the distance from that tag. `defaultRunner` declared no `maxBuffer`, so Node's 1 MiB `spawnSync` default applied and the command failed closed with `ENOBUFS` once the range passed it, which is worst exactly when a release is due.

Reproduced on `main` before the fix, and it is not marginal: 461 commits and **1,248,452 bytes** at `v1.6.1..HEAD`.

```
$ .gaia/cli/gaia-maintainer release changelog --draft
changelog: git log failed: spawnSync git ENOBUFS
```

After: exit 0, 376 lines of rendered block, empty stderr, against that same 1.2 MB log.

## How

- `MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024` on the runner rather than on `collectCommits`'s call site, so `lastTag` and every other caller of this file's runner are covered by one bound.
- 64 MiB and that name match `wiki/util/git.ts:16` and `util/git-env.ts:40`, the CLI's two other git-output readers. `regen-regions.ts:933`'s 32 MiB is the region-digest spawner, a different job.
- File-local, like all three precedents. A shared export would mean touching three other files that sit at two different values, which is outside this fix.

## Guard proofs

Both new tests were proven able to fail before being relied on.

- `defaultRunner reads child stdout past the 1 MiB spawnSync default` — RED before the fix with the exact production symptom (`Error: spawnSync … ENOBUFS`, `code: 'ENOBUFS'`), GREEN after. It drives a real child that writes 2 MiB, so it fails if the bound is removed by any means, not just if the constant is renamed.
- `collectCommits throws with the spawn error rather than returning no commits` — the failure arm the issue notes is unexercised. Proven by mutation: replacing the `throw` with `return []` turns it RED; the file was restored byte-identical afterwards.

`bash .gaia/tests/shell-lint.sh` is not implicated (no shell changed) and no bats suite touches these paths.

## Quality Gate

All eight steps from `wiki/decisions/Quality Gate.md`. Root and CLI typecheck clean; root lint clean and `lint:cli` clean; CLI 130 files / 1870 tests, root 29 / 155, zero console warnings; Playwright 8 passed; dev smoke HTTP 200; production build clean.

`.gaia/cli/gaia-maintainer` is rebundled. The adopter `gaia` binary is byte-identical, checked rather than assumed.

## CHANGELOG

No entry. `.gaia/cli/src` and `.gaia/cli/gaia-maintainer` are both release-excluded (`.gaia/release-exclude:121,133`) and absent from `.gaia/manifest.json`; the one shipped path in the area, `.gaia/cli/gaia`, is unchanged. Nothing here is adopter-observable.

## Out-of-scope finding, filed

**#1157**: `bump.ts:65` carries the identical unbounded `defaultRunner`, and `bump.ts:235` runs the same `git log --format=%s%n%b`, so `gaia-maintainer release bump` dies with the same `ENOBUFS` on this tree today. That is release runbook **step 5**, one step ahead of the step 7 this PR fixes, so the runbook needs both.

Filed rather than folded in: `bump.ts` is untouched by this PR, so the touched-file waive in `wiki/concepts/PR Merge Workflow.md` `#### Cross-remit findings` does not apply; and raising a spawn bound is a behavior change, which the drain plan's batching rule does not license across files.

`preflight.ts:43` and `commit-and-tag.ts:52` declare the same unbounded runner but read `git log --format=%s`, `rev-parse`, and `status --porcelain`, which are orders of magnitude smaller. Not at risk, noted in #1157 so the next reader does not re-derive it.

### DIFF
--- a/.gaia/cli/gaia-maintainer
+++ b/.gaia/cli/gaia-maintainer
@@ -17674,9 +17674,11 @@ var HELP_TEXT20 = `Usage: gaia-maintainer release changelog [--draft] [--version
 `;
 var HELP_TOKENS20 = /* @__PURE__ */ new Set(["--help", "-h", "help"]);
 var UNEXPECTED_EXIT10 = 2;
+var MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024;
 var defaultRunner2 = (command, args, options) => spawnSync2(command, args, {
   cwd: options.cwd,
   encoding: "utf8",
+  maxBuffer: MAX_GIT_BUFFER_BYTES,
   stdio: ["ignore", "pipe", "pipe"]
 });
 var takeValue7 = (argv, index, flag) => {
@@ -21721,13 +21723,13 @@ import path21 from "node:path";
 
 // src/wiki/util/git.ts
 import { execFileSync as execFileSync3 } from "node:child_process";
-var MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024;
+var MAX_GIT_BUFFER_BYTES2 = 64 * 1024 * 1024;
 var runGit = (args, options = {}) => {
   const cwd = options.cwd ?? process.cwd();
   const result = execFileSync3("git", args, {
     cwd,
     encoding: "utf8",
-    maxBuffer: MAX_GIT_BUFFER_BYTES,
+    maxBuffer: MAX_GIT_BUFFER_BYTES2,
     stdio: ["ignore", "pipe", "pipe"]
   });
   return result.toString();
--- a/.gaia/cli/src/release/changelog.test.ts
+++ b/.gaia/cli/src/release/changelog.test.ts
@@ -8,6 +8,8 @@ import {mkdtempSync, readFileSync, rmSync, writeFileSync} from 'node:fs';
 import {tmpdir} from 'node:os';
 import path from 'node:path';
 import {
+  collectCommits,
+  defaultRunner,
   graduateChangelog,
   groupCommits,
   renderBlock,
@@ -48,6 +50,16 @@ const failResult = (status: number): SpawnSyncReturns<string> => ({
   stdout: '',
 });
 
+const spawnErrorResult = (error: Error): SpawnSyncReturns<string> => ({
+  error,
+  output: ['', '', ''] as never,
+  pid: 0,
+  signal: null,
+  status: null,
+  stderr: '',
+  stdout: '',
+});
+
 const buildLogOutput = (
   commits: {body?: string; subject: string}[]
 ): string => {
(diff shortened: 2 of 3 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.yYkGN6wkUF

PR #1160, opened 2026-08-02
fix(cli): bound the release bump git spawn (#1157)

### LINKED ISSUE #1157: release bump dies with ENOBUFS: bump.ts's defaultRunner sets no maxBuffer either, so the release is blocked one step before #1147

stevensacks:
## What

`gaia-maintainer release bump` fails closed on this repository today:

```
$ .gaia/cli/gaia-maintainer release bump
bump: git log failed: spawnSync git ENOBUFS
```

`collectCommits` (`bump.ts:245`) reads every subject **and body** since the last tag.
Measured for `v1.6.1..HEAD`: **1,250,573 bytes** over 468 commits, past Node's 1 MiB
`spawnSync` default, which `bump.ts`'s own `defaultRunner` did not raise.

This is #1147's defect in the second file that carries it. Bumping is release runbook
**step 2** (`### 2. Apply the bump`); the changelog #1147 fixed is **step 5**
(`### 5. Graduate the CHANGELOG`). So #1158 alone did not unblock a release, and both had
to land.

> **Correction, made during the audit round.** This PR originally cited steps 5 and 7,
> inherited verbatim from #1157's issue body. Both numbers are wrong; the runbook
> (`.claude/commands/gaia-release.md`) numbers them 2 and 5. The substantive claim is
> unaffected and is in fact stronger: bump runs three steps ahead of the changelog rather
> than one. See the out-of-scope findings below for where the wrong numbers came from.

## Approach

Rather than adding a fourth copy of the 64 MiB literal, this hoists it to a shared
`.gaia/cli/src/util/git-buffer.ts` and adopts it at all four git-output readers, which is
what PR #1158's audit round asked for on this issue. The value is identical at every site,
so the three pre-existing readers are unchanged in behavior. `regen-regions.ts` keeps its
own 32 MiB bound: it spawns for region digests, not history reads.

## Verification

- **End-to-end**: `release bump` was exit 2 with `ENOBUFS`, now `v1.6.1 -> v1.7.0 (minor)`,
  exit 0. `release changelog --draft` still exit 0 with empty stderr, so both runbook steps
  are clear.
- **RED proof**: the real-child regression test failed before the fix with the production
  symptom verbatim (`code: 'ENOBUFS'`, `errno: -55`). It drives `defaultRunner` against a
  2 MiB child rather than asserting on the constant, so it fails if the bound is removed by
  any means, not only if the identifier changes.
- **Mutation proof**: the second test pins the fail-closed contract and is green in both
  directions, so it was proven by injecting a `return` ahead of `expectSuccess`'s throw.
  The mutation was confirmed applied with a diff before the run, and the source restored
  byte-identical after.
- **Adopter surface**: the hoist touches `wiki/util/git.ts` and `util/git-env.ts`, both in
  the adopter bundle, so `.gaia/cli/gaia` (manifest `owned`) changed. The bundle delta is
  only the constant's relocation at an identical value. `gaia wiki state --json` smoke-runs
  clean, and the distribution harness is 17/17.
- **Quality Gate**: all eight steps. Typecheck clean both workspaces; lint clean both; CLI
  130 files / 1872 tests, root 29 / 155, zero console warnings; Playwright 8; dev smoke
  HTTP 200; build clean. Plus `shell-lint` (193 scripts, 181 suites) and the distribution
  harness.

## Deliberately out of scope

`preflight.ts:43` and `commit-and-tag.ts:52` declare the same unbounded runner and are left
alone, with the issue's claim measured rather than trusted. Two independent measurements
agree it is not at risk: preflight's largest read is `git log --format=%s` (subjects only,
no bodies) at 36,700 bytes over `v1.6.1..HEAD`, and 71,722 bytes over the *entire*
repository history as the audit member measured it, ~14x under the 1 MiB default with no
body-carrying growth mechanism. `commit-and-tag` performs no history read at all. Raising a
spawn bound is not provably zero-behavior, so Invariant 9 does not license folding them in.

- **`.gaia/cli/src/release/changelog.ts:4`** — the module docblock opens `Step 7 of the
  maintainer release runbook`, but `release changelog` is `### 5. Graduate the CHANGELOG`.
  A maintainer reading the file to place it in the release flow gets a wrong answer from
  the line that exists to give the right one. This is the likely origin of #1157's own
  wrong "step 5 / step 7" claim, which propagated into this PR before the audit caught it.
  Waived rather than filed: this PR already changes `changelog.ts`.
  
  

The same defect in three files this PR does **not** touch (`preflight.ts:4`,
`scrub-wiki.ts:4`, `commit-and-tag.ts:4`) is filed as **#1161**. `bump.ts:4` and `:69`
carried it too and are corrected here, since this PR owns that file.

## CHANGELOG

No entry. `release bump` is maintainer-only (`.gaia/cli/gaia-maintainer` is release-excluded
and absent from the manifest), and the adopter binary's only delta is a constant moving
between modules at the same value. Nothing adopter-observable changed and no adopter action
is required.

Closes #1157

### DIFF
--- a/.gaia/cli/gaia
+++ b/.gaia/cli/gaia
@@ -14658,6 +14658,11 @@ var readAutomationConfig = (repoRoot) => {
 
 // src/util/git-env.ts
 import { execFileSync } from "node:child_process";
+
+// src/util/git-buffer.ts
+var MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024;
+
+// src/util/git-env.ts
 var execGaiaGitRaw = (args, cwd) => {
   const env = { ...process.env };
   delete env.GIT_DIR;
@@ -14669,11 +14674,10 @@ var execGaiaGitRaw = (args, cwd) => {
     env,
     // Node's 1 MiB default throws ENOBUFS on a large-output git command
     // (`git log` over a long history, `git status` on a very dirty tree).
-    // Matches `runGit` in wiki/util/git.ts, whose callers already hit that
-    // ceiling. The docblock above declares this the one chokepoint every
-    // git-shelling resolver routes through, so the ceiling belongs here
-    // rather than at the next caller to need it.
-    maxBuffer: 64 * 1024 * 1024,
+    // The docblock above declares this the one chokepoint every git-shelling
+    // resolver routes through, so the ceiling belongs here rather than at the
+    // next caller to need it.
+    maxBuffer: MAX_GIT_BUFFER_BYTES,
     stdio: ["ignore", "pipe", "pipe"]
   });
 };
@@ -28631,7 +28635,6 @@ var inspectWorkingTree = (cwd, runner = defaultRunner) => {
 
 // src/wiki/util/git.ts
 import { execFileSync as execFileSync5 } from "node:child_process";
-var MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024;
 var runGit2 = (args, options = {}) => {
   const cwd = options.cwd ?? process.cwd();
   const result = execFileSync5("git", args, {
--- a/.gaia/cli/gaia-maintainer
+++ b/.gaia/cli/gaia-maintainer
@@ -14657,6 +14657,11 @@ var readAutomationConfig = (repoRoot) => {
 
 // src/util/git-env.ts
 import { execFileSync } from "node:child_process";
+
+// src/util/git-buffer.ts
+var MAX_GIT_BUFFER_BYTES = 64 * 1024 * 1024;
+
+// src/util/git-env.ts
 var execGaiaGitRaw = (args, cwd) => {
   const env = { ...process.env };
   delete env.GIT_DIR;
@@ -14668,11 +14673,10 @@ var execGaiaGitRaw = (args, cwd) => {
     env,
     // Node's 1 MiB default throws ENOBUFS on a large-output git command
     // (`git log` over a long history, `git status` on a very dirty tree).
-    // Matches `runGit` in wiki/util/git.ts, whose callers already hit that
-    // ceiling. The docblock above declares this the one chokepoint every
-    // git-shelling resolver routes through, so the ceiling belongs here
-    // rather than at the next caller to need it.
-    maxBuffer: 64 * 1024 * 1024,
+    // The docblock above declares this the one chokepoint every git-shelling
+    // resolver routes through, so the ceiling belongs here rather than at the
+    // next caller to need it.
+    maxBuffer: MAX_GIT_BUFFER_BYTES,
     stdio: ["ignore", "pipe", "pipe"]
   });
 };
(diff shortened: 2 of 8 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.dafu1baiuX

PR #1168, opened 2026-08-02
fix(cli): name the roster parse failure instead of the roster membership (#1044)

### LINKED ISSUE #1044: An unparseable audit-ci.yml is swallowed into an empty auditor set, so the maintainer is sent to fix the roster instead of the YAML syntax error
**Location:** `.gaia/cli/src/release/region-registry.ts:59` (the `catch` around the `.gaia/audit-ci.yml` parse)

## Failure mode

```ts
try {
  parsed = parseYaml(readFileSync(rosterPath, 'utf8'));
} catch {
  return new Set();
}
```

The bare catch makes an **unparseable** roster indistinguishable from a roster that legitimately lists no auditors. Both produce an empty set.

That empty set then reaches `region-scan.ts`, which throws `unrewrittenMessage` — a message whose stated remedies are to re-add the member to `.gaia/audit-ci.yml`, or to delete the marker pair. Neither is the fix when the real cause is a YAML syntax error a few lines up in that same file.

The failure is loud, not silent: both `buildManifest` call sites catch and exit non-zero via `manifest_build_failed`. It is **misdirected**. The maintainer is handed a confident diagnosis of the wrong problem and sent to edit a file that is already broken in a way the message never mentions.

Tests cover the unparseable case's empty-set return, but not the downstream message the maintainer actually reads.

## Suggested fix

Distinguish absent from unparseable at this catch — return a discriminated result rather than an empty set on parse failure — and have `region-scan.ts` name the parse failure directly when the roster could not be read, instead of falling through to `unrewrittenMessage`.

The shape of that signal is a real choice (a thrown error carrying the YAML message, a sentinel value, or a result type), and the adjacent error-handling conventions in `release/` settle it.

Handler: prompt

stevensacks:
Closes #1044

## The defect

`rosterAgentPaths` wrapped the `.gaia/audit-ci.yml` read in a bare `catch` that returned an empty set. That made a roster the build could not read indistinguishable from one that legitimately lists no auditors. The empty set reached `region-scan.ts`, which threw `unrewrittenMessage`:

> `.claude/agents/code-audit-frontend.md` carries the `'audit-remit'` region's marker pair, but that region's regeneration command does not rewrite it. […] Fix by either: re-adding the corresponding member to `.gaia/audit-ci.yml`, or deleting the marker pair and its body from `.claude/agents/code-audit-frontend.md`.

Neither remedy fixes a YAML syntax error a few lines up in that same file. The failure was already loud; it was **misdirected**. Loud is not the same as correctly directed.

Reproduced by test before the fix: that message is the literal output the new `region-scan.test.ts` case captured on a red run.

## The fix

Three failure shapes are now told apart, and each throws with the roster named:

```
.gaia/audit-ci.yml is not valid YAML. The Code Audit Team agent definitions are read from
that roster, so the release manifest cannot be built until the file itself is fixed; this is
not a roster-membership or marker problem. missed comma between flow collection entries (2:3)

 1 | auditors: [
 2 |   - unterminated
-------^
```

- **cannot be read** (`EISDIR`, `EACCES`) — a separate arm on purpose: `EISDIR`'s own message is `EISDIR: illegal operation on a directory, read` and carries **no path at all**, so describing an IO failure as a syntax error would be a smaller copy of the misdirection being fixed.
- **not valid YAML** — carries js-yaml's own text, including its line and column.
- **no top-level mapping** — an empty, truncated, comment-only, bare-scalar, or top-level-list roster. Added after audit round 1; see below.

A roster that is **absent** still returns the empty set (declaring no auditors is legitimate), and so does a **mapping** that carries no `auditors` key or a non-list one.

## The decision this carried

The issue named the signal shape as a real choice (thrown error / sentinel / result type) and deferred to `release/`'s conventions. Those conventions split on **caller choice, not on "is this a reader"**: a discriminated result appears exactly where a caller has more than one non-exceptional continuation, and each instance says so in its own docblock (`readCommittedManifest`, `graduateChangelog`, `scanRegion`). Where the build must abort, `release/` throws, at ~15 sites.

This call site has one caller (`scanEntry` ← `scanRegionDeclarations` ← `buildManifest`) and one correct continuation, and its two siblings *in the same function* already throw into the same `manifest_build_failed` path. A result type would have changed `RegionRegistryEntry.rewrites` for every future entry to hand an abort-only caller a discriminant it never reads.

## What the audit changed

Round 1 (`code-audit-maintainer-node`) returned zero Critical, zero Important, and one Suggestion that was **right and worth the round**. It tested six roster shapes with `tsx` and found the original scoping decision was partly wrong: an empty, truncated, comment-only, bare-scalar, or top-level-list roster parses *successfully*, so it slipped past both new arms and still produced the original misdirection.

The deferral reasoning ("the nearest YAML spelling of lists no auditors") holds for a mapping with no `auditors` key. It does not hold for a file whose top level is not a mapping at all: a 0-byte roster declares nothing, so it belongs with the unreadable ones. Fixed rather than deferred, with its own test and mutant.

It also raised the CHANGELOG as a cross-remit finding, correctly — see below.

## Behavior change worth naming

In a tree where **no** shipped file carries the region markers, a broken roster builds a manifest today and aborts after this change. In this repo it already aborts, since five `.claude/agents/code-audit-*.md` carry the pair, so what changes here is *which* message the maintainer gets. `release` is maintainer-only, so no adopter path reaches this code.

## Verification

- **RED first**, and each guard proven able to fail on its own: reverting the parse arm fails 2 of 18, the read arm 1 of 18, the non-mapping guard 1 of 18. Three mechanisms, three mutants, no over-broad assertions.
- **Quality Gate**, all steps as written, re-run after the last edit. Typecheck and lint clean in both workspaces, zero warnings. CLI 130 files, 1875 passed and 1 skipped; root 29 files, 155 passed and 1 skipped; Playwright 8 passed and 1 skipped; dev smoke HTTP 200; build clean.
- Simplify ran 4 review agents; 2 findings applied (the `rewrites` throw contract, and a duplicated region id removed rather than synchronized).
- `shell-lint.sh` clean. Distribution harness **17/17**, including `16-audit-remit-parity.sh`.
- Adopter binary `.gaia/cli/gaia` verified **byte-identical** across every rebundle (`f1c67c01a4e8…`).

## CHANGELOG

**Entry added** under `### Fixed`. The first call here was "no entry", reasoning from the byte-identical adopter bundle; the audit flagged it and the audit was right. That argument came from S4, a comment-only change where it was reinforcing rather than load-bearing. The gate turns on whether the **product surface** moves, counting a bugfix in any shipped *or maintainer* surface, and in-repo precedent includes an entry that records byte-identical rebundled binaries and takes an entry anyway.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.4hYfh3DdQx

Issue #1131, opened 2026-07-31
update-gaia Step 7d names absent-pre-image as the revert-vs-report criterion, contradicting the in-scope creation rule below it


stevensacks:
`.claude/skills/update-gaia/SKILL.md:622` (Step 7d, "Writes are confined and backed up")

The bullet names the wrong discriminator for revert-versus-report, and the paragraph under it now states the opposite outcome from the same stated premise.

## Failure mode

Line 622 gives the rule as:

> a write inside the region's own directories is reverted to what it held before the run, and a write anywhere else in the tree, **which has no pre-image to restore from**, is reported instead

That parenthetical names *absence of a pre-image* as the reason a write is reported rather than reverted. The paragraph directly under it then says an in-scope creation, **"having no pre-image, is removed"**.

Both statements are individually correct about outcomes, but they attribute two opposite outcomes to the same stated cause, two clauses apart. A reader who takes 622's reason at face value derives "no pre-image, therefore reported" for a file the regeneration created inside the region's own directories, which is wrong: such a file is removed. An agent reasoning from the stated rule rather than memorising both outcome lists gets it backwards.

The real discriminator is **scope**: inside the snapshot there is a pre-run record of the directory's contents, so both a deletion and a creation can be acted on; outside it there is no record at all, so nothing can be. Absence of a pre-image *for that path* is a consequence, not the criterion.

Secondary: saying so required introducing "snapshot" as a term in the paragraph below, and that word appears nowhere else in this file's region prose, so the split is now stated twice on two different vocabularies.

## Suggested fix

Rewrite 622's clause to name scope as the criterion and drop the pre-image parenthetical, e.g. "a write anywhere else in the tree, where this run holds no record of what was there before, is reported instead". Then the paragraph below can stop re-stating the outside case in its own terms and simply enumerate the in-scope outcomes. Introduce "snapshot" once, where the mechanism is described, or avoid it and say "what this run recorded before the spawn" consistently in both places.

Handler: prompt

stevensacks:
**A second defect now sits in this same bullet (line 622).** Recorded here rather than as its own issue: the filing recipe dedups on `path=`+`line=` alone and this issue already holds `path=.claude/skills/update-gaia/SKILL.md line=622`, so a separate issue would be a duplicate by that rule. Whatever drains this one should expect to fix two independent claims in the bullet, not one.

Surfaced by PR #1175 (drain step S7) as a cross-remit finding from its `code-audit-maintainer-node` round.

[...]

PR #1169, opened 2026-08-02
fix(cli): keep the regen-regions sweep inside the repository

### LINKED ISSUE #1128: regen-regions snapshots through symlinked scope subdirectories, so a revert can write outside the repository
`.gaia/cli/src/update/regen-regions.ts:641` (`collectScopeDigests`)

The snapshot walk descends into symlinked subdirectories, and the sweep then writes back through lexical paths, so a revert can land outside the tree the mechanism exists to confine writes to.

## Failure mode

`collectScopeDigests` enumerates a region's scope directory with:

```ts
entries = readdirSync(absDir, {recursive: true}) as string[];
```

`readdirSync`'s recursive walk descends **into** a symlinked subdirectory. The per-entry `lstatSync` only refuses to follow the *final* component, so a regular file living behind a symlinked directory is stat-ed and read through the link and recorded under a lexical key such as `.claude/agents/linkdir/precious.txt`.

`sweepScope` then resolves that key with `path.resolve(root, relPath)`, which is also purely lexical, and performs `rmSync` / `writeFileSync` / `chmodSync` against it. Those calls resolve the symlinked component at the OS level, so the write lands on the link's real target, which can be anywhere on the filesystem, including outside the repository entirely. The after-only removal loop deletes through the same traversal.

Two things are wrong at once when this fires:

1. The module doc and the sweep's own comment both promise the runner never writes outside the region's directories. Through a symlinked scope subdirectory, it can.
2. `report.confined` names a repo-relative path that does not describe where the write actually landed, so the report is misleading even when the revert "succeeds".

Reaching it needs an adopter's region scope directory to contain a symlinked subdirectory. That is unusual, which is why this is a suggestion rather than an Important, but nothing in the code prevents it and the blast radius is arbitrary.

Pre-existing: the recursive walk and the lexical resolve both predate PR #1116, which only made symlinks visible to the sweep as entries in their own right. The audit member that surfaced it explicitly scoped it out of that PR's changed lines.

## Suggested fix

Stop trusting the lexical key at write time. Either of:

- Prune the walk at a symlinked directory instead of descending into it, so nothing behind a link is ever snapshotted, or
- Record `realpathSync.native(dirname)` alongside each entry at snapshot time and refuse the revert (reporting instead) when it no longer matches at sweep time.

Pruning is the smaller change and matches how the sweep already treats a link at the final position: recorded as itself, never followed. A test should create a scope subdirectory that is a symlink to a directory outside the root, have the regeneration program modify a file behind it, and assert the outside file is untouched and the entry is reported rather than restored.

Handler: prompt

### LINKED ISSUE #1130: regen-regions drops an unreadable symlink from the snapshot, so a transient readlink failure can delete an adopter's link
`.gaia/cli/src/update/regen-regions.ts:673` (`collectScopeDigests`)

The symlink arm records an entry only when `readlinkSync` succeeds; its `catch` skips the path entirely:

```ts
if (stat.isSymbolicLink()) {
  try {
    digests.set(repoRelative, {kind: 'symlink', target: readlinkSync(abs, 'buffer')});
  } catch {
    // skipped
  }
  return;
}
```

## Failure mode

Presence in the snapshot answers a different question from restorability, and this conflates them. `sweepScope`'s removal loop treats any path present in `after` but absent from `before` as something the spawn created, and unlinks it.

If `readlinkSync` fails transiently during the **before** pass (`EIO`, `ENOMEM`, a racing unlink) and succeeds during the **after** pass, the adopter's pre-existing symlink is absent from `before` and present in `after`. The removal loop then unlinks a link the regeneration command never touched and reports it as `{action: 'removed'}`, which reads to the adopter as the confinement mechanism cleaning up a stray write rather than as data loss.

Narrow: the trigger is a kernel-level transient between the two passes, not anything the regeneration program can cause. The regular-file arm a few lines below has the same shape and is the established deliberate pattern there, so this is a consistency gap rather than a live defect. Filing it because the two arms answer to different consequences: skipping an unreadable *file* loses a revert, while skipping an unreadable *link* can actively delete one.

## Suggested fix

Record presence even when the target cannot be read, and let the sweep refuse to restore it rather than refusing to see it. For example `{kind: 'symlink', target: undefined}`, with `sweepScope` pushing `{action: 'reported'}` for such an entry instead of restoring it. Presence-tracking is what keeps the removal loop from mistaking a pre-existing path for a spawn creation, and it is independent of whether the pre-image can be written back.

Worth considering the same treatment for the regular-file arm while in there, since the reasoning is identical, but that arm's current behavior is deliberate and predates this, so changing it is a separate decision.

Handler: prompt

### LINKED ISSUE #1134: regen-regions rewrote filter has no test for the symlink-declared-path case it was changed for
`.gaia/cli/src/update/regen-regions.ts:1110` (`runRegeneration`)

The `rewrote` filter was changed from a digest-only comparison to `isUntouched`, so a declared path that is a symlink is no longer reported as unchanged however the regeneration command repointed it. The behavior change is real and correct; it has no test.

## Failure mode

Not a live defect in the shipped behavior, a coverage hole that lets a correct fix silently regress. Reverting the filter to the old form:

```ts
const rewrote = decl.paths.filter(
  (declPath) => digestOf(before.get(declPath)) !== digestOf(after.get(declPath))
);
```

leaves the entire 62-test suite green while the misreport returns: a declared path that is a symlink has no digest on either side, so both sides compare `undefined === undefined`, the path is called unchanged, and it is omitted from `report.ran[].rewrote` even when the spawn pointed it somewhere else. Step 9 renders `rewrote` to tell the adopter which files now intentionally differ from the release copy, so an omitted path is one the adopter is never told about.

The mutation that was run against this line during review (replacing the `isUntouched` branch with `false`) killed tests 11 and 22, but those cover the regular-file path and would fail for any breakage of the filter at all. Nothing exercises the symlink-declared-path case the change exists for. Test 12d-i covers an **undeclared** link retarget, which exercises `sweepScope`, not this filter.

## Suggested fix

Add a case with a declared path that is a symlink: declare it in the manifest, have the regeneration program repoint it, and assert the path appears in `report.ran[0].rewrote`. Confirm it fails against the `digestOf` form and passes against `isUntouched`, so the fixture pins the behavior rather than the implementation.

A declared path being a link is unusual, which is why the field is informational-only and this is a suggestion rather than a defect. The reason to cover it anyway is that the whole point of the change was this case, and nothing currently holds it in place.

Handler: prompt

stevensacks:
`gaia update regen-regions` promises that a region's regeneration program may rewrite the paths it declares and nothing else: anything it touches beside them is put back from a pre-image, anything it creates is removed, and the runner itself never writes outside the region's own directories. Two ways that promise failed, one symlink-shaped and one presence-shaped, plus the missing test that let a fix silently regress.

## What was wrong

**`#1128`, the snapshot walked through links and the sweep wrote through them.** `readdirSync(dir, {recursive: true})` decides descent with a *following* stat, so it walks straight into a symlinked scope subdirectory. Verified on Node v24.18.0 rather than assumed: for `linkdir -> <outside>`, the walk returns `["linkdir", "plain.txt", "linkdir/precious.txt"]` and `lstat` calls that third entry a regular file. Its bytes live outside the repository; its snapshot key does not say so; and `sweepScope` resolves keys lexically, so a revert read and wrote outside the tree while `report.confined` named a path inside it.

That defect has three shapes, and only the first was named by an issue. **Second**, a program that *replaces* a scope subdirectory with a link elsewhere: `existsSync(<outside>/extra.md)` was `true` on `main` before this change, with the in-repo path left empty and the report calling it `restored`. **Third**, a link *above* the scope root, which `lstatSync(absDir)` resolves straight through because it follows every component but the last. Both passes then read through that link, which is self-consistent only while it is there; when the program replaces it with a real directory, the restore resolves the same keys lexically and materializes the adopter's out-of-tree file **inside** the repository, at a path that never held it.

**`#1130`, an unreadable path was deleted as if the program had created it.** Presence in the snapshot answers "was this here before the run"; restorability answers "can I put it back". Five IO arms conflated them by dropping the entry, so the removal loop unlinked what the adopter already had and reported it as a stray write cleaned up. The issue named the `readlink` arm; the file, entry, directory, and **scope-root** arms have the same defect and are **more** reachable, needing no mock at all: a mode-000 file the program `chmod`s came back `{"action":"removed"}` and was gone, a mode-000 directory of three files yielded three `removed` rows, and a scope directory whose own parent lost its search bit lost every file under it.

**`#1134`, the `rewrote` filter's symlink case had no test.** Reverting `isUntouched` to the old digest-only comparison left all 62 tests green while the misreport returned: a declared path that is a link has no digest on either side, so it read as unchanged however the program repointed it, and Step 9 never told the adopter about it.

## What changed

- The snapshot walks by hand from an explicit pending list and never traverses a link at any position, so a link is recorded as itself and nothing behind one is read or written. **A scope directory is asked the same lexical question every write is asked** before it is walked, so a scope reached through a link is not snapshotted rather than read one way and written another.
- **Every write asks first** whether the components between the root and the key are real directories, and reports rather than resolving through one that is not. Undoing the program's creations before restoring pre-images is kept, because it is what makes an actual restore the common outcome, but the guarantee no longer rests on it.
- **A path is read as a spawn creation only when the snapshot actually looked inside every directory above it.** The walk keys a node precisely when it does NOT enumerate it (unreadable, unstattable, a link recorded and not descended, a regular file, a FIFO or socket or device node), and it keys nothing where it either enumerated the node or established its absence: a real directory it descends, whose children speak for it, and a scope root whose ENOENT establishes the whole subtree absent. So `before.has(ancestor)` is the exact test for "this pass neither looked in here nor found it absent". Node kind is never a reason to skip recording a node that is there, however plainly childless the kind looks, because what the program puts at that path next is a different node. Anything that turns up under such a node is reported rather than removed, because it may equally be something the adopter already had that just *arrived* at that key. `mv` is the ordinary way that happens, and a move leaves no second copy.
- **That reaches every level, down to the scope root itself**, whether it was unreadable, a link, or behind one: an unrecorded scope root is an absent one. Only the scope root's `lstat` gate discriminates on errno, since ENOENT there is the ordinary "this region has no directory in this tree" case rather than a failure to look. The scope key is derived through `path.resolve` rather than from a declared path's `dirname`, so the shelter does not rest on a declaration being well shaped.
- `isUntouched` compares permission bits alongside content, since the restore writes them back. A program that only relaxes an in-scope file's mode was judged to have changed nothing.
- **`sweepScope`'s docblock names six limits**, each one the audit rounds forced into the open: three that surface as `reported`, plus three that name their own arms. A file written THROUGH an in-scope symlinked directory is not reverted and gets no row for the directory, and what reaches the file itself has three answers depending on where the link points, verified by execution: outside the root nothing reports it and nothing should; inside the root but outside every scope the whole-root `git status` half reports it, unless git is unavailable **or that path is ignored**; and pointing into another scope directory the bytes land at a key the snapshot enumerates, so the file is removed like any other undeclared creation. The fifth is a directory the program creates in scope: where the after pass can enumerate it, it is left behind empty and without a row, since the walk keys no directory it descends while what the program put inside it is keyed and removed; where the after pass cannot enumerate it, the directory takes a `reported` row of its own and its contents are never keyed, so they stay. `HELP_TEXT`'s "never silently kept" stays narrowed to a regular file or symlink for exactly that reason. The sixth is the price of the invariant covering a node with nothing to read at all: an in-scope FIFO, socket, or device node holds a `reported` row for as long as it sits there, on every run rather than once, because an `unreadable` entry is never untouched.

## Verification

Quality Gate as `wiki/decisions/Quality Gate.md` writes it, re-run on the branch tip after the last edit rather than quoted from an earlier run. Typecheck and lint clean in both workspaces at zero warnings; CLI vitest and root vitest green; Playwright green; dev smoke HTTP 200; build clean. Distribution harness against the rebuilt binaries, and CI's committed-binary freshness check replicated locally (`pnpm -C .gaia/cli bundle` then `git status --porcelain`, empty). `shell-lint` clean; the two region bats suites green under `bats5`. No bats suite drives this subcommand, verified rather than assumed. Simplify ran its four review agents; three findings applied, one skipped with the reason recorded below.

**Eighteen mechanisms, eighteen mutants, every one killed**, all proven against the final tree rather than an earlier shape:

| Mutant | Reverts | Kills |
|---|---|---|
| A | the prune, walk descends into a symlinked directory | `12h` (+ the `12d` family) |
| B | the loop order, removals last again | `12h-i` |
| C | the readlink catch, back to a bare skip | `12i` |
| D | the `rewrote` filter, back to the digest-only form | `11a` |
| E | `isUntouched`, an unreadable pre-image counts as untouched | `12i`, `12j`, `12k-i` |
| F | the write-site lexical guard, never asked | `12h-ii` |
| G | the unreadable-file arm, back to dropping the entry | `12j` |
| H | the unreadable-directory arm, back to dropping the entry | `12k` |
| I | the unenumerated-ancestor check, never asked | `12h-iv`, `12h-v`, `12h-vi`, `12k`, `12k-ii`, `12n` |
| J | the scope-directory lexical gate, never asked | `12h-iii` |
| K | permission bits not compared | `12l` |
| L | the unstattable-entry arm, back to dropping the entry | `12k-i` |
| M | the scope-root gate, back to swallowing every errno | `12k-ii` |
| N | the unwalked-ancestor scope root, recorded no longer | `12h-iii` |
| O | the linked scope root, recorded no longer | `12h-iv` |
| P | the shelter, narrowed back to unreadable ancestors only | `12h-v`, `12h-vi` |
| Q | the gate overwrites the walk's entry for the same node | `12m` |
| R | the non-file arm, back to dropping the entry | `12n`, `12n-i` |

`11a` passed the moment it was written, which is `#1134`'s own point: the behavior ships correct and nothing held it in place. Mutant D is the proof the issue asked for. Mutant A also perturbs the `12d` link-to-file cases as formulated (descending into a link to a file makes `readdir` fail), which is a property of the mutant rather than a claim about those tests.

Every permission fixture (`12j`, `12k`, `12k-i`, `12l`) is non-vacuous under `root`: each asserts the row that appears only when the read genuinely fails, so a privileged run fails loudly instead of passing on the wrong path.

## Notes for review

- **Five audit rounds; four refused, the fifth a clean pass, and the shape of what they found is the story of this PR.** Round 1 demonstrated the unreadable-file and unreadable-directory data loss against the shipped bundle with no mocks, and produced the above-scope-link fixture that overturned this branch's own reasoning that such a link was self-consistent and therefore harmless. **Round 2 caught that the repair was incomplete**: it closed every arm inside the walk and left the `lstat` gate above it. **Round 3 caught that that repair was also incomplete**: two further arms at the same gate still recorded nothing, and de-symlinking the path with a `mv` walked the adopter's only copy into the removal loop. **Round 4 caught that the shelter itself was the wrong question**: it asked whether an ancestor was *unreadable*, when the property that matters is whether the pass *enumerated* it, so a symlinked subdirectory or a plain file one level inside the scope still lost the adopter's only copy. Each round found the same defect one level further out, and round 4 replaced the predicate with a smaller, exact one instead of widening it again. **Round 5 closed the class**: it re-derived the enumeration from scratch on sixteen hermetic fixtures against the shipped bundle, found every row safe, and named the invariant that actually holds, which is stronger than non-enumeration. Its one Important is on a different axis, not which nodes get an entry but which of two entries for one node wins, and it is fixed here. **Round 6 traced that precedence rule exhaustively and cleared it** (both declaration orders against all four values the walk can put at a shared key, crossed with the three gate arms, plus three-deep nesting), and then caught something this PR had introduced in the round before: consolidating the contract into one invariant block had replaced an accurate bounded acknowledgment with a **false safety claim**, that a FIFO needing no key "is safe there and not a hole" because nothing can resolve through one. True only while the node is still a FIFO. **That is fixed in the code rather than in the prose**, which is what this round is: the walk now keys a non-file node like every other node it does not enumerate, and the caveat disappears from five prose surfaces at once instead of being restated correctly on each.
- **Round 7 cleared the new arm on the questions that mattered and found the sentence describing it was still not exact.** It confirmed nothing opens a FIFO on the snapshot, restore, or report path, that `isUntouched`/`isRestorable`/the removal loop all handle the new entry correctly, and that no consumer mishandles the new row; it independently reproduced limit 4's third arm and confirmed `HELP_TEXT` was right to stay narrow. What it caught is that the biconditional written the round before is falsified by the scope-root gate's ENOENT arm, which keys nothing and enumerates nothing. The behavior there is correct, since ENOENT establishes the whole subtree absent, which is the positive proof the invariant asks for; the sentence had no room for it. **The imperative underneath was the real hazard**: "an arm that stops keying a node it does not enumerate breaks it" reads, against ENOENT, as a description of something already broken, and a maintainer repairing the odd arm out would shelter every not-yet-existing scope root and silently turn the removal loop off for the ordinary case. The invariant is now stated three-valued, and the converse is named as the easier of the two ways to break it, because it looks like tightening.
- **The last round's fix is a deletion, and that is the point.** Keying the `!stat.isFile()` arm makes the removal invariant a biconditional rather than a proxy, so the invariant block loses its second establisher, `sweepScope` loses its FIFO limit, `hasUnenumeratedAncestor` needs no hedge, and the wiki's second and fourth limits stop needing an exception. It costs one `reported` row per FIFO, socket, or device node in scope per run, which for any real adopter is none, and it makes a program-created FIFO removable, which strengthens `HELP_TEXT`'s claim rather than weakening it. Verified through the rebuilt binary in all three shapes: the adopter's moved-in copy survives with both paths reported, a created FIFO is removed, and an untouched one costs exactly the one row.
- **Eight rounds, and the last four were clean passes whose findings were all prose. That pattern is the reason this merges here.** Round 6 raised one Important and five Suggestions, round 7 two and four, round 8 three and four, and two of round 8's three Importants were introduced by round 7's own repair. The findings were not converging because refining the invariant's statement is what generated them: each rewrite made the block richer, and a richer block makes more falsifiable claims. Round 8's remaining findings are filed as #1172 with that diagnosis attached and the instruction to delete claims rather than reword them, which is the only edit to this block that is convergent. The behavior underneath has been stable since round 4 and is pinned by 79 fixtures and 18 mutants.
- **A mutation pass proves the mechanisms you thought to name, and nothing else.** Thirteen mutants were green on the tree round 3 refused and fifteen on the tree round 4 refused, because in each case none of them reached the arm that round found. That is worth stating plainly rather than presenting seventeen-of-seventeen as coverage.
- **The contract is now stated once.** Four rounds of prose findings had one root cause: the same guarantee was restated in five places plus the wiki, each round patched a subset, and the drift always ran toward claiming more than the code delivers. The removal invariant now has a single authoritative statement on `collectScopeDigests`, and the module docblock and the ancestor test point at it. That is also where the precedence rule this round adds now lives, and the absence of such a place is plausibly why two writers to one map went unnoticed for five rounds.
- **Asked directly whether this should escalate to a design spec rather than another repair, the fifth round said no, and said why**: the class is closed, the invariant fits in a sentence with sixteen probed rows behind it, and this round's finding has a three-line fix. A spec would buy a sixth artifact restating the same contract.
- **Both the round-2 and round-3 Criticals were verified before they were applied.** Round 2's repro label named the scope root itself at mode 0600, but that case is already shielded, since `readdir` needs only the read bit and the per-entry `lstat` arm catches the rest; a probe separated the two and the fix targets the mechanism the member described rather than the label it carried (the member withdrew the label in round 3). Round 3's Critical was reproduced independently in both shapes, a linked scope root and a linked ancestor, each deleting `KEEP.md` and reporting it `removed`, before a line was changed.
- **The wiki's symlink sentence has now been rewritten three times, and the third time is the one that is true.** It first claimed writes behind a link are neither reverted nor reported; round 2 "corrected" it to say the git-status pass still reports tracked writes there; round 3 pointed out that git tracks a symlink as a link and never descends into it, so that named an empty set. It now says what the code does: nothing behind the link is reverted or seen, and the directory itself is reported.

### Findings recorded rather than fixed

- **Filed as #1170, not waived** (`regen-regions.ts`, `normalizeRepoPath`): it strips a leading `./` but leaves a trailing `/` and an interior `/./` through, so a declared path never matches its snapshot key and the sweep reverts the region's own freshly written output, reporting it as a stray write cleaned up while `ran[]` calls the region run. Round 2 raised the trailing-slash shape and it was waived into this body as out of scope. **Round 4 raised the `/./` shape of the same defect in the same function, which is that waive being observed to spread**, so it is filed rather than recorded in a body that only a merged PR carries. Still not fixed here: it is declared-path parsing rather than write confinement, and belongs to none of the three issues this PR closes. The `/./` shape's second consequence, that it also defeated this PR's own ancestor shelter, IS closed here, by deriving the scope key through `path.resolve`.
- **Declined, with rationale** (the `rewrote` filter): an unreadable *declared* path now always lands in `ran[].rewrote`, because `isUntouched` answers `false` for an unreadable pre-image. There is no honest answer available here: both `true` and `false` assert an observation that was not made, and the report has no third state. For a declared path, which the program exists to rewrite, "rewrote" is the safer default. The alternatives are worse than the defect: splitting `isUntouched`'s answer per consumer re-creates exactly the presence-versus-restorability conflation `#1130` exists to prevent, and special-casing the call site adds a second rule to a report line that gates nothing.
- **Filed as #1172, accepted-and-noted rather than folded in** (`regen-regions.ts:761`, with dependent sites at `:27` and `:768`). Round 8's three Important findings are all documentation: the invariant is stated over the KEY where it is only true of an ANCESTOR (an enumerated directory has no entry, so an empty in-scope directory the program replaces with moved-in content is removed, executed and confirmed); the arm census says two scope-root arms key when three do; and the module docblock's new snapshot sentence contradicts the invariant block about directories. **None is fixable in code**: the member probed the control case and sheltering enumerated directories regresses a reachable one, where a non-empty subdirectory replaced by a link today emits `removed` then `restored` and would become `reported` with the restore then refusing on `resolvesLexically`. Every hazard the three claims create is caught by the suite rather than only by the prose (keying descended directories fails test 13; pruning the `!resolvesLexically` arm fails `12h-iii` and mutants N and O). Taken under the accept-and-note arm of the digest-economics rule: the member is marked, nothing else rotates its digest, and folding these in buys a full re-dispatch solely to re-earn the marker. **DRAIN's spread override is why this is filed rather than waived**: this exact class, prose in this file claiming more than the mechanism delivers, has now recurred in rounds 4, 5, 6, 7 and 8, which is a waive observed to spread.
- **Waived, not filed** (`regen-regions.ts:584`, `performBackup`): `existsSync` returns true for a FIFO and `copyFileSync` then opens it `O_RDONLY`, which blocks forever with no writer. This runs *before* the spawn, so the spawn timeout does not apply, and the surrounding `try`/`catch` never fires because the call does not fail, it never returns: the command emits no report, no stderr, and never exits, stalling the whole update flow. Verified by the audit member (alive after 10s, killed with SIGKILL; the same fixture with the FIFO undeclared completes instantly). Waived rather than filed because it is **pre-existing and untouched by this PR** (nothing here changes `performBackup`), it is a non-security finding on a file this PR already changes, and the spread override does not fire: one call site, first observation, no prior waive of it. Reaching it requires an adopter to have replaced a shipped region file with a FIFO. The fix, when someone takes it, is to `lstat` the source before copying and route a non-regular declared path through the existing `region_regen_backup_failed` channel so the run continues and reports.

  
  

- **Filed as #1171** (`.claude/skills/update-gaia/SKILL.md:759`): Step 9's summary vocabulary defines a `reported` confined entry as "a write outside the region's own directories that had no pre-image to restore from". Neither term holds. The sweep reports several strictly in-scope cases (an unreadable pre-image, an unenumerated ancestor, a path whose components are not all real directories, a created directory the after pass cannot enumerate), and as of this PR it also reports an in-scope FIFO that nothing wrote to at all, on every run. The same line tells the maintainer that "a `reported` entry is the one the adopter has to look at", so the error reaches the adopter as advice. Filed rather than waived because this PR does not change that file and it is not a machinery path, so the waive rule cannot route it. `#1131` already covers the sibling confusion at line 622 of the same file and its fix does not reach line 759.
- **One claim was deliberately NOT widened.** `HELP_TEXT` says "a regular file or symlink the snapshot could examine is never silently kept". Recording FIFOs makes that claim stronger, so it stays true, but widening it to "anything the snapshot could examine" would make it false: a directory the program creates in scope is examined and IS silently kept, verified by probe. That is stated as `sweepScope`'s fifth limit instead of being smuggled into the help text.
- **One limit is a real cost, and it is stated rather than hidden.** A region whose directory is reachable only through a symlink is no longer swept at all: it reports nothing and reverts nothing. Reading through a link the writes will not resolve through is precisely what put an out-of-tree file inside the repository, so the two passes are made symmetrical instead. The alternative, recording each scope root's real path and writing back through it, preserves the sweep for that adopter but deliberately writes outside `root` and threads a new field through the snapshot; that is a design fork, not this fix.
- **CHANGELOG: one `### Fixed` entry, citing `#1128` and `#1130`.** `regen-regions` routes on the adopter binary, `.gaia/manifest.json` classes `.gaia/cli/gaia` as `owned`, and `update-gaia/SKILL.md` runs it on adopter machines against their own tree, so the product surface moves. `#1134` is deliberately not cited: it pins already-shipped behavior and has nothing for an adopter to read.
- **One Simplify finding was skipped.** A shared traversal helper with `release/scrub.ts`'s `walkFiles` was proposed. It is module-private to another domain and returns only plain-file paths where this walk must also record links; factoring it would edit a file outside this diff; and it decides descent from `Dirent.isDirectory()`, which is `false` for `UV_DIRENT_UNKNOWN` on filesystems that do not fill `d_type`, silently pruning a real subdirectory out of a confinement snapshot. `lstat` is exact and worth its syscall here.

Closes #1128
Closes #1130
Closes #1134


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.XpbdRYFAXE

Issue #1129, opened 2026-07-31
cli-tests: the distribution-harness paths filter covers none of the files the marker-strip assertions inspect


stevensacks:
**Location:** `.github/workflows/cli-tests.yml:206` (the `distribution-harness` job's `dorny/paths-filter` `code:` filter)

## Failure mode

The PR lane for the distribution scenarios filters on `.gaia/cli/**`, `.gaia/release-exclude`, `.gaia/release-scrub.yml`, `.gaia/tests/distribution/**`, and `.github/workflows/cli-tests.yml`. Every file the marker-strip scenario's assertions actually inspect sits **outside** that set: `.claude/hooks/lib/*.sh`, `.gaia/scripts/*.sh`, `.gaia/statusline/gaia-statusline.sh`, `.gaia/audit-ci.yml`, `.prettierignore`.

So a PR that adds a `# gaia:maintainer-only:` block to, say, `.gaia/scripts/summary-verify.sh` with a bare `#` separator on both sides never runs the harness. The doubled-separator guard (`03-marker-strip.sh`, assertion 4) first fires at `release.yml`'s pre-publish gate rather than on the PR that introduced the defect. `distribution.yml` is `workflow_dispatch`-only, so it is not a second chance either.

Pre-existing for assertions 1-3 as well, not new. The fourth assertion makes the gap visible because its subject files are entirely outside the filter.

Not a false-green in the release sense: `release.yml` still gates before the tarball is built, so the artifact never ships to an adopter. The cost is feedback timing. The defect is found at release rather than at review, by which point it is several PRs old and the author who introduced it has moved on.

## Suggested fix

A real tradeoff, which is why this is filed rather than folded into the PR that surfaced it:

1. **Widen the filter** to cover `#`-marker-bearing paths (`**/*.sh` and friends). Simple, but every shell-touching PR then pays a full staging build: Node + pnpm setup, exclude-regex compile, full rsync + scrub.
2. **Add a narrow job** that builds staging and runs only `03-marker-strip.sh` when a `#`-marker-bearing path changes. Cheaper per PR, but a second staging-building job to maintain.

Decide between them rather than assuming the widening. Whichever is chosen, the deciding question is what a PR that adds a maintainer-only block should pay to learn its separator placement is wrong.

Handler: prompt

stevensacks:
[...]

PR #1175, opened 2026-08-03
debt(cli): the regen-regions remainder, five issues in one file (#1125, #1126, #1170, #1172, #1174)

### LINKED ISSUE #1125: regen-regions spawns a redundant git rev-parse per status call, 4N git processes for an N-region manifest
`.gaia/cli/src/update/regen-regions.ts:860`

`gitStatusPaths` resolves the repository-to-root offset by spawning a second git process:

```ts
const prefix = execGaiaGitRaw(['rev-parse', '--show-prefix'], root).replace(/\r?\n$/u, '');
```

That call sits inside `gitStatusPaths`, which `runRegeneration` invokes twice per region (once before the spawn at line 1080, once after at line 1097). The prefix is invariant for a given `--root` for the whole life of the run: it cannot change between the before and after snapshots, and it cannot differ between regions, because every region is regenerated against the same `--root`.

**Failure mode.** A manifest declaring N regions pays 4N git spawns where 2N would do: N extra `rev-parse --show-prefix` before each region's snapshot and N more after. There is no wrong output and no incorrect classification, only repeated work. It is unmeasurable at the one region GAIA ships today and grows linearly with an adopter's own declared regions, so this is a latency cost on a many-region manifest rather than a defect an adopter can observe as wrong behavior.

**Suggested fix.** Resolve the prefix once per run and pass it in, rather than deriving it per call. Two shapes fit the existing code: compute it alongside `realRoot` where the run's inputs are loaded and carry it on `RegionContext` (which already threads `realRoot` to every region), or memoize on first use keyed by root. The first keeps the data flow explicit and matches how `realRoot` is already handled. Either way `gitStatusPaths` takes the prefix as a parameter and its `git status` call becomes the only per-region git spawn.

Behavior must not change: the existing nested-root tests (13c, 13d, 13e in `regen-regions.test.ts`) already pin the rebasing semantics and should stay green untouched.

Handler: prompt

### LINKED ISSUE #1126: regen-regions spawn bounds have no injection point, so the timeout arm of the killed-cause mapping is untestable end-to-end
`.gaia/cli/src/update/regen-regions.ts:931`

The two runaway guards on the regeneration spawn are module-level constants with no injection point:

```ts
const SPAWN_MAX_BUFFER_BYTES = 32 * 1024 * 1024;
const SPAWN_TIMEOUT_MS = 5 * 60 * 1000;
```

`trySpawn` classifies a killed child into `cause: 'timeout' | 'maxBuffer' | 'external'` by reading the `code` field off the thrown error (`ETIMEDOUT`, `ENOBUFS`, or neither). The `maxBuffer` arm is driven end-to-end by test 9c, which writes 33 MiB to stderr and asserts `cause: 'maxBuffer'`. The `timeout` arm is not: driving it for real would mean a suite that waits five minutes, so test 9d calls `spawnFailureCause('ETIMEDOUT')` directly and its own comment concedes the limitation.

**Failure mode.** The mapping from code string to cause is covered, but the wiring from a real timeout kill to that mapping is not. If a future change alters which field carries the error code (for example reading the nested `error.error.code` instead of the top-level `code`, or a Node version that populates them differently), test 9c catches the break for `maxBuffer` and nothing catches it for `timeout`. The reported cause silently degrades to `external`, and an adopter whose regeneration command legitimately runs past five minutes is told the kill came from outside GAIA rather than from the ceiling GAIA imposed, which is the exact misreport the `cause` field was added to prevent.

**Suggested fix.** Thread both bounds through the existing `RunOptions` (which already carries `cwd`), defaulting to the current values so no caller changes and no shipped behavior moves. Then add a test alongside 9c that passes a small timeout against a `sleep` script and asserts `cause: 'timeout'`, so both arms are proven on the same path. The mapping unit test can stay as a cheap guard on the string values.

Handler: prompt

### LINKED ISSUE #1170: regen-regions: normalizeRepoPath leaves a trailing slash and an interior /./ through, so a declared path never matches its snapshot key and the sweep reverts the region's own output
`.gaia/cli/src/update/regen-regions.ts:300`, `normalizeRepoPath`.

## Failure mode

`normalizeRepoPath` converts `\` to `/`, collapses repeated separators, and strips a **leading** `./`. It leaves two shapes through:

- a **trailing** `/` — a declared `a/b.md/` stays `a/b.md/`
- an **interior** `/./` — a declared `a/./b.md` stays `a/./b.md`

Neither is refused by `normalizeDeclaredPaths`' four guards (empty, absolute, `..` segment, parent-is-repo-root), so both reach `declaredSet` and the scope-dir computation verbatim.

The snapshot canonicalizes its own keys through `path.relative`, so neither shape can ever equal its snapshot key. `declaredSet.has(relPath)` is therefore false for the region's own declared path, and the sweep treats the file the regeneration command just wrote as an undeclared write and **reverts it to its pre-image**, while `ran[]` reports the region as run and `confined[]` reports a `restored` row that reads as the runner cleaning up a stray write. The region is silently left un-regenerated.

This is precisely the property `normalizeDeclaredPaths`' own docblock (lines 314-318) says holds "by construction": *"a declared `./a/b.md` would otherwise never match its own snapshot entry, and the sweep would revert the very file the regeneration just wrote while reporting the run a success."* The leading-`./` case is handled; these two are the same defect in the same function.

Both shapes were reproduced with runnable probes by the `code-audit-maintainer-node` audit member, in two separate rounds on PR #1169: the trailing slash in round 2, the interior `/./` in round 4. The `/./` probe returned `ran.rewrote: [".claude/agents/two.md"]` alongside `confined: [{"action":"restored","path":".claude/agents/one.md"}]`, with `one.md` back to its pre-run content on disk.

The `/./` shape had a second consequence, on the confinement shelter: `path.posix.dirname('.claude/agents/./one.md')` is `.claude/agents/.`, a scope key `relativeKey` can never produce, so the ancestor shelter silently failed to match. **That consequence is already closed** on PR #1169, which derives the scope key through `path.resolve` rather than from the declared path's `dirname`. The declared-path defect itself is untouched and is what this issue covers.

## Trigger and reachability

`.gaia/manifest.json` is release-generated, so the trigger is a hand-authored or hand-edited declaration rather than adopter input. Low reachability, but the outcome is silent: nothing is refused, nothing warns, and the report reads as a successful run.

## Suggested fix

Normalize the two shapes in `normalizeRepoPath` (strip a trailing `/`, drop interior `.` segments), or route through `path.posix.normalize`. If the latter, keep the existing `..` guard on the **pre-normalized** value, since `path.posix.normalize` resolves `..` away and would otherwise disarm the guard that refuses a parent segment. The existing empty-entry guard then refuses a bare `/`.

Add a declaration-shape test per shape, asserting that each either canonicalizes to the snapshot's key or is refused outright, and check the sibling consumer `isInsideScope`, which compares raw scope-dir strings against snapshot-framed keys and has the same sensitivity.

Handler: prompt

### LINKED ISSUE #1172: regen-regions' removal-invariant block makes three claims the mechanism does not deliver, including one over the key that an enumerated directory falsifies
`.gaia/cli/src/update/regen-regions.ts:761` (the removal-invariant block on `collectScopeDigests`), with two dependent sites at `:27` and `:768`.

Three claims in the module's designated single statement of its write-confinement invariant say more than the mechanism delivers. All three were confirmed by execution against the built binary during PR #1169's eighth audit round; none is a behavior defect, and the correct disposition for every one of them is prose.

**Read the "how to fix this" section before editing. The instruction there is not stylistic: refining these sentences is what produced them.**

## Failure mode

**1. `:761` states the invariant over the KEY, where it is only true of an ANCESTOR.** The sentence permits deletion "only where this pass POSITIVELY ESTABLISHED that nothing was at that key before". The removal loop's key test is `before.has(relPath)`, and the one node kind the walk enumerates *without* keying is a real directory. So a key the before pass observed as an empty directory is absent from `before` and is deleted.

Executed shape: `docs/region/sub` is an empty real directory in scope; the regeneration command runs `rmdir docs/region/sub` then `mv <outside>/only-copy.txt docs/region/sub`. The report is `{"action":"removed","path":"docs/region/sub"}`, the key is gone, and because `mv` moves rather than copies, the adopter's only copy is gone with it. The same holds at the scope root. That is the exact outcome the paragraph's own closing clause ("deleting it destroys the only copy") disclaims.

Note the asymmetry that is already correct in the same block: "no entry for this **ancestor** means this pass enumerated it or observed it absent" is true, and is what `hasUnenumeratedAncestor` reads. Only the statement quantified over the key is wrong.

**2. `:768` undercounts the scope-root gate: three arms key, not two.** It says "the two scope-root gate arms that find something they may not walk all get a key". Three arms call `recordUnexaminable()`: the `!resolvesLexically` arm, the non-ENOENT `lstat` arm, and the `!stat.isDirectory()` arm. All three are individually mutant-proven (mutants N, M, O). The descriptor "find something they may not walk" fits only the third: the first keys the scope root having never looked at it, because what it found was a non-directory *above* it, and the second keys on an errno meaning "cannot tell" rather than "found".

**3. `:27` contradicts the invariant block about directories, in the disabling direction.** The module docblock says the pass "hashes every regular file and reads every symlink ... and records the presence of everything else it finds there without looking inside it". An enumerable directory is "everything else it finds there", and the pass does exactly the two things that clause denies: it looks inside it and it records no key for it.

## Why each of these has a bounded blast radius

Worth stating so this is triaged as documentation rather than as a latent data-loss bug. Every hazard the three claims create is caught by the existing suite, not only by the prose. A maintainer who "repairs" the docblock by keying each descended directory makes `hasUnenumeratedAncestor` true for every key in scope, and test 13 fails. A maintainer who prunes the `!resolvesLexically` scope-root arm as over-keying fails fixture `12h-iii` and mutants N and O.

The data-loss shape in 1 is real but essentially unreachable and **must not be fixed in code**. Sheltering enumerated directories regresses a reachable case: for a non-empty subdirectory replaced by a symlink, the sweep today emits `removed` for the link then `restored` for the file beneath it; shelter the directory and the link becomes `reported`, the restore then fails `resolvesLexically`, and the file is never written back. The limit is correct; only its description is wrong.

## How to fix this, and why the obvious approach is the wrong one

**Delete claims. Do not refine them.** This file's prose was rewritten across five consecutive audit rounds, and the pattern is legible in the round counts: round 6 raised one Important and five Suggestions, round 7 raised two and four, round 8 raised three and four. It never converged, and the reason is that each rewrite made the block richer and therefore made more falsifiable claims, so two of round 8's three Importants were introduced by round 7's own repair.

The three surfaces above are all claims the block does not need to make: a census of which arms key (2), a universally quantified statement over the key rather than the ancestor (1), and a second summary of node handling in a docblock that already defers to the invariant block for the invariant itself (3). Removing them is strictly convergent; rewording them is what generated this issue.

If a positive statement is wanted for 1 rather than a deletion, the accurate one is that deletion requires the pass to have established that nothing **restorable** was at that key: either it found nothing there, or it found a directory it enumerated, whose pre-image is its children and is written back beneath the key once the node now standing there is gone.

## Related

Surfaced by PR #1169 (`#1128` + `#1130` + `#1134`), which rewrote this block. That PR merged with these accepted-and-noted under the PR Merge Workflow's accept-and-note arm, because the member was already marked and nothing else was rotating its digest. The PR body carries the full finding list.

Handler: prompt

### LINKED ISSUE #1174: regen-regions: performBackup opens a declared FIFO and blocks forever, stalling /update-gaia with no report
`.gaia/cli/src/update/regen-regions.ts:584`, in `performBackup`.

## Failure mode

`performBackup` copies each declared path aside before a region is regenerated. It guards
with `existsSync` (`:574`) and then copies (`:586`):

```ts
if (!existsSync(srcAbs)) return;
...
copyFileSync(srcAbs, destinationAbs);
```

`existsSync` returns `true` for a FIFO, and it never opens the node, so it returns
immediately. `copyFileSync` must then open the source `O_RDONLY`, and opening a FIFO for
reading **blocks until a writer opens the other end**. With no writer, it blocks forever.

Three things make this worse than an ordinary stall:

- **The surrounding `try`/`catch` (`:584-595`) never fires.** It catches errors; nothing
  has failed. The call does not throw, it never returns.
- **It runs before the spawn**, so the subcommand's spawn timeout does not cover it.
- **The command emits no report, no stderr, and never exits.** The adopter sees
  `/update-gaia` stop dead partway through with no message and no path named, on a tree
  the merge walk has already partially written.

Verified by PR #1169's audit member rather than reasoned about: the process was still
alive after 10s and had to be `SIGKILL`ed, and the same fixture with the FIFO undeclared
completes instantly.

**This ships.** `regen-regions` is in the adopter CLI (`.gaia/cli/src/index.ts:46`), and
`.claude/skills/update-gaia/SKILL.md` Step 7d runs it on adopter machines against their
own tree with `--root .`.

## Trigger

An adopter has a named pipe at a path a region declares, i.e. they replaced a shipped
region file with a FIFO. Not reachable by accident; it takes deliberate action. Severity
is `suggestion` for that reason, and the cost when it does happen is the reason it is
filed rather than left as a body waive: a hang names nothing, so the symptom gives a
diagnosing adopter no route to the cause.

## Suggested fix

`lstat` the source before copying and route any non-regular declared path through the
existing `region_regen_backup_failed` structured-error channel (`:589-594`), so the run
continues and reports instead of stalling.

This is a correctness improvement independent of the hang: "copy this aside so it can be
restored" has no meaning for a FIFO, a socket, or a device node, and `copyFileSync`
against a device node would read bytes from the device.

## Provenance

Recorded as a waive in PR #1169's body (`### Findings recorded rather than fixed`) under
the cross-remit touched-file rule, which was correctly applied: the finding is
pre-existing, nothing in that PR touches `performBackup`, and the spread override did not
fire (one call site, first observation). Filed now on the maintainer's decision, because a
waive records rather than tracks and its only durable surface is that merged PR's body.
The dedup key above is reused byte-for-byte from the waive entry.

Handler: prompt

stevensacks:
Tech-debt drain step S7: the `regen-regions.ts` remainder. Five issues, one file, one branch, one review unit.

Closes #1125
Closes #1126
Closes #1170
Closes #1172
Closes #1174

## What changed

**#1170 (important), a declared path never matched its own snapshot key.** `normalizeRepoPath` stripped only a leading `./`, so a declared path carrying a trailing `/` or an interior `/./` reached `declaredSet` verbatim. The snapshot canonicalizes its own keys through `path.relative`, so neither shape could ever equal its key: the sweep read the region's freshly written output as an undeclared write and reverted it to its pre-image, while `ran[]` still reported the region as run. Normalization now drops `.` segments at any position and a trailing separator, which is the property `normalizeDeclaredPaths`' docblock already claimed held by construction.

Hand-rolled rather than `path.posix.normalize`, which resolves `..` away and would disarm the parent-segment guard that reads its output. Verified that no shape which should be refused escapes: a leading empty segment survives, so the absolute-path guard still fires on `/a/`.

**#1174, `performBackup` opened a declared FIFO and blocked forever.** `existsSync` answers "is something here" without opening the node; `copyFileSync` then has to open it, and opening a FIFO for reading blocks until a writer arrives. The surrounding `try`/`catch` never fired because nothing failed, and it runs before the spawn so no timeout covered it: `/update-gaia` Step 7d stopped dead with no report, no stderr, and no exit, on a partially written tree. The source is now `stat`ed once, **following the link**, so the guard asks the one question a backup has to answer: will the copy read bytes from a regular file. A FIFO, socket, or device node is reported through the existing `region_regen_backup_failed` channel, whether the declared path names one directly or reaches one through a link.

**#1125, 4N git spawns for an N-region manifest.** `gitStatusPaths` derived `--show-prefix` itself and `runRegeneration` calls it twice per region, for a value that is a property of `--root` alone. Resolved once and carried on `RegionContext` where `realRoot` already rides.

**#1126, the timeout arm of the killed-cause mapping had no end-to-end coverage.** Both spawn bounds now ride on `RegionContext`, resolved in `run()` from an optional `RunOptions` override that defaults to the shipped value. New test 9e drives a real timeout kill and asserts `cause: 'timeout'`, making 9c and 9e twins on the two arms.

**#1172, three claims the removal invariant does not deliver.** Deleted rather than reworded, per the issue's convergent instruction: rewriting this block across five audit rounds is what produced the issue. Prose only, deliberately, because sheltering enumerated directories regresses a reachable case. A fourth site was found and deleted: the module docblock restated the key-quantified invariant in the very sentence declaring it "stated in full there and not restated here".

## Judgment calls worth reviewing

1. **#1174's scope, settled the hard way across three audit rounds.** I originally refused *every* non-regular kind, symlinks included, which is wider than the issue asked for. That generalization was wrong, and each round found the next defect in it rather than in the issue's actual subject: round 1 showed refusing a declared link left the region's own write following it to overwrite an out-of-tree file with nothing held anywhere; round 2 showed the link-as-link backup I replaced it with dangled and could be written through; round 3 showed the compensating `region_regen_declared_symlink` signal was stderr-only against a caller that discards stderr, gated on a backup having been written rather than on the hazard, and carrying a claim false in two of three target shapes.

   **The special case is now retracted entirely rather than repaired a fourth time.** Statting *through* the link collapses the whole thing into `!stat.isFile()`: a symlink to a regular file is copied exactly as the pre-branch base copied it, and a symlink to a FIFO is refused, which the base opened and hung on. Round 4 confirmed against `main`'s shipped binary that base parity holds byte-for-byte on the regular-file case, that `main` **hung** on a symlink-to-FIFO where this branch refuses cleanly, and that HEAD never loses a backup `main` would have made. Round 4 also noted the deleted signal could never have reached an adopter through the documented invocation, so removing it cost nothing that was reaching anyone.

2. **#1170 moves one refusal reason.** A declared `.` (also `./`, `././`) now normalizes to the empty string and is refused as `paths carries an empty entry` rather than `paths carries a path whose parent is the repository root: .`. Both refuse and the protection is identical. A bare `/` keeps its own "absolute path" refusal, which round 2 restored after the audit pointed out it had been collapsed into the empty-entry reason too. `normalizeRepoPath` also has a second consumer the issue did not name, `checkOperand`, which compares its output against the manifest's shipped keys; an operand written `.gaia/scripts/./x.sh` now matches its shipped key where it previously did not. The raw-operand `..` guard reads the unnormalized value and is untouched, and the realpath containment check is what actually holds containment.

3. **#1126 threads only the time bound, not both.** The issue asked for both. The output bound needs no seam: a program can breach the shipped 32 MiB inside a suite and test 9c does exactly that, so a `spawnMaxBufferBytes` override would have been dead surface. Only the five-minute bound cannot be reached in a suite.

## Verification

- 92 tests in `regen-regions.test.ts` (79 at the branch point), full CLI suite 130 files / 1912 passed / 1 skipped.
- **Mutants, each killing only its own arm**, each restored with a verified-clean tree. P (trailing-slash strip reverted) kills 1e + 1g. Q (dot-segment filter reverted) kills 1e + 1h + 1i, with 1d still green, so the new filter subsumes the old leading-`./` strip rather than replacing it. R (pre-fix `performBackup`) **hangs rather than fails**, because `copyFileSync` blocks the worker in a syscall vitest's own timeout cannot reach. S (null-prefix arm returns an empty status) kills 13b. T (timeout reads the module constant) kills 9e. W (destination probe back to `existsSync`) kills 14d, and the failure shows it writing *through* the dangling link. Y (drop the `override > 0` conjunct) kills 9f. U (source probe back to `lstat`) kills only 14c.
- **The mutant that vindicates the retraction:** restoring the exact pre-retraction shape (`lstat` plus `!isFile() && !isSymbolicLink()`) leaves 14b passing in 29ms and makes **14e hang indefinitely**. The code that had already cleared three audit rounds still stalled `/update-gaia` on a symlink to a FIFO, which is #1174's own filed defect reachable through one indirection. Every round until then had been spent on the signal instead.
- Quality Gate green end to end, re-run whole after the retraction: simplify (four review agents; reuse, simplification and altitude clean, efficiency's one finding applied), typecheck root + CLI, lint root + `lint:cli`, CLI and root suites, Playwright 8 passed / 1 skipped, dev smoke HTTP 200, `pnpm build`.
- `region-regen-convergence.bats` + `region-marker-conformance.bats` under bash 5: 11/11. `shell-lint.sh` passed.
- **The rebundled adopter binary was smoke-tested directly**, which is the check no suite performs. One fixture declaring a FIFO at `.claude/agents/./one.md` and `.claude/agents/two.md/` exercised three members at once: exit 0 immediately, `not a regular file` on stderr, `confined` empty, and `two.md` holding the regenerated content, so neither malformed shape was reverted.

Distribution scenario 17 was not run: it exercises `gaia update merge-region`, a different command this diff does not touch. No distribution scenario covers `regen-regions` at all, which is #1129's premise rather than a gap introduced here.

## Audit

Five rounds, all by `code-audit-maintainer-node`, the only member the roster dispatches for this diff. Verified non-empty and plausible at each wave, and this is the main checkout rather than a worktree, so #1055's fail-open precondition does not hold. Every round was dispatched synchronously with no `model` passed, so the member's own pinned `opus` stands.

**Round 1** — clean pass, marker earned: 0 Critical, 1 Important, 4 Suggestions, 1 cross-remit. It verified rather than asserted on all three flagged calls, fuzzing **16,916 declared-path shapes** against the four guards with zero escapes, every accepted candidate equal to its own canonical relative key, and a base-vs-HEAD verdict delta of 877 cases that are all refusal-to-refusal or accept-to-refusal. It also confirmed 9c genuinely breaches the shipped 32 MiB, which is what makes #1126's one-bound decision correct. All six dispositions applied in `df03b4bb` and `c4d1465f`: the symlink Important (since superseded, see judgment call 1); a scope-key comment rationale made unreachable by this branch's own #1170, deleted; an invariant clause vacuous for an *empty* enumerated directory, bounded; `paths: ['/']` refusing as "empty entry" rather than "absolute path", restored with new test 1j; and `spawnTimeoutMs: 0` reaching Node as `timeout: 0`, meaning *no* timeout, now honoured only when positive.

**Round 2** — cleared: 0 Critical, 2 Important, 2 Suggestions. Both Importants were introduced by round 1's own repair, and both were verified by running the shipped bundle against hermetic fixtures. The destination guard still followed links, so a backed-up link with a relative target dangled and `existsSync` called a backup sitting right there absent; and backing a link up as a link removed the only signal that shape produced. Fixed in `0de1e24e` and `a63d2d9e`. The round also caught a test of mine that passed under both the guarded and unguarded resolver, replaced with a direct unit test on an exported `resolveTimeoutMs`.

**Round 3** — cleared: 0 Critical, 3 Important, 1 Suggestion. All three sat on the signal round 2 had added. This is the round whose findings triggered the retraction rather than a fourth repair; see judgment call 1.

**Round 4** — clean pass: 0 Critical, 0 Important, 3 Suggestions. It answered every question by execution in an isolated `git archive` copy: confirmed the retraction left nothing orphaned and rebuilt both bundles to compare them byte-for-byte against the committed artifacts; A/B'd the pre-increment bundle against HEAD's over **nine occupied-destination shapes**, all identical on exit code, `backedUp[]`, error codes and destination bytes, confirming the destination-first reorder is behavior-identical; and established base parity against `main`'s shipped binary as described above. Its three Suggestions were applied rather than noted, because two of them were claims stated at higher certainty than the evidence supports, which is **#1172's own defect class** and #1172 is a member of this PR: merging them would have closed that issue while shipping fresh instances of it.

**Round 5** — clean pass: **0 Critical, 0 Important, 0 Suggestions**, empty findings sidecar. It confirmed the trimmed comments still make the `lstat`-not-`existsSync` decision legible to a future simplifier, proved 14c's new assertion is uniquely load-bearing by finding a mutation only that assertion catches, and independently rebuilt both bundles to verify they are byte-identical.

**Cross-remit, `.claude/skills/update-gaia/SKILL.md`** (raised in rounds 1 and 4): its Step 7d backup guarantee is falsified for a FIFO/socket/device, and the Step 7d invocation discards stderr so `region_regen_backup_failed` never reaches the adopter at all. **Not waivable** (a skill doc is neither gate machinery nor a file this PR changes) and **not separately fileable**: the filing recipe dedups on path+line alone and #1131 already holds that exact path and line. Recorded as a comment on #1131, which is already routed to drain step S9, the step that exists to reconcile this prose against S6/S7's shipped behavior.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.N5AKvNFHn5

Issue #1176, opened 2026-08-03
gaia init's CLAUDE.md rename silently no-ops since the H1 was removed, so every adopter scaffold gets a title-less CLAUDE.md


stevensacks:
`.gaia/cli/src/init/rename.ts:146`

`gaia init`'s rename step is supposed to rewrite `CLAUDE.md`'s first H1 to the new project title. It does that with

```ts
const next = original.replace(/^#\s.*$/mu, `# ${title}`);

if (next !== original) {
  atomicWriteFileSync(target, next);
}
```

`String.prototype.replace` with a non-matching regex returns the original string, so when `CLAUDE.md` carries **no** `# ` line at all, `next === original`, the write is skipped, and the function returns normally. There is no warning and no non-zero exit: the step reports success having done nothing.

That precondition is no longer met. `CLAUDE.md` opened with `# GAIA React` until #1173 restructured it; that commit removed the H1, and the file now opens at `## Response style`. The CLI's own docblock at `.gaia/cli/src/init/rename.ts:8` still states the assumption it depends on (``CLAUDE.md "# GAIA React" heading → "# <Title>"``), so the template and the CLI now disagree with nothing enforcing the agreement.

**Failure mode.** Every adopter who scaffolds from the template and runs `gaia init` gets a `CLAUDE.md` with no project title and no H1, while `gaia init` reports the rename step as done. The adopter's own instruction file silently keeps GAIA's generic heading structure instead of naming their project.

This is already failing in CI: distribution scenario `08-gaia-init-cli-sequence.sh:160` asserts the rewritten H1 and fails with `rename did not rewrite CLAUDE.md first H1 to '# Test Project' (got: '')`. It went unnoticed because `cli-tests.yml` is path-filtered to `.gaia/cli/**` and friends, and #1173 touched only `CLAUDE.md` and `wiki/`, so the job skipped on that merge and first turned red on the next unrelated PR that touched the CLI. **The harness job is not in the repository's required-check set, so it does not block merges** — it went red without stopping anything.

**Suggested fix.** Two parts, and the second is the one that keeps this from recurring:

1. Restore a top-level H1 to `CLAUDE.md` (`# GAIA React`), which is what the rename step and the scenario both expect.
[...]

PR #1177, opened 2026-08-03
debt(cli): parse git status through one shared -z reader (#1115)

### LINKED ISSUE #1115: inspectWorkingTree keeps the porcelain-quoting bug regen-regions just fixed, and misclassifies a wiki rename
**Location:** `.gaia/cli/src/wiki/util/branch.ts:121` (`inspectWorkingTree`)

## Failure mode

`inspectWorkingTree` parses `git status --porcelain=v1` with the same hand-rolled quote-stripping and ` -> ` rename split that #1032 just removed from `.gaia/cli/src/update/regen-regions.ts`. The two blocks were byte-identical; only the `regen-regions.ts` copy was fixed, so the defect now survives here alone.

With `core.quotePath` on (git's default), a rename of a path carrying a non-ASCII byte arrives as `"old" -> "new"`. Stripping only the outer quotes yields `old" -> "new`, and the subsequent ` -> ` split produces `old"` and `"new`, each with a stray quote and with the C-style escapes still unexpanded.

The consequence here is worse than it was in `regen-regions.ts`, because this function's output is not merely reported, it is classified:

```ts
for (const candidate of paths) {
  if (candidate.startsWith('wiki/')) { hasWikiChanges = true; } else { hasNonWikiChanges = true; }
}
```

A rename of a non-ASCII path under `wiki/` yields one half starting with `wiki/` and one half starting with `"`, so a change that is purely a wiki rename sets **`hasNonWikiChanges = true`**. Both live callers gate on that flag (`.gaia/cli/src/wiki/sync-land.ts:349`, `.gaia/cli/src/wiki/chain.ts:271` and `:332`), so a pure-wiki rename is misread as a mixed working tree and takes the wrong branch.

Latent today because it needs a `wiki/` path containing a non-ASCII byte (or a quote or a control character) *and* a rename. A plain space does not trigger git's quoting, so the many wiki pages with spaces in their names are unaffected.

## Suggested fix

Apply the same fix `regen-regions.ts` now carries: pass `-z` and parse NUL-delimited records. Under `-z` git never quotes, and a rename emits its origin path as its own trailing record, so both the unquoting and the ` -> ` split disappear rather than needing to be hardened.

`inspectWorkingTree` already separates exec from parse via its injectable `CommandRunner`, so the record→paths step extracts cleanly. Extracting it to a shared helper that both this and `regen-regions.ts`'s `gitStatusPaths` call would stop the two copies from drifting again, which is what let this one survive its sibling's fix.

Watch the one trap: `records[index - 1]` is R/C → take whole is **not** equivalent to a cleared-flag lookahead, because it misclassifies an origin path whose own first two characters contain `R` or `C` (e.g. `Config/old`). Clear the flag before classifying the next record.

Handler: prompt

stevensacks:
Closes #1115

## What

`inspectWorkingTree` (`.gaia/cli/src/wiki/util/branch.ts`) carried the same hand-rolled porcelain parse that `regen-regions.ts` had already replaced: strip the outer quotes, split on ` -> `. With `core.quotePath` on, git reports a rename of a path carrying a non-ASCII byte as `"old" -> "new"`, so that parse yields `old"` and `"new`. The second half does not start with `wiki/`, and because this function's output is *classified* rather than merely reported, a working tree holding nothing but a wiki rename came back with `hasNonWikiChanges = true`. Both callers gate on that flag (`wiki/sync-land.ts:349`, `wiki/chain.ts:271` and `:332`), so `gaia wiki sync land` refused to land and `wiki chain commit` refused to commit.

The record parser moves to `.gaia/cli/src/util/git-status.ts` as `porcelainZPaths`, and both call sites use it. `inspectWorkingTree` now passes `-z`.

## Why the shared helper is parse-only

Each caller keeps its own exec, deliberately. `regen-regions.ts` goes through `execGaiaGitRaw`'s hardened env with `--porcelain -z`; `inspectWorkingTree` goes through its injectable `CommandRunner` with `--porcelain=v1 -z -uall`, and needs `-uall` so an untracked wiki page is not collapsed to a directory name. The record-to-paths step is the whole of what they shared, and the whole of what drifted.

## Verification

- The defect confirmed live rather than assumed: a temp repo, a committed `wiki/concepts/Café Notes.md`, and a `git mv`. Pre-fix `inspectWorkingTree` returns `hasNonWikiChanges: true` with paths `["wiki/concepts/Caf\303\251 Notes.md\"", "\"wiki/concepts/Caf\303\251 Notes Renamed.md"]`; post-fix it returns `hasNonWikiChanges: false` with both paths intact.
- All 8 behavior tests in `branch.test.ts` fail against the pre-fix implementation; one of them pins the `-z` flag on the argv, because dropping it silently reinstates the defect.
- Quality Gate green: typecheck and lint clean in both workspaces, CLI 131 files / 1919 tests, root 29 / 155, Playwright 8, dev smoke HTTP 200, build clean.
- `shell-lint` clean. Both binaries rebundled. Distribution harness 16 PASS / 1 FAIL, the FAIL being the standing `08-gaia-init-cli-sequence.sh` failure tracked as #1176, unrelated to this diff.

## Note on the fixture churn

`chain.test.ts` and `sync-land.test.ts` change only where they stub `git status`: the argv gains `-z` and the canned stdout switches from `\n` to `\0`. No assertion changes.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.a2Y8Ojtu9x

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.QKfgszb0XB

PR #1178, opened 2026-08-03
debt(update-gaia): the Step 7d / Step 9 reporting contract, four issues in one file (#1117, #1131, #1132, #1171)

### LINKED ISSUE #1117: update-gaia's Step 9 summary template has no row for merge[], so cleanly-updated shared/wiki-owned files are counted and never printed
Surfaced by the pre-merge prose audit on #1108 as an out-of-scope finding. The referring line is new in that PR; the defect body is pre-existing, so it was not folded in.

## The gap

`.claude/skills/update-gaia/SKILL.md` Step 7 sorts a clean overwrite into one of two buckets by class (`:351`): `owned` → `overwrite[]`, `shared` / `wiki-owned` → `merge[]`. `:416` confirms `merge[]` is a report-counts-only bucket, the same handling `overwrite[]` gets.

Step 9's literal summary template (`:662`-`:668`) prints `Overwritten:`, `Added:`, `Removed:`, `Skipped:`, `Conflicts:`, `Deleted:`. There is no `Merged:` row.

`git grep -nIE 'Merged|merge\[\]' .claude/skills/update-gaia/SKILL.md` returns hits at 13, 353, 385, 416, 739, none of them a printed row.

## Failure mode

An agent printing the Step 9 template as written drops the count for every cleanly-updated `shared` and `wiki-owned` file. On a wiki-heavy release that is the largest bucket of the run, so the adopter's summary reports fewer touched files than the walk actually wrote, with no indication anything is missing. The `Skipped:` and `Overwritten:` rows they do see are both truthful, which is what makes the omission hard to notice.

## The ambiguity, and why this is filed rather than fixed

Two readings, and the source does not settle which is intended:

1. `merge[]` is meant to print its own row and the template omits it. Fix: add `Merged: <n>` after `Overwritten: <n>`.
2. Merges are deliberately folded into the `Overwritten:` count, and the two internal list names are an implementation detail. Fix: say so at `:416` so the folding is explicit.

Guessing wrong ships a wrong summary contract, so this needs a maintainer's read of the intent rather than a mechanical edit.

## Suggested fix

Resolve which reading is intended, then either add the `Merged: <n>` row to the Step 9 template or state the fold at `:416`. Either way `:13`'s "which bucket a clean overwrite reports under" then resolves somewhere observable.

Handler: prompt

### LINKED ISSUE #1131: update-gaia Step 7d names absent-pre-image as the revert-vs-report criterion, contradicting the in-scope creation rule below it

### LINKED ISSUE #1132: update-gaia Step 9 points at paragraphs for failed-kind remedies that state no remedies, inviting improvised adopter advice
`.claude/skills/update-gaia/SKILL.md:745` (Step 9, "Failed or refused regenerations")

The cross-reference promises more than the paragraphs it points at deliver, in a step where the gap invites an agent to invent adopter-facing advice.

## Failure mode

Step 9 item 2 says:

> Step 7d's `failed[].kind` paragraph and its `cause` paragraph are the authority on both fields' **values and remedies**.

Those two targets carry values and *meanings*, not remedies:

- Line 615 says "three values, each with its own remedy" and then glosses what each value **means**: `exit` is a program that ran and refused, `spawn` is an interpreter that never launched, `killed` is a program cut short by a signal.
- Line 617 says "the remedies differ" and then names the **ceilings** (5 minutes, 32 MiB) and which causes are GAIA's doing.

Neither states a remedy for any value. So an agent instructed to report each kind with its remedy makes the hop, finds no remedy at either target, and has to improvise adopter-facing advice, in a step whose very next paragraph forbids fabricating a command for a refusal. Inventing a remedy is the same failure as inventing a command.

The wording this replaced ("7d's shape block above is the authority on the set and glosses each kind") described the target accurately. The current wording was written to fix a different defect, a Step 9 enumeration that omitted `cause` entirely, and over-corrected.

## Suggested fix

Either narrow the promise to what the targets carry ("...are the authority on both fields' values and what each one means"), or make the remedies real by stating them where the values are glossed: `exit` means read the program's own stderr, `spawn` means the interpreter is missing or not executable, `killed` with `cause: timeout`/`maxBuffer` means the command exceeded a ceiling this flow imposed, `killed` with `cause: external` means something outside the update stopped it. The second is more useful and is what item 2's "each has a different remedy" already implies is available; the first is the honest minimum.

Handler: prompt

### LINKED ISSUE #1171: update-gaia Step 9 defines a reported confined entry as a write outside the region's directories, which it need not be, and need not be a write
`.claude/skills/update-gaia/SKILL.md:759` (Step 9 summary rendering, item 10 "Confined writes")

The line tells the maintainer what to write in the adopter-facing summary, and its definition of a `reported` entry is wrong on both of its terms.

## Failure mode

Line 759 gives the vocabulary as:

> `restored` and `removed` are writes the regeneration made outside its declared paths and the runner undid, `reported` is a write outside the region's own directories that had no pre-image to restore from and was left in place

Neither half of the `reported` clause holds.

**It is not keyed on being outside the region's own directories.** `regen-regions.ts`'s `sweepScope` emits `reported` for several strictly in-scope cases: a path whose pre-image could not be read at all (an `unreadable` entry), a path sheltered because the before pass never enumerated its ancestor, a path whose components between the root and the key are not all real directories so the restore refuses rather than resolving through them, and a directory the after pass cannot enumerate. Every one of those sits inside a region's own directory and reports.

**It is not always a write.** An in-scope node with nothing to read (a FIFO, socket, or device node) is recorded present with no pre-image, and an `unreadable` entry is never "untouched", so it takes a `reported` row on every run even when the regeneration command did not touch it and nothing about the tree changed.

The consequence is in the adopter's hands, not just the prose: the same line goes on to say "a `reported` entry is the one the adopter has to look at". An adopter told that a `reported` row means a write outside the region's directories that could not be reverted will read an untouched FIFO, or a file the runner deliberately declined to restore, as a stray write left behind by the regeneration. A maintainer rendering the summary from this text tells them exactly that.

## Suggested fix

State what `reported` actually means and make location a separate axis rather than the criterion: a `reported` entry is one the runner surfaced and deliberately left alone, either because there was no pre-image to put back or because it could not establish that writing the path was safe. Then say separately that such an entry may be inside or outside the region's own directories, and that an entry outside them was never a candidate for reverting in the first place.

## Related

`#1131` covers the same underlying confusion at line 622 of this file, where the revert-versus-report rule names absence of a pre-image as its discriminator instead of scope. That issue's suggested fix rewrites 622 alone and leaves this line saying what it says now, so draining one does not drain the other. Fixing them together is the cheaper unit, and they share a vocabulary.

Handler: prompt

stevensacks:
Drain step S9. Four `tech-debt` issues, all prose defects in `.claude/skills/update-gaia/SKILL.md`'s Step 7d / Step 9 region. No code change, no test change.

Sequenced behind S6 and S7 deliberately: this prose *describes* the `regen-regions.ts` behavior those steps changed, so every claim here is written against shipped behavior rather than behavior in flight.

## What changed

**`#1131` + `#1171`, authored as one vocabulary.** `#1171` says outright that these are the same confusion at two sites and that fixing 622 alone leaves 759 saying what it says now.

Step 7d gave revert-versus-report as "a write inside the region's own directories is reverted, and a write anywhere else, **which has no pre-image to restore from**, is reported", two clauses above a paragraph attributing the opposite outcome to the same stated cause: an in-scope creation, "**having no pre-image**, is removed". Both statements are individually correct about outcomes and they cannot both follow from the stated reason. **Scope** is the real discriminator; pre-image presence discriminates restore-from-remove *within* scope.

Step 9 item 10 defined a `reported` entry as "a write outside the region's own directories that had no pre-image to restore from". Neither term holds, and the same line tells the adopter this is the row they have to look at:

- `sweepScope` emits `reported` from six strictly **in-scope** arms: an unenumerated ancestor, a path that does not resolve lexically (on both the creation and the restore side), a failed unlink, a non-restorable pre-image, and a failed restore. `restored` and `removed` are in-scope only. Out-of-scope writes reach `reported` through a separate whole-root `git status` delta.
- It need not be a **write**: an in-scope node with nothing to read is recorded present with no pre-image and is never `isUntouched`, so a FIFO, socket, or device node the spawn never touches takes a `reported` row on *every* run. The module says so itself.

One vocabulary now serves both sites: the **snapshot** is the run's record of the region's own directories, and a **pre-image** is one path's state in it.

**`#1132`.** Step 9 item 2 calls 7d's `kind` and `cause` paragraphs "the authority on both fields' values **and remedies**", and item 2's own reason for reporting each kind distinctly is that each has a different remedy. Neither target stated one: the `kind` paragraph opened "three values, each with its own remedy" and then glossed *meanings*, and the `cause` paragraph named the ceilings. An agent making the hop found no remedy and had to improvise adopter-facing advice, in a step whose next paragraph forbids fabricating a command for a refusal.

Closed from the target side rather than by narrowing item 2's promise, which would leave both of those clauses pointing nowhere. Each remedy is read off `trySpawn` and `spawnFailureCause`.

**`#1117`, the DECISION OWED.** `UpdateMergeReport` tracks seven lists and the Step 9 template printed a row for six. `merge[]` — the bucket a clean `shared` / `wiki-owned` overwrite reports under — had none, so on a wiki-heavy release the largest group of the run was counted and never printed. Every row the adopter *did* see was truthful, which is what made the omission hard to notice.

The issue named two readings and asked for a maintainer's read of intent. Settled from the code on three convergent points, so no design call was improvised:

1. Six of seven lists map 1:1 to a printed row; `merge[]` is the sole omission.
2. **Handling results** puts `merge[]` in the "report counts only" class beside `overwrite[]`, `skip[]`, `add[]`, `removed[]`, and the Step 9 table is the only place a count is reported — so under the fold reading that line names a bucket reported nowhere.
3. The skill's own opening states the class distinction is observable as "which bucket a clean overwrite reports under", which under a fold is false in every observable sense.

## Verification

The Quality Gate **skips by its own written rule**: no staged file matches its source or gate-affecting-config globs, and the page names `.claude/**` markdown as skipping straight to commit. Its quick-check grep returns nothing.

Run instead, because CI will not: `shell-lint.sh`, the full 17-scenario distribution harness, and `audit-scope-routing-parity.bats`. The standing `.claude/` audits are clean — no absolute paths, no UAT/SPEC narrative refs, no historical phrasing, maintainer-only markers balanced.

**Worth flagging for `#1129` (S19).** `cli-tests.yml`'s distribution-harness `code` filter covers `.gaia/cli/**`, `.gaia/release-*`, `.gaia/tests/distribution/**`, and `cli-tests.yml` — no `.claude/**` path. Scenarios 03, 15, 16, and 17 all read shipped `.claude/` content, and scenario 17 greps *this exact file* for the release-resolved `merge-region` / `regen-regions` invocations. So a `.claude/skills/**`-only pull request skips the job that asserts on it. That is a third victim path for the filter gap, after `wiki/**` and `CLAUDE.md`; recorded on `#1129` rather than filed separately.

Closes #1117
Closes #1131
Closes #1132
Closes #1171


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.obtld5agIX

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.SdTCDBvAMa

PR #1181, opened 2026-08-03
debt(cli): fail the init rename when CLAUDE.md carries no H1 (#1176)

### LINKED ISSUE #1176: gaia init's CLAUDE.md rename silently no-ops since the H1 was removed, so every adopter scaffold gets a title-less CLAUDE.md

stevensacks:
Closes #1176

## What was wrong

`renameClaudeMd` rewrote `CLAUDE.md`'s first `# ` line with a whole-file `String.replace`, which returns the original string on a non-match. Two failures came out of that one line:

- A `CLAUDE.md` with **no** heading took no write and the step returned as though it had succeeded. That precondition stopped holding when the template's own `# GAIA React` heading was dropped, so every adopter scaffolding from the template got a title-less `CLAUDE.md` and no warning.
- `# ` opens a shell comment as well as a markdown heading, so a `CLAUDE.md` whose first `# ` line sat **inside a fenced code block** matched that comment: the step reported success, still wrote no title, and replaced a line of the adopter's own example with the project name.

## The design call, and why

The issue names two directions: insert an H1 when none exists, or refuse loudly and make the template's H1 a checked invariant. **Refuse loudly**, settled from the code:

- **The module never authors content it was not given.** All four rewrites replace a value the seed already carries; none creates a key that is absent. `renameIndexPage`'s own comment states the reason it declines to go further, that a broader rewrite "would clobber `title` keys in extra routes a user may have added (data loss on a diverged file)". Inserting a heading at a position of its own choosing is exactly that guess.
- **The docblock states the heading as a source string** and knows how to mark an optional target, which it does for `_index.ts` (`when the keys exist`) and did not do here.
- **Insert-mode would trade a silent no-op for a silent insert.** `gaia init` would succeed and the distribution scenario would pass, leaving the shipped template title-less indefinitely while every adopter received an H1 the template does not have, a permanent divergence in a `manifest: shared` file that `/update-gaia`'s three-way merge then carries per adopter.
- The refusal can only fire when the shipped template is broken, since at `gaia init` time the file is the template's, unmodified.

## What changed

- The title is the first `# ` line **above the file's first fenced block**; a fence ends the search. `run` turns "no such heading" into a structured `claude_md_heading_missing` error with exit 1, raised **before the first write**, so a run that cannot finish has not renamed `package.json` either and the step is not recorded as completed. A missing `CLAUDE.md` stays tolerated: deleting it outright is an unambiguous choice about a file the adopter owns.
- **Stopping at the first fence rather than pairing openers with closers is the deliberate part.** Whether a later fence closes an earlier one is a CommonMark question about delimiter character and run length, and every wrong answer puts the scan back inside a code block, which audit round 2 demonstrated with a nested sample and with `~~~` inside a backtick block. Nothing has to close for this rule, so no nesting can defeat it, and the fence state machine is gone rather than extended.
- CRLF: `\r` no longer reads as the whitespace after a bare `#`, and the rewritten line keeps the ending every other line has.
- The template's `# GAIA React` heading is restored. Without it the new refusal would fire on every scaffold.
- A test pins the shipped template's heading through `claudeMdHasH1`, the predicate the step itself uses, so tightening the production rule cannot leave the test pinning a looser one. It uses the existing `resolveRepoRootFromImportMeta` fixture that `lint-pin-parity.test.ts` and `distribution-guards.test.ts` already use for repo-level invariants.
- Both binaries rebundled; `HELP_TEXT`'s exit-code line and the module docblock state the precondition.

**Recovery, stated correctly.** `gaia init resume` cannot replay this step: `buildRenameArgv` returns `null` without `step_args.rename`, which only `markStepCompleted` writes, so a step that never completed has no args to resume from. Add a heading, re-run `gaia init rename`, then `gaia init resume` continues with the remaining steps, which is what `.claude/commands/gaia-init.md` already describes. An earlier revision of this branch claimed resume replayed it; audit round 1 caught that and it is corrected here, in the test comment, and in the CHANGELOG.

## Scope note, read this rather than assuming the gap is closed

The template-invariant test narrows the window; it does not close it. No CI lane runs on a `CLAUDE.md`-only pull request: `cli-tests.yml` filters to `.gaia/cli/**` and friends, and `tests.yml` gates every real step behind its own incremental-delta filter. So a future edit that drops the heading still reaches `main` unnoticed and reddens on the next CLI pull request. The difference is that it now reddens **`Vitest (.gaia/cli)`, a required check**, rather than only the advisory distribution harness. Closing the filter gap itself is #1129.

## Verification

- Quality Gate, every step, both workspaces: typecheck, lint, Vitest (root 29 files / 155 tests, CLI 131 / 1930), Playwright 8 + 1 skipped, dev smoke HTTP 200, `pnpm build`. Simplify ran four review agents: two clean, one applied, one skipped with reason.
- **Distribution harness: 17 PASS / 0 FAIL.** `08-gaia-init-cli-sequence.sh` was the standing known-red and this change retires it.
- `shell-lint.sh` clean: 193 scripts, 181 bats suites, 1 husky hook. No bats suite covers `.gaia/cli/src/init/**`; its coverage is the co-located Vitest suites plus distribution scenarios 07 and 08, all run.
- **Every new mechanism proven able to fail.** Six mutants, each killed by named fixtures: dropping the fence stop (4 tests), dropping the `\r` strip (1), dropping the `\r` carry (1), the pre-fix whole-file predicate (2), reverting the precondition hoist (3), and dropping the round-1 fence toggle (2). **One mutant survived and was acted on rather than excused:** tightening the heading pattern to `/^# +\S/` had no fixture that could tell it from `/^#\s/`, and it would have rejected `#\tTitle`, a valid ATX heading, so the tightening was dropped rather than given a fixture invented to justify it.
- Both failure paths driven end-to-end through the rebundled adopter binary: headingless and nested-fence fixtures each give a structured error, exit 1, `package.json` untouched, no `init-state.json`, and the fenced line preserved byte-for-byte.

## Audit

Three rounds of `code-audit-maintainer-node`, the only member `resolve-audit-spawn.sh` named. Rounds 1 and 2 each raised one Important, both real, both reproduced independently before fixing; round 3 cleared. The findings shrank each round (any `# ` inside any fence → a mis-nested fence with no heading above it → a hand-typed newline in `--title`) and each repair made the code smaller, which is what distinguishes this from a repair-your-own-repairs runaway.

### Filed

- **#1182**, `--title` is never validated, so a title carrying a newline injects lines into `CLAUDE.md` and breaks the module's documented idempotency (three identical runs append three times, each exiting 0). Round 3's Important. Out of scope here: different defect, different cause (flag validation, not heading detection), and pre-existing. A waive was licensed, and it was filed instead so it is tracked rather than merely recorded.

### Waived

- `.gaia/cli/src/init/rename.ts:148` — the trailing-`\r` rule has two copies, `withoutCr` on the read side and an inline `endsWith('\r')` on the write side; two statements of one rule that must stay in lockstep, with no live failure. Round 3 Suggestion. Non-security, on a path this pull request already changes, and one rule at two adjacent sites has not been observed to spread, so the waive term applies cleanly. Folding it in would buy a full re-dispatch of an already-marked member for a nit.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.1wEuzaM3pe

PR #1183, opened 2026-08-03
debt(audit): evaluate the over-budget remedy set independently (#1149)

### LINKED ISSUE #1149: gaia-audit: over-budget remedy set collapses to the one remedy #1143 blocks, so splittable files are never proposed
`.claude/skills/gaia/references/audit.md:227`

## Failure mode

Step 3 offers three remedies for an over-budget auto-load file: "propose one of: inline facts → wiki, consolidate duplicated sections, or split into narrower files." Only the first requires a `promote` action, and therefore only the first is exposed to the same-file `promote`+`shrink` drift-check hazard tracked in #1143. The other two are expressible as a `shrink`/`replace` on a single path, or a `shrink` plus a new sibling file, and neither touches `source_expect_sha256` at all.

Nothing in Step 3 says the three remedies are independent, so an audit run that finds the wiki-extraction route blocked concludes the file is unfixable and proposes nothing.

This happened concretely. In `KNOWLEDGE-2026-08-02-1425.md`, the audit reported `CLAUDE.md` at 495/400 words and `.claude/rules/playwright.md` at 211/200 lines, then declined to act on either, filing `oos-001` (now #1143) as the blocker. Every sentence of that finding's `failure_mode` reasons about moving content to the wiki. Splitting `playwright.md` was never evaluated, even though its "Asserting on errors, watch both channels" section spans lines 63-140 (78 lines) and lifting it into a sibling rule file lands the parent near 140 lines, comfortably under the 200-line budget, with no `promote` involved.

The consequence is recurring, not one-time: for as long as #1143 is open, every audit round re-reports the same over-budget files, declines them for a reason that only covers one of three remedies, and dedups its own re-filing against the already-open issue. The work never gets proposed and the `over budget` nudge never clears.

## Suggested fix

Make the remedy set explicit and independently evaluated in Step 3: require the audit to consider each of the three in turn and propose the first that is expressible, rather than treating the file as blocked when one remedy is unavailable. Record the blocked remedy per-remedy rather than per-file, so a report says "wiki extraction blocked by #1143, split proposed instead" rather than "no action proposed."

Note that the third remedy shifts the target for a `.claude/rules/*.md` file: the budget is per-file, so splitting one rule into two rules satisfies it. Step 3 should say whether that is intended or whether a sibling rule counts against a separate aggregate, because the audit's own Step 1 inventory already tracks the conditionally-loaded rule aggregate (8,792 words in the cited report) without setting a budget on it.

Handler: prompt

stevensacks:
Closes #1149

## What was wrong

`/gaia-audit` Step 3 offered three remedies for an over-budget auto-loaded file as a
pick-one menu, with nothing saying they are independent:

> For each over-budget file, propose one of: inline facts → wiki, consolidate duplicated
> sections, or split into narrower files.

Only the first needs a `promote`, so only the first is exposed to the same-file drift that
blocks it. A run that found that route blocked concluded the **file** was unfixable and
proposed nothing at all. The other two were never blocked and were never evaluated.

Not theoretical: `KNOWLEDGE-2026-08-02-1425.md` reported `CLAUDE.md` and
`.claude/rules/playwright.md` over budget, declined both, and reasoned only about the wiki
route. The condition is still live and worse, `CLAUDE.md` is now 570/400 words.

## What changed

Step 3 now evaluates all three in order, proposes the first that applies and is expressible
as an action, and records a blocker **against the remedy** rather than against the file. Each
remedy names its action shape, and the file may no longer be reported as unfixable on the
strength of one blocker.

The report schema's `Over-budget files:` line carries the per-remedy outcome, so the
requirement has a landing site in the strict schema Stage 1 writes.

## Decisions

**The DECISION OWED, settled from the code, not from a question.** The budget is **per file**,
and splitting one over-budget rule into two that are each under budget satisfies it. There is
no aggregate over `.claude/rules/*.md` and none is introduced. Three convergent readings:

1. `.gaia/scripts/check-updates.sh`'s `project_drift` is the oracle that *raises* the
   `over budget` nudge Step 3 answers. It loops per rule file against `AUDIT_RULE_BUDGET=200`
   and breaks on the first over; there is no total anywhere in the signal. An audit asserting
   an aggregate would propose work that cannot clear the indicator.
2. Step 3's own table row: "Any single `.claude/rules/*.md` | ≤200 lines", rationale
   "Focused rule".
3. The schema's one aggregate, `Auto-load total ... (budget: {total budget})`, was filled by
   the live run as `600 = 400 CLAUDE.md + 200 hot.md`, the two **unconditional** files. The
   rules aggregate is reported as inventory with no budget attached.

**One qualifier travels with it.** A split is a remedy only when the extracted section is a
self-contained topic that can carry its own `paths:` scope. Rules load by `paths:` glob, so
two same-glob siblings always co-load: halving a file satisfies the count and reduces nothing
a session loads. This does not reject the issue's own example; it makes the proposer name the
sibling's scope.

**The issue's premise is corrected.** It states the split is "expressible as a `shrink` plus a
new sibling file". The shrink half is expressible; **nothing creates the sibling**. The action
vocabulary is four types, and `promote`'s target is a wiki page which it also indexes and logs,
so using it for a `.claude/rules/*.md` sibling is a category error *and* re-enters the same
same-file drift that blocks remedy 1. So a split routes to `## Out-of-scope findings`, which
the playbook already defines as what "none of the four action types above can fix", and Stage 2
files it as tracked work with the section, the sibling, and both resulting counts named.

**Adding a fifth action type was declined.** It would touch the report template, `## Ordering`,
the per-action loop, post-apply verification, and the "four action types" phrasing at three
sites. That is a design fork, not a `difficulty:medium` prompt fix.

**No issue numbers in the prose.** This file is `manifest: owned` and ships to adopters, where
a `gaia-react/gaia` issue number identifies nothing. The blocking condition is described. A
guard pins the whole file against any `#<digits>` citation.

## Verification

- New guard `.gaia/tests/lib/doc-audit-remedy-set.bats`, **20 tests**, in the directory
  `audit-ci-tests.yml` actually runs. Its paths filter already covers both
  `.claude/skills/gaia/references/audit.md` and `.gaia/tests/lib/**`, verified in the filter
  block, so no workflow change is needed. (The guard shipped at 16 tests and grew to 20
  through the audit rounds below.)
- **RED first:** 15 of the original 16 failed before the edit (the 16th is a whole-file
  absence, green by construction and mutation-proven separately).
- **Mutation-proven, one assertion at a time:** citing an issue number fails the portability
  check; reverting the schema line fails the schema check; renaming the Step 3 heading and
  emptying the file each leave **zero** surviving tests.
- `.gaia/tests/lib/` full suite at merge: **403 pass, 0 fail**, including
  `doc-difficulty-prose.bats`, which greps this same file. The guard itself: **20 pass, 0
  fail**, re-confirmed under both bash 5.3 and bash 3.2.57.
- `bash .gaia/tests/shell-lint.sh`: passed (193 scripts, 182 bats suites).
- Quality Gate: **skipped by its own skip rule**, read from `wiki/decisions/Quality Gate.md`
  rather than remembered. No staged file matches its source or gate-affecting-config globs;
  the diff is markdown plus one `.bats` suite.

### A guard of mine was hollow, and the second commit is the repair

`extract_section_or_fail` exists so a renamed heading fails loudly instead of greening every
absence check on an empty string. Both call sites piped it into `normalize_ws`, and a
pipeline's exit status is its **last** command's, so the guard's `return 1` was discarded and
`setup` handed back an empty string with status 0. Caught by running the mutation rather than
trusting the helper: renaming the heading left one test passing. Capture first, normalize
second, and assert the file is non-empty rather than merely present.

## Audit rounds

Four rounds, both maintainer members (`code-audit-maintainer-prose`,
`code-audit-maintainer-shell`) each round. Findings narrowed monotonically and every repair
kept the changed block the same size or smaller:

| Round | Prose | Shell |
| --- | --- | --- |
| 1 | 3 Important, 2 Suggestions | clean, 2 Suggestions |
| 2 | 3 Important (two were defects in round 1's own repairs) | clean, 4 Suggestions |
| 3 | clean, 3 Suggestions | clean, 1 Suggestion |
| 4 | **clean, 0 findings** | **clean, 0 Important, 3 Suggestions** |

Round 2's third Important was pre-existing and out of scope, and is **filed as #1184** rather
than waived: it is an `important` in a rule that gates human review, and a waive records
rather than tracks.

### Round 4 Suggestions, accepted and noted rather than folded in

All three are `severity:suggestion` on `.gaia/tests/lib/doc-audit-remedy-set.bats`, and the
reporting member states each has **no live failure mode**. Per the PR Merge Workflow's digest
economics, the PR is clean and both members hold valid current-digest markers with nothing
else rotating either digest, so folding a nit in would buy a full member re-dispatch purely to
re-earn a marker.

- `doc-audit-remedy-set.bats:143` (also 207, 218, 227, 233), shellcheck SC2016 x5 on
  single-quoted `grep -F` backtick patterns. **No action, and the member agrees**: the single
  quotes are load-bearing, since `grep -qF` must see the backticks literally, and
  `shell-lint.sh`'s own header names SC2016 on bats as a structural false positive of the bats
  execution model.
- `doc-audit-remedy-set.bats:103`, the terminator validator does not model `extract_section`'s
  `next`, so a *duplicate* `## Step 3, Auto-load budget` heading with every other H2 removed
  would pass the validator while extraction ran to EOF. The second call site is structurally
  immune, and the member marks the fix optional.
- `doc-audit-remedy-set.bats:269`, the issue-number guard matches `#[0-9]{3,}`, so `#42` would
  pass a test named "cites no issue number anywhere". The repo is past #1000.

- **`.claude/skills/gaia/references/audit.md:223`** — the Budget cell of the `wiki/README.md`
  row in Step 3's budget table holds a stray `,` where a not-applicable marker belongs, so a
  literal Stage 1 reading that cell resolves a budget of `,` for the file rather than reading
  it as "no auto-load budget". Pre-existing, non-security, on a path this pull request already
  changes, and never observed to spread, so the waive term applies cleanly and the spread
  override does not fire.
  Dedup key: `v1 class=doc-table-cell-malformed path=.claude/skills/gaia/references/audit.md line=223`

## CHANGELOG

`### Fixed`. `manifest: owned`, so the product surface moves and the change is
adopter-observable. Decided from the worthiness list, per the retired byte-identical-bundle
test. No **Action required** marker: nothing an adopter must do.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.UAgDlR8jeM

PR #1186, opened 2026-08-03
debt(audit): partition the verification round's Run and Skip gate (#1184)

### LINKED ISSUE #1184: gaia-audit: the verification round's Run and Skip recommendations both fire on a CONFLICT-driven replace, so neither earns (Recommended)
`.claude/skills/gaia/references/audit.md:390`

## Failure mode

The classification-verification round's Run and Skip recommendations overlap, so on a whole class of reports nothing decides which option carries the `(Recommended)` tag.

The two bullets are:

- `:389` **Recommend Run** "when the report contains any memory `delete` / `delete-entry`, or any CONFLICT-driven `replace`, those actions are irreversible or carry contradiction risk."
- `:390` **Skip is eligible** "when the actions are only git-reversible `shrink` / `replace` on in-repo (non-memory) files."

There is only one action type behind both names. The `### Shrink / Convert` block at `:331` carries `type: replace` under the id `shrink-{nnn}`, and Step 2 emits its CONFLICT resolutions as that same `replace` ("Emit a `replace` swapping the contradicting line for a wikilink"). Two other places in the file count exactly four action types, confirming `shrink` and `replace` name one thing.

So a report holding only CONFLICT-driven `replace` actions on in-repo, non-memory files satisfies **both** bullets: `:389` because they are CONFLICT-driven replaces, `:390` because they are git-reversible shrinks on in-repo files. The gate then presents one of the two options first with `(Recommended)` on no stated basis, and the contradiction-risk case `:389` exists to catch can be presented to the operator as skip-eligible. The round is the only thing that hardens Stage 1's classifications before they drive edits, and for memory entries those edits are irreversible with no git undo, so a wrongly-recommended Skip is not cosmetic.

Reachable through the ordinary path: a CONFLICT between a `.claude/rules/*.md` file and its canonical wiki page resolves to exactly such a `replace`, and Step 2 classifies those on every run.

## Suggested fix

At `:390`, drop the alias and let the exclusion carry the distinction, so the two bullets partition instead of overlapping:

> Skip is eligible (a legitimate recommendation) when the actions are only git-reversible, **non-CONFLICT-driven** `shrink` actions on in-repo (non-memory) files

The same `` `shrink` / `replace` `` alias appears at `:429` and `:510`. Neither is defective, both simply name the one type twice; drop them for consistency if the fix settles on one name, but the defect is `:390` alone.

Handler: prompt

stevensacks:
Closes #1184

## What was wrong

`/gaia-audit`'s classification-verification round asks the operator whether to run, and the recommendation is dynamic. The two bullets that decide it did not partition:

- **Run** is recommended when the report holds any memory `delete` / `delete-entry`, **or any CONFLICT-driven `replace`**.
- **Skip was eligible** when the actions were only git-reversible `` `shrink` / `replace` `` on in-repo (non-memory) files.

`shrink` and `replace` name **one** action type, not two. `### Shrink / Convert` is the only block template carrying `type: replace`, under the id `shrink-{nnn}`, and Step 2 emits its cross-store CONFLICT resolutions as that same block. Three other places in the file count exactly four action types and name this one `shrink`.

So a report holding only CONFLICT-driven replaces on in-repo, non-memory files satisfied **both** bullets, and the gate tagged one of them `(Recommended)` on no stated basis, including the Skip the Run bullet exists to prevent. Reachable on the ordinary path: a `.claude/rules/*.md` file contradicting its canonical wiki page resolves to exactly that action, and Step 2 classifies those on every run.

## The fix

**Skip eligibility is now the narrow, exclusive carve-out.** It applies only when every action is a git-reversible, non-CONFLICT-driven `shrink` on an in-repo (non-memory) file, and Run is recommended in every other case. The exclusion removes the overlap; the closing sentence removes the gap. The `` / `replace` `` alias is dropped in favor of the one name `## Ordering` uses, since the alias is what made the two bullets read as disjoint sets.

**The exclusion is resolvable from the report.** This is the audit's find, and it overturned a decline I had made. Keying on "non-CONFLICT-driven" without saying how to establish it is not merely under-specified, it is **fail-open**: a CONFLICT-driven action is emitted as the same `shrink` block as any other, and Step 2's project-internal branch requires no contradiction token in its `reason`, so an executor reading reasons forms a confident "non-CONFLICT-driven" and recommends Skip on exactly the report class the Run bullet exists to force. The report already carries the handle, so no schema change was needed: only `Conflicts: 0` in the Summary establishes the exclusion, and any nonzero count means it cannot be established per action, so Run.

**The `(Recommended)` tag is stated once, as a rule.** It was baked into the Run option's literal, which is correct only when Run is the recommendation, so the Skip-recommended arrangement was rendered nowhere. The order is no longer encoded twice either.

**Three surfaces state this criterion, and all three are now fixed and pinned.** The rule, the Skip option's `description` (what the operator reads at the `AskUserQuestion`), and `wiki/concepts/GAIA Audit.md`. The wiki page was found by two audit members independently after the first pass corrected only the two `audit.md` surfaces. Not cosmetic: Step 2 classifies a contradiction between two committed files as a project-internal CONFLICT and emits a `replace` on the non-authoritative one, so a later run could find the playbook contradicting the wiki page **on this very subject** and resolve it in the wrong direction, reinstating the defect.

`:429` and `:510` keep the `` `shrink` / `replace` `` alias deliberately: neither makes a partition claim, so the alias cannot produce this defect there.

## Guard

New `.gaia/tests/lib/doc-audit-verification-gate.bats`, 12 doc-grep tests over frozen literals, covering all three surfaces plus the tag rule. `audit-ci-tests.yml`'s `code` filter gains `wiki/concepts/GAIA Audit.md` so an edit to the newly pinned page re-triggers the suite that guards it; the workflows member verified by picomatch that no other pattern matches that path, so the entry is load-bearing rather than decorative.

Written test-first: 5 of the original 7 failed against the pre-fix prose and 2 passed, which is what proved the section extraction was live rather than vacuous.

## Verification

- Suite 12/12; `.gaia/tests/lib/` full run green; `bash .gaia/tests/shell-lint.sh` passes and now lints 183 suites where it linted 182; `.gaia/scripts/tests/` green (1355).
- Quality Gate skipped per its own "When to skip" section: no staged file matches the source or gate-affecting config lists.
- **Ten mutations, each killed by named tests**, run from a committed tree: dropping either half of the Run bullet, reverting the rule, reverting the option description, dropping or weakening the `Conflicts:` clause, dropping the tag rule, baking the tag into a label *or* a description, reverting the wiki paragraph. Renaming either section anchor, or emptying either source file, fails every test loudly rather than greening the absence checks. The shell member independently reproduced the mutation set in a scratch tree under both bash 3.2.57 and bash 5.

## Audit

Five rounds. Round 1 dispatched prose + shell; rounds 2 and 3 added `code-audit-github-workflows` once the paths-filter edit put workflow YAML in the diff; rounds 4 and 5 re-dispatched that member alone. Every finding is fixed or dispositioned below. Round 5 came back clean.

- `.github/workflows/audit-ci-tests.yml:315` — the sandbox cost block states the `sandbox:` globs "match 459 of 1471 tracked files"; at this tree it is 460 of 1495. Pre-existing, comment-only, no runtime effect, and the stated ratio ("a bit under a third") still holds at 30.8%. **Deliberately not fixed:** re-tuning a hard count recreates exactly the drift this PR removed one paragraph below it, where a prose description of a same-file list was rewritten three times and falsified three times before being deleted. The next edit to that block should delete the count rather than refresh it.

Two audit Suggestions were accepted and noted rather than applied, both on the digest economics: the gate is green and every member is marked, so folding either in buys a full re-dispatch for a nit. `audit.md:392` could name the field the `(Recommended)` tag goes on ("on its label"); the reporting member itself called the block shippable as-is. And the guard suite could carry a `# shellcheck disable=SC2016` explaining that its single-quoted backtick literals are load-bearing; there is no live failure, and `shell-lint.sh` holds `.bats` at `warning`, so that tier never surfaces in the gate.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.ZoSCjp8yTD

PR #1189, opened 2026-08-03
debt(audit): let a promote carry its own source edit (#1143)

### LINKED ISSUE #1143: audit.md: promote+shrink pair on same file breaks the source_expect_sha256 drift check
**Location:** `.claude/skills/gaia/references/audit.md:355`

**Failure mode:** Step 3's own prescribed remedy for an over-budget in-repo auto-loaded file ("inline facts → wiki") cannot be safely expressed as a promote+shrink action pair when the promote's source and the shrink's target are the SAME in-repo file (e.g. extracting a section of `CLAUDE.md` or `.claude/rules/playwright.md` into a new wiki page, leaving a pointer behind). The fixed global Ordering (shrink before promote) means the shrink mutates the source file first; when the promote step later runs and computes `source_expect_sha256` against that same now-already-edited file, it sees its own prior action as drift and skips, so the wiki page either never gets created (content lost from `CLAUDE.md`/the rule with nothing to show for it) or, if the executor is lenient about the `source_expect_sha256`/`expect_sha256` field-name mismatch, the outcome depends on an interpretation this playbook does not pin down. Concretely blocked today: `CLAUDE.md` is 495/400 words over budget and `.claude/rules/playwright.md` is 211/200 lines over budget; both have real, not-yet-in-wiki content (`CLAUDE.md`'s "## Before responding" section; `playwright.md`'s "Asserting on errors, watch both channels" section, confirmed absent from `wiki/dependencies/Playwright.md` by grep) that a future audit cannot safely move without this gap being closed first. The knowledge-audit round deliberately proposed no action for either file rather than construct an unsafe pair.

**Suggested fix:** Either (a) extend the promote schema so `delete_source_after` can target a partial-file removal (a before/after pair on the source, applied atomically with the wiki write, not as a separately-ordered shrink), or (b) carve out an explicit exception in the Ordering section for a promote/shrink pair sharing one path, running that pair together (promote-then-its-own-shrink) ahead of the general shrink phase, or (c) recompute the drift check for such a promote against the pre-shrink content Stage 1 recorded rather than the live file. Needs a real design decision among these, not a one-line prompt fix.

Handler: plan

stevensacks:
Closes #1143

## What was wrong

`/gaia-audit` Step 3's first remedy for an over-budget auto-loaded file is to lift a section onto a wiki page. Expressing that took a `promote` whose `source_path` is the over-budget file plus a `shrink` (`type: replace`) on that same file leaving a wikilink behind, so the two actions always named the one file. `## Ordering` runs every `shrink` before every `promote`, so the shrink moved the file out from under the promote's recorded `source_expect_sha256`; the promote then read its own pair as drift and skipped, leaving a wikilink pointing at a page that was never written.

The remedy was therefore documented as unavailable outright. It is the only one of the three that reaches the wiki, so a file whose real fix was that route had no fix at all. `CLAUDE.md` is the live instance: it is the always-loaded root file, it cannot be split into two of itself, and its over-budget section needs exactly this route.

A second, compounding defect: the per-action loop's two drift bullets name `expect_sha256` and `before:`/`expect:`. A `promote` carries `source_expect_sha256` and `target_expect`, so **neither bullet reached it** and nothing in the playbook said what Stage 2 checks for a promote at all.

## What changed

`promote` performs the source-side edit itself, so the pair is never constructed.

`delete_source_after: true` becomes `source_action: {delete | replace | keep}` with a `source_before` / `source_after` pair scoped to the `replace` mode. The apply arm acts on the mode, **last**, after the target page, `wiki/log.md` and `wiki/index.md` are written, so a failed wiki write leaves the source intact; an unrecognized mode fails closed rather than guessing. Post-apply verification covers the source side. The promote's drift check is now stated: `source_expect_sha256` against `source_path`, and `target_expect` against `target_page` unless the page is being created.

`## Ordering` and the `delete-entry`-only dependency gate are **byte-unchanged**. That is the point of the shape, not an accident of it.

## Why this shape, and not the two the issue also offered

The issue named three directions and asked for a real decision among them.

- **(b) an ordering carve-out plus a gated shrink** has an in-repo precedent: #593 fixed a structurally similar hazard (an action removing a reference before the action that makes the removal safe) with a reorder plus `depends_on`, and `## Ordering`'s rationale paragraph still records it. It is distinguishable. There the two edits were to **different files** (the index pointer vs the source), so merging them into one action was not available and a cross-action gate was the only shape. Here both edits concern **one file the promote already names and already guards with a whole-file sha**, so merging is available. Taking (b) would put one action type in two ordering phases, widen the dependency gate from `delete-entry` to `replace`, and make Stage 2 detect the pairing, all in the section whose interaction with a two-action pair is the defect.
- **(c) recompute the drift check against Stage 1's recorded content** was declined outright: it makes the check tautological, so a genuine external edit between the two stages stops being caught, and it leaves the source mutated first, so any *other* promote failure still loses the content. It fixes the false positive, not the atomicity.

`promote` **already** mutated its source and already guarded it with `source_expect_sha256`, so partial replacement narrows an existing capability rather than adding a mechanism.

## Guards

New `.gaia/tests/lib/doc-audit-promote-source.bats` (17 tests), in the directory `audit-ci-tests.yml` already runs. Positive half: the three modes, the pair scoped to one of them, the promote's own drift check, the three apply arms, the source-last rule *with its reason*, the fail-closed arm, and the source-side verification. Negative half, which is what keeps this fix from being undone by a well-meaning edit: a whole-file absence check on `delete_source_after` (a half-applied rename leaves Stage 2 with no arm for what Stage 1 wrote), plus pins that `## Ordering` keeps one flat type order with no same-path carve-out and that the `depends_on` gate stays `delete-entry`-only.

`doc-audit-remedy-set.bats` group 3 moves with the contract it guards: three tests pinned remedy 1 as blocked and one pinned a sentence that existed only because it was. Those now pin the new contract, including a negative on **both** superseded sentences so the blocked framing cannot return. Groups 1, 2, 4 and 5 are untouched. `doc-audit-verification-gate.bats` is untouched.

**Guard proofs: 24 mutations from a committed tree**, each asserted to have actually changed the file before the suites ran (a no-op mutation proves nothing). Every mutation killed exactly its intended test and no others. Two vacuity guards proven separately: truncating the file fails every test loudly through the `-s` guard rather than greening the absence checks, and renaming the `### Promote` heading fails the four scoped tests with the "would pass vacuously" diagnostic rather than silently matching an empty section.

## CHANGELOG

`### Fixed`, decided from the worthiness list rather than the bundle hash (the byte-identical-adopter-bundle test is retired). `audit.md` is `manifest: owned` and the change moves what Stage 1 may propose and what Stage 2 does with a source file, so the product surface moves. Cites the issue rather than the PR, matching the four most recent entries above it.

## Scope

No issue number appears in the shipped prose; `audit.md` is `manifest: owned` and a `gaia-react/gaia` number identifies nothing on an adopter clone. The tracked-tree sweep found no third surface stating this contract: `wiki/concepts/GAIA Audit.md` describes only the generic four-type / sha-plus-snippet drift shape, which stays true.

## Audit

Seven rounds, `resolved_mode=local`, main checkout so `#1055`'s fail-open precondition never held. Every member dispatched with **no `model` passed**, sidecars pre-cleared before each wave, the expected tree recaptured before each. `code-audit-maintainer-prose` cleared at round 5 with zero Important; `code-audit-maintainer-shell` refused twice and cleared at round 7.

Two findings are worth surfacing because they were defects in *this branch's own guards*, not in the prose:

- The shell member's first refusal found that the `## Ordering` absence check used needles the document's own vocabulary does not use, so two plausible carve-outs passed with zero failures. The root cause was my mutation proof: it inserted a sentence carrying both literals unbackticked, uniform in exactly the dimension the rule discriminates on.
- Its second refusal moved the entire promote-only bullet verbatim from step 2 into step 3's apply arm with **all 46 tests still green**, because two tests named a placement property and scoped to a section spanning both steps. It also named the process failure behind it: the same class had already gone uncaught once here, when a restructure deleted a check and the suite was edited to match rather than allowed to fail.

Both are fixed and both fixes were re-proven independently rather than taken from the report.

The prose member's round-3 report measured `### Per-action loop` at every commit and found monotonic growth (373 → 417 → 458) against narrowing findings, which is a runaway rather than convergence. That measurement was taken as the finding: the round that followed restructured rather than reworded, dissolving the promote drift special case into the generic bullets, and the section came back down. Final deltas: **+44, +41, −56, +15**.

Three, all non-security, all on paths this pull request already changes, so the touched-file term applies and DRAIN rule 4's spread override does not fire for any of them. `audit_path_is_machinery` returns false for all three, so the eligibility is the touched-file term rather than the machinery term.

- `.gaia/tests/lib/doc-audit-remedy-set.bats:212` — the comment recording why a hollow absence needle was dropped says the string "never appeared in **this file's** history". Run literally against the `.bats` file the comment lives in, `git log -S` returns two commits (the ones that added and removed the needle). The substantive claim is true and was verified of `audit.md` and its pre-rename path, both empty; only the file the sentence names is wrong. Comment accuracy, no assertion affected.
  `<!-- gaia-debt-key: v1 class=holistic/unclassified path=.gaia/tests/lib/doc-audit-remedy-set.bats line=212 -->`
- `.gaia/tests/lib/doc-audit-promote-source.bats:161` — `omit unless source_action=replace` occurs twice in the Promote schema, so the `grep -qF` is satisfied by either occurrence while the test is named for the pair. Dropping the qualifier from `source_after` alone would pass. No failure mode traced.
  `<!-- gaia-debt-key: v1 class=holistic/unclassified path=.gaia/tests/lib/doc-audit-promote-source.bats line=161 -->`
- `.claude/skills/gaia/references/audit.md:249` — the report template's four-backtick fence opens at 249 and closes at 296, so `## Summary`, the `## Actions` preamble and the whole `### Delete` schema render inside the template while the other three schemas sit outside it. Pre-existing since `5c055386` and untouched here; possibly deliberate, since the report genuinely has those sections. Confirmed not to affect any suite's section scoping.
  `<!-- gaia-debt-key: v1 class=holistic/unclassified path=.claude/skills/gaia/references/audit.md line=249 -->`

**No disposition-sidecar entry was written, and that was checked rather than assumed:** `code-audit-frontend` was never in the spawn set across seven rounds, so it holds no earned marker for this content (`dfffa5e1….ok` is absent, as is `dfffa5e1….dispositions.json`). An entry keyed to a marker that does not exist could never be valid. Same reasoning S9a and S9e recorded.

## New tech-debt filed

`#1190`, doc-conformance tests pinning a placement's *rationale* but not the operative clause, so the clause can move with the suite green. Two live instances. **Filed rather than waived although a waive was licensed**, because the class has been observed to spread: it produced this pull request's round-6 Important, was fixed there, and the mutation sweep then found it in two more places. A waive records rather than tracks, and this class has now demonstrated it recurs.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.iKqlsaGpZw

Issue #1191, opened 2026-08-03
gaia init strip-branding never validates --title, and its Storybook escaping omits line terminators, so a multi-line title writes an unparseable preview.tsx


stevensacks:
`.gaia/cli/src/init/strip-branding.ts:139`

`gaia init strip-branding` takes `--title` as free text and never checks its shape: `parseFlags` at `:75` validates presence and nothing else. `debrandStorybook` then splices that title into a single-quoted JavaScript string literal in `.storybook/preview.tsx`, escaping backslashes and single quotes at `:139` but not line terminators.

**Failure mode.** A title carrying a newline or a carriage return emits an unterminated string literal:

```js
const BRAND = {
  brandTarget: '_blank',
  brandTitle: 'Bad
Injected',
};
```

Verified by evaluating the emitted block: `new Function(...)` throws `Invalid or unexpected token`. `.storybook/preview.tsx` is left syntactically invalid, so Storybook fails to start or build in the adopter's freshly scaffolded project, and the step still exits 0 and records itself completed. `writeReadme` at `:119` splices the same unvalidated title into the generated README, which is cosmetic beside this.

**Reachability.** `/gaia-init` collects the project title at Step 2 Q4 and passes it straight through to the shell at `.claude/commands/gaia-init.md:272`, with nothing in between checking its shape. This is the same input path `#1182` describes for `gaia init rename`. `strip-branding` is init Step 3 and `rename` is Step 6, so with `#1182`'s fix in place a multi-line title corrupts `preview.tsx` at Step 3 and then fails the run at Step 6, leaving a half-initialized tree. Both commands ship in the adopter binary `.gaia/cli/gaia`.

**Suggested fix.** Reject a title carrying a line ending in `parseFlags`, as `#1182` does for `rename`, so the run fails before any file is written:

```ts
if (/[\n\r]/u.test(title)) {
  return {message: '--title must be a single line', ok: false};
}
```

Refuse rather than escape or trim: the title is an identity value the init flow is given, not one it may rewrite, which is the convention `rename`'s `--kebab` arm already sets. The one open choice is whether the two commands each carry their own check or share one, since they take the same value from the same source; `src/init/util/` is where a shared check would sit.

Found while verifying `#1182`'s premise during its fix. Out of scope there: different file, and not a path that pull request changes.

Handler: prompt

PR #1192, opened 2026-08-03
debt(cli): refuse a multi-line --title in gaia init rename (#1182)

### LINKED ISSUE #1182: gaia init rename never validates --title, so a newline in it injects lines into CLAUDE.md and breaks the documented idempotency
`.gaia/cli/src/init/rename.ts:113`

`parseFlags` validates `--kebab` against `/^[a-z][\d a-z-]*$/u` and validates `--title` not at all. `--title` is free text: `/gaia-init` collects it at Step 2 Q3 and passes it straight through to the shell at `.claude/commands/gaia-init.md:274`, with nothing in between checking its shape.

`renameClaudeMd` assigns the title into one element of the file's line array, so a title carrying a newline becomes two lines in `CLAUDE.md`. The idempotency guard on the next line then compares `lines[index]`, which is now only the first of those two lines, against the freshly rebuilt two-line heading. They never match, so the rewrite runs again on every invocation.

**Failure mode.** `gaia init rename --title $'Bad\nInjected' --kebab bad` exits 0 and reports success. Three identical invocations leave:

```
# Bad
Injected
Injected
Injected

Body
```

`markStepCompleted` records `rename` as done each time. This contradicts the module's own stated contract at `.gaia/cli/src/init/rename.ts:17-18`, "Idempotent: re-running with the same args is a no-op once the rename has been applied", and it silently accumulates content in a file the adopter owns.

Pre-existing rather than newly introduced: the earlier `String.replace` form appended on every run for the same reason.

**Suggested fix.** Reject a multi-line title in `parseFlags`, beside the existing `--kebab` check, so the run fails at exit 1 before any file is written:

```ts
if (/[\n\r]/u.test(title)) {
  return {message: '--title must be a single line', ok: false};
}
```

Refuse rather than trim or escape. The title is an identity value the init flow is given, not one it may quietly rewrite, and the adjacent `--kebab` arm already sets the convention of rejecting a malformed identity rather than normalizing it. `rename.test.ts` has no case for a title carrying a line ending; add one alongside the existing `exit 1 on invalid kebab` test.

Found by `code-audit-maintainer-node` while auditing PR #1181, and out of scope for that pull request's own issue (#1176), which is about heading detection rather than flag validation.

Handler: prompt

stevensacks:
Closes #1182

`gaia init rename` validated `--kebab` against a regex and `--title` not at all, though `--title` is free text `/gaia-init` collects at Step 2 and passes straight through. A title carrying a line ending landed as two lines in `CLAUDE.md`, and the idempotency guard compared only the first of them against the freshly rebuilt two-line heading, so it never matched: every re-run appended again while exiting 0 and recording the step completed, against the module's own contract that re-running with the same args is a no-op.

`titleFailure` now holds the shapes `--title` may not take, checked in `parseFlags` ahead of the first write:

- **a line ending**, the defect above;
- **blank or whitespace-only**, added from audit round 1. `--title ''` wrote `# ` as the heading, a `CLAUDE.md` carrying a heading and no title, which is exactly the state `claudeMdPreconditionMet` refuses when the file arrives that way. Admitting it re-created through the flag the state #1176 eliminated.

Refuse rather than trim or escape, because the title is an identity value the step is given rather than one it may rewrite, which is the convention the `--kebab` arm already sets.

## Verification

- **RED first**, for both rules: three rows for line endings, two for blank titles, all failing before their guard existed.
- **Eight mutants, all killed.** `[\n]` only → the CR row fails; `[\r]` only → both LF rows fail; line guard removed → all three fail; `test(title.trim())`, i.e. normalize instead of refuse → the trailing-newline row alone fails; blank guard removed → both blank rows fail; `title === ''` → the whitespace row alone fails; `if (false) return titleProblem` → five fail; and the blank rule relocated below `renamePackageJson` → both blank rows fail, which is the mutant that earned round 2's added assertion (it survived the whole suite before it).
- **Quality Gate, both workspaces**, re-run after the last edit. CLI: typecheck, lint, 131 files / 1935 passed + 1 skipped. Root: typecheck, lint, 29 files / 155 passed + 1 skipped, build clean, dev smoke HTTP 200, Playwright 8 passed + 1 skipped.
- **`bash .gaia/tests/shell-lint.sh`** passed.
- **Distribution harness `run-all.sh`: 17 PASS / 0 FAIL**, and `08-gaia-init-cli-sequence.sh` re-run against the final bundle.
- **End-to-end through the rebundled `.gaia/cli/gaia`.** Three consecutive runs with `$'Bad\nInjected'` each exit 1 with `{"code":"invalid_arguments","message":"--title must be a single line"}`, leaving `CLAUDE.md` and `package.json` byte-unchanged with no init state recorded; a `\r` title behaves the same; a valid title still exits 0, writes `# Hello World`, and re-runs as a no-op.

Both binaries are rebundled.

## Audit

Three rounds of `code-audit-maintainer-node`, converging Important+Suggestion → Suggestion → clean.

## Out of scope, filed

- **#1193** (audit round 1, independently reproduced before filing): `--title` reaches a `String.replace` replacement template with live `$` tokens and unescaped quotes at three sites in this file. `Q1 $1 Report` injects capture group 1, `$& Corp` injects the whole match, and `Steve's App` writes an unterminated string literal, each leaving the seeded language files unparseable while the step exits 0 and records itself done. Pre-existing, in three functions this pull request does not otherwise touch. Filed rather than waived although the touched-file term licensed a waive, because the unvalidated-`--title` class has now been observed to spread three times.
- **#1191** (found while verifying this issue's premise): `gaia init strip-branding` takes the same unvalidated `--title` and escapes only `\` and `'`, so a multi-line title writes a `.storybook/preview.tsx` that fails to parse. Not a path this pull request changes, so no waive was licensed.

## Waived findings

None.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.faAWuQ3mQG

Issue #1190, opened 2026-08-03
Doc-conformance tests pin a placement's rationale but not the operative clause, so the clause can move with the suite green


stevensacks:
**Location:** `.gaia/tests/lib/doc-audit-promote-source.bats:263`

**Failure mode:** a doc-conformance test pins the *rationale* for a placement rather than the operative clause it justifies, so the operative clause can be moved or reordered while the rationale stays put and the suite runs green. The document left behind is self-contradictory: it still explains why something happens where it no longer happens. Two live instances in this suite.

The first is the sharper one and nothing else pins it. The test at `:263` asserts `The source goes last so a wiki write that fails leaves it intact` but not the operative `then act on the source **last**` in `.claude/skills/gaia/references/audit.md`'s promote apply arm. Reordering that arm so the source edit precedes the wiki writes, while leaving the rationale sentence in place, keeps every test green. The coherent variants both die (reorder plus rewrite the rationale, or drop the rationale), so what survives is exactly the incoherent partial edit a reviewer is least likely to catch by reading.

The second, at `:387`, has the same shape in the post-apply verification arm: the test pins the reason an absence check on `source_before` would be wrong, but nothing rejects the absence check being added back. Consequence is narrower, a false-negative `applied-partial` on a correct apply rather than content loss.

**Why this is filed rather than waived:** the class has been observed to spread. It first surfaced as an Important in the same pull request's earlier round, where four placement assertions scoped to the whole `### Per-action loop` instead of to step 2, and the entire promote-only bullet could be relocated into the apply step with 46/46 green. That instance is fixed; these two are the same defect in two more places, found by mutation after it.

**Suggested fix:** for each of the two, add a needle for the operative clause alongside the existing rationale needle, in the same scope the sibling assertions use. Both literals occur exactly once in the target file and in no other suite, so neither is a tautology. The general rule is already written into the suite's `scoped` helper header (a test pinning where something sits must scope to that place); this extends it to pinning *what* sits there, not only *why*.

Handler: prompt

PR #1194, opened 2026-08-03
debt(audit): read every member's findings array from stdin (#1187)

### LINKED ISSUE #1187: Code Audit Team specs stage every member's findings sidecar to one shared filename, so a parallel round can publish one member's findings under another's name
`.claude/agents/code-audit-maintainer-prose.md:165`

Every Code Audit Team member spec tells the member to stage its findings sidecar through the shared writer with a bare placeholder path:

```bash
findings_sidecar="$(bash .gaia/scripts/audit-write-findings.sh \
  --root "$AUDIT_ROOT" \
  --member code-audit-maintainer-prose \
  --base "$BASE_SHA" \
  --findings /path/to/findings.json)"
```

The same `--findings /path/to/findings.json` line appears twice in each of the five specs: `code-audit-maintainer-prose.md:165` and `:233`, `code-audit-frontend.md:1036` and `:1231`, `code-audit-maintainer-node.md:181` and `:282`, `code-audit-maintainer-shell.md:210` and `:311`, `code-audit-github-workflows.md:177` and `:278`.

## Failure mode

`wiki/concepts/PR Merge Workflow.md` prescribes dispatching every member **in parallel from one tool-call message**, and members dispatched from one session share a scratchpad directory. Handed `/path/to/findings.json` as the only guidance, a member naturally writes `findings.json` there, so every concurrently dispatched member converges on the same staging path.

The published sidecar path is correctly per-member and per-audit-key (`<audit-key>.<member>.findings.json`), so the resulting artifact is *named* right. The hazard is its *content*: the staging file is what the writer reads and publishes. If member A stages to `findings.json` and member B overwrites it before A calls the writer, A publishes B's findings under A's name. A staging file left by an earlier round is republished the same way, as a fresh report.

That sidecar is each member's report of record, and it is what the merge gate's no-op classifier reads to distinguish a real pass from a member that completed with its report lost (`audit-noop-detect.sh --findings`). So contamination is silent: the gate still goes green, and the operator sees a plausible findings block attributed to the wrong member.

Observed live rather than theorized. During a three-member round on PR #1186, the `code-audit-maintainer-prose` member reported finding a stale `findings.json` in the shared session scratchpad holding another member's findings from an earlier run, and worked around it by writing its own to a distinct filename. Its own note: if members keep colliding on that default filename, one will eventually publish another's findings under its own name.

This is a spec-prose defect, not a defect in `audit-write-findings.sh`, which faithfully publishes whatever path it is handed.

## Suggested fix

Have each spec prescribe a staging filename that is unique per member and per round rather than a bare placeholder, mirroring the naming the published sidecar already uses (the member name plus the audit key are both in scope at that point in each spec, as `--member` and `--base` on the same call). The five specs are the surface to change; the shared writer needs no change.

Handler: plan

stevensacks:
Closes #1187

## What

Every Code Audit Team spec prescribed the findings-sidecar write with a bare `--findings /path/to/findings.json`. Members are dispatched in one parallel wave and share a session scratchpad, so that placeholder made one filename the filename every member picks: one member's array can reach another member's published sidecar under that member's name, and a file left by an earlier round republishes as a fresh report. Both are silent, because the sidecar is the report of record and `audit-noop-detect.sh --findings` reads it to tell a real pass from a member whose report was lost, so the gate still greens over findings attributed to the wrong reviewer. Observed live on PR #1186.

All ten call sites (two per spec, five specs) now pipe the array in through a quoted heredoc, and each spec's trailing sentence states the rule the command encodes instead of offering a staged temp file as the alternative.

## Why stdin rather than a unique filename

The issue suggests naming the staging file per member and per round, mirroring the published sidecar's `<audit-key>.<member>` naming. That closes the cross-member collision but not the stale-round one: the audit key is a base sha plus a branch and rotates between neither, so a member's second round reuses the identical name. Removing the staging file removes both. It is also the shape the writer already supports (`--findings -`), already documents, and already uses in its own suites (`.gaia/scripts/tests/audit-refusal-roundtrip.bats:91`), and `AUDIT_KEY` is not a resolved shell variable in a member session, so the issue's naming would have needed each member to derive the key first.

## Guard

`.gaia/tests/lib/doc-audit-findings-staging.bats`, 8 tests, roster resolved by the `code-audit-*.md` glob so a sixth member joins the guard by existing, with a five-name floor that makes an empty glob fail rather than green every loop. It pins the stdin call shape, the quoted delimiter, heredoc closure, and the operative prose rule (the instruction, not its rationale, per #1190's lesson).

## Verification

- 8 mutants, every test killed by at least one: empty spec → roster floor; the original shared placeholder → stdin/staging/general-path checks; a per-member staging filename → stdin + general-path; unterminated heredoc → closure; unquoted delimiter → delimiter + stdin + closure; operative sentence dropped → prose check; superseded sentence restored → alternative check; prose handing the writer a path → general-path. Every mutant reverted.
- The prescribed command was proved against the real writer in a scratch repo before it was authored into five files: it returns the published path, a finding title carrying `$VAR`, a backtick and single quotes survives byte-intact, and the `[]` arm publishes `{"schema":1,...,"findings":[]}`.
- `bash -n` over all 14 fenced blocks in the five specs that call the writer: every one parses.
- `.gaia/tests/lib` 449/449 and `.gaia/scripts/tests` 1355/1355 green under bash 5; `bash .gaia/tests/shell-lint.sh` clean.
- Quality Gate skips by its own rule: no `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config in the diff.

Note for reviewers: this changes the audit agent definitions, so any clearance already earned on another open pull request is invalidated and those members need re-running before that PR merges.

## Audit rounds

Five rounds of `code-audit-maintainer-shell`, the only member this diff dispatches (it owns both `.claude/agents/code-audit-*.md` and `.gaia/**/*.bats`; `CHANGELOG.md` is out of scope). No Critical and no Important at any round. Six Suggestions across rounds 1-4 were all fixed in the PR, each repair making the suite smaller rather than richer: a header claim corrected, then two counts and an equality replaced by one line-by-line match, then a whole test folded into that rule, then an overclaiming comment deleted. Round 3 found a real hole that way (`--findings - < file` and a collapsed one-line call both slipped the old pair of checks), and round 4's whitespace relaxation closed another (an unterminated heredoc whose delimiter line carried trailing whitespace previously counted as neither opened nor closed, and passed).

## Accepted Suggestion (resolved by the operator, not filed)

Round 5 raised one Suggestion and it is accepted rather than fixed: `doc-audit-findings-staging.bats:138` compares opener and terminator totals, so two coincident drifts can cancel (one collapsed one-line call contributing a terminator without an opener, plus one anchored opener that is never terminated, totals 1 == 1). The member scopes it as the shape the check has had since it was written, not something this round introduced, and reaching it needs both drifts in one spec at once. Four rounds of narrowing, non-gating findings on one 170-line test file is where a further re-dispatch stops paying for itself, so this is recorded here rather than folded in. It is in-scope for this PR's own new code, so it is neither a cross-remit waive nor a tech-debt filing; the next edit to that test should add an anchored line-shape assertion so the collapsed form fails on its own.

## Note on dispatch staleness (not a defect in this diff)

Every member dispatched in this session was rendered from the pre-change copy of its own spec, so their instructions still carried the shared `findings.json` placeholder while the on-disk files no longer do. The auditing member noticed, followed the on-disk contract, and wrote its sidecar through the heredoc. Worth knowing when reading the rounds: agent definitions are loaded when a session starts, so this fix reaches members dispatched from sessions started after it lands.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.1Ylm5EmFKy

Issue #1193, opened 2026-08-03
gaia init rename splices --title into a String.replace template, so a $ token or a quote in the title writes unparseable language files


stevensacks:
`.gaia/cli/src/init/rename.ts:238`, repeated verbatim at `:258` and `:273`

`replaceStringPropertyAll`, `replaceTopLevelStringProperty`, and `replaceMetaTitle` each build a `String.replace` replacement template as `` `$1$2${escaped}$2` ``, where `escaped` is `newValue.replaceAll(/[$\\]/gu, String.raw`\$&`)`. That escaping is for a TypeScript literal, not for the replacement parser: **a backslash carries no meaning in a replacement string**, so `$` followed by a digit, `&`, `` ` ``, or `'` is still read as a match reference. The same functions also never escape the quote character they wrap the value in.

**Failure mode.** Verified by running the three functions verbatim against a seeded `common.ts`:

| `--title` | result |
|---|---|
| `Q1 $1 Report` | `siteName: 'Q1 \siteName:  Report'` — capture group 1 injected, backslash left over |
| `$& Corp` | `siteName: '\siteName: 'GAIA' Corp'` — the whole match injected, quotes and all |
| `Steve's App` | `siteName: 'Steve's App'` — unterminated string literal |

Each leaves `app/languages/en/common.ts` and `app/languages/en/pages/_index.ts` unparseable while `gaia init rename` exits 0 and `markStepCompleted` records the step done, so `gaia init resume` will not retry it. Recovery on a fresh scaffold is `git checkout -- app/languages`, and the next build fails loudly, which is what keeps this out of the Critical tier.

**Reachability.** `/gaia-init` collects the project title at Step 2 Q4 as free text and passes it straight through at `.claude/commands/gaia-init.md:274`. An apostrophe in a project title is ordinary English (`Steve's App`), so the quote arm needs no unusual input at all. Ships in the adopter binary `.gaia/cli/gaia`.

**Suggested fix.** Use a function replacement so `$` is inert, which is the shape `.gaia/cli/src/init/strip-branding.ts:138` already uses one file over, and escape separately for the target literal (backslash first, then the captured quote character):

```ts
source.replace(pattern, (_m, prefix: string, quote: string) =>
  `${prefix}${quote}${escapeForLiteral(newValue, quote)}${quote}`
);
```

[...]

PR #1195, opened 2026-08-04
debt(cli): refuse a multi-line or blank --title in gaia init strip-branding (#1191)

### LINKED ISSUE #1191: gaia init strip-branding never validates --title, and its Storybook escaping omits line terminators, so a multi-line title writes an unparseable preview.tsx

stevensacks:
`gaia init strip-branding` took `--title` as free text and never checked its shape: `parseFlags` validated presence and nothing else. `debrandStorybook` then spliced that value into a single-quoted JavaScript string literal in `.storybook/preview.ts`, so a title carrying a line ending emitted an unterminated literal and left a freshly scaffolded project unable to start or build Storybook, while the step exited 0 and recorded itself completed. `writeReadme` splices the same unvalidated value into the generated README.

`/gaia-init` collects the title once at Step 2 Q4 and passes it straight through to both `strip-branding` (Step 3) and `rename` (Step 6), so this is the same input path `#1182` fixed one file over.

## The decision this step owed

**Where `--title` validation lives: one shared checker at `.gaia/cli/src/init/util/title.ts`**, not a copy per command. Settled from the code:

- Two real callers the day it is extracted. `#1182` left the rules in `rename.ts`; `strip-branding` needs the identical ones now.
- The rules are the same on the same grounds, not merely similar. A blank title writes `# ` as the `CLAUDE.md` heading via `rename`, and `.gaia/templates/README.md`'s first line is literally `# {{PROJECT_TITLE}}`, so `strip-branding` writes that same headed-but-titleless state into the adopter's README. A line ending breaks `renameClaudeMd`'s first-line guard and `debrandStorybook`'s quoted literal alike.
- `init/util/` already exists and holds `state.ts`, so no new convention is invented.

**The policy half, which the follow-on `#1193` inherits: refuse identity-shape violations at the flag boundary, escape value characters at each sink.** Refuse a line ending and a blank, because neither is a title and no sink can render either faithfully. Escape quotes, `$` and backslashes per sink, because `Steve's App` and `Q1 $1 Report` are ordinary titles and each sink escapes differently: a single-quoted JS literal, a markdown heading, a JSON string, a `String.replace` replacement. A single canonical escape at the boundary would be wrong for at least one of them, so the checker refuses and never rewrites the value.

## Scope

`#1191`'s title names two defects. The second is closed by the first rather than fixed separately, and that is verified rather than argued: the three accept-a-quote / backslash / `$` tests were written in the red phase and passed against an unmodified `debrandStorybook`. Once a line ending cannot reach it, the surviving escape set is complete for a single-quoted literal, and the function-replacement form already neutralizes `$`. U+2028/U+2029 are deliberately not in the refuse set: they are legal in ES2019+ string literals, so they carry no consequence, and matching `rename`'s exact rule is what keeps the two commands consistent.

## Verification

- **Five mutants, all killed**, each killing the right rows in *both* command suites, which is what proves the checker is genuinely shared: deleting the line-ending rule (6 tests), deleting the blank rule (4), narrowing `/[\n\r]/u` to `/[\n]/u` (2), `title === ''` for `title.trim() === ''` (2), and moving `removeIfPresent`, the first write, ahead of the refusal (all 5 refusal rows). The last is what earns `expectNothingWritten` its FUNDING assertion.
- **Quality Gate in full, both workspaces.** Typecheck, lint and tests clean on each side (root 29 files / 155 tests, CLI 131 / 1943), Playwright 8 passed + 1 skipped, dev smoke 200, build clean. `rename`'s 21 existing tests pass unchanged, which is the extraction's behavior-preserving proof.
- **Distribution harness 17 PASS / 0 FAIL** (scenario 07 covers this command directly), and `shell-lint.sh` clean.
- **Driven end-to-end through the rebundled adopter binary**, not only the unit suite: all four refusal inputs exit 1 with the right structured error, leave the scratch tree byte-identical by whole-tree checksum, and write no init state. The accept path `Steve's $1 App` exits 0, emits `brandTitle: 'Steve\'s $1 App'`, and parses under `new Function`, which is the issue's own oracle turned around to prove the escape half of the policy.

Closes #1191

## Out-of-scope findings and their disposition

Audit round 1 (`code-audit-maintainer-node`) raised two Critical findings in `rename.ts` at the three `String.replace` sinks: a `$` escaped with a backslash is still read as a capture reference, and the wrapping quote is never escaped, so `Q1 $1 Report` and `Steve's App` corrupt the seeded language files while the command exits 0. Both were reproduced by running the shipped CLI.

**Disposition: already filed as `#1193`, open, and scheduled.** These are not new findings. `#1193` names the same three call sites and was filed during the previous step under DRAIN rule 4's spread override; the plan's ratified answer to that step's escalation keeps it as its own row rather than folding it in here, on two grounds: Invariant 9 closes the batch-by-class route because both fixes change behavior, and the batch-by-one-decision route fails its own resolves-as-a-side-effect test, since neither fix closes any part of the other. What the two share is the policy this pull request settles, which is why this row goes first and binds the next.

The member's report also identified something that **was** this pull request's own defect, and that is fixed here: the new module's policy statement, the strip-branding test comment, and the CHANGELOG entry each read as a claim that every sink already escapes an ordinary title correctly. That is true of `debrandStorybook` and false of `rename`'s sinks. The policy is now stated as what a sink owes rather than as what every sink does, and the strip-branding rows say out loud that they assert one sink only.

Audit round 3 raised one Suggestion, at `.gaia/cli/src/init/rename.ts:41`: the exit-1 cause list that round 2 widened to name an invalid title still omits its peer, `--kebab must be a kebab-case identifier`. Real, and left for a follow-up rather than fixed here.

A stopping rule was written into this step's checkpoint **before** round 3 ran: round 3 is the last repair round, and anything it raises that is not Critical or Important is accepted and noted. Round 3 landed on round 2's own repair, which is the runaway shape, and each widening of a prose list invites the next one. The rule was recorded ahead of the finding precisely so honoring it would be a decision rather than a rationalization.

Nothing here gates the marker: rounds 2 and 3 both cleared with zero Critical and zero Important.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.XdMSv3TsCP

Issue #1202, opened 2026-08-04
The distribution suite has two competing CLI-stderr diagnostic mechanisms and neither lives in lib/lib.sh


stevensacks:
**Location:** `.gaia/tests/distribution/lib/lib.sh:11` (the frozen shared-primitive API), with the duplication spread across nine scenario files.

## Failure mode

Every distribution scenario that invokes the CLI expecting success suppresses its stderr, so the `||` failure branch has no diagnostic to report. The directory now solves that in **two different ways**, and neither lives in the shared library the directory has for exactly this purpose.

**Mechanism A, re-run unsuppressed.** Seven scenarios keep `2>/dev/null` and, on failure, re-run the identical command with stderr visible before failing:

- `07-gaia-init-strip-branding.sh:59`
- `08-gaia-init-cli-sequence.sh:116`
- `10-gaia-scaffold-templates.sh:71`
- `11-gaia-setup-cli-flow.sh:62`
- `12-gaia-ping-events.sh:52`
- `13-gaia-update-merge-workspace.sh:56`
- `14-gaia-update-deps.sh:68`

The block is hand-copied in all seven.

**Mechanism B, capture to a file.** `17-gaia-update-merge-region.sh` redirects stderr to a capture file and prints it from the failure branch (`report_cli_stderr` / `fail_with_stderr`, added when `#1045` was fixed).

Both are file-local. `lib/lib.sh` already hosts `pass` / `fail` / `log` / `require_cmd` and its own header calls itself *"Shared primitives for .gaia/tests/distribution scenarios"*, so this is precisely the kind of helper it exists to hold.

The cost is not cosmetic. A scenario author wanting this diagnostic has no shared helper to reach for, so the next one either copy-pastes one of the two blocks or reaches back for a bare `2>/dev/null` out of habit, which is the defect `#1045` was filed for in the first place.

## Which mechanism should win

Mechanism B, on the merits, and this is a recommendation rather than a settled call:

- It reports the bytes the **failing** run actually emitted. A re-run is a second, different execution: it can succeed (leaving the failure unexplained), fail differently, or double the side effects on the one path already going wrong.
- It is what `#1045`'s own suggested fix names.

Mechanism A's one real advantage is that it needs no capture file and no cleanup, which matters for a scenario with no scratch dir of its own. A shared helper should therefore take the capture path as a parameter rather than assume a `$FIXTURES`-style directory exists.

## Suggested fix

[...]

Issue #1201, opened 2026-08-04
Byte-identity assertions built on $(cat …) strip trailing newlines, so a trailing-newline regression greens the tests named for byte identity


stevensacks:
**Location:** `.gaia/scripts/tests/cost-reprice.bats:181`, and roughly 87 further candidate sites across 14 files.

## Failure mode

Byte-identity assertions across the bats suites are built on `$(cat f)` command substitution, which strips **all** trailing newlines from both sides before comparing. A regression that only adds or drops a trailing newline is therefore invisible to the very tests whose names promise byte identity.

The shape, at the cited line:

```sh
before="$(cat "$LEDGER")"
run run_reprice "$RATES_FULL"
[ "$status" -eq 0 ]
[ "$(cat "$LEDGER")" = "$before" ]
```

A writer that appended a trailing newline to the ledger passes this green.

## Scope

`#1047` named three sites and they are fixed (`region-regen-convergence.bats` x2, `region-marker-conformance.bats`). The same class is unfixed elsewhere. Candidate counts per file, from `grep -rE '\[ *"\$\(cat [^)]*\)" *!?= *"\$|^[[:space:]]*(local )?[a-z_]*(before|after|first|second)[a-z_]*="\$\(cat '`:

| File | Candidates |
| --- | --- |
| `.gaia/scripts/tests/cost-reprice.bats` | 29 |
| `.gaia/scripts/tests/write-audit-remits.bats` | 18 |
| `.gaia/tests/lib/spec-folder-layout.bats` | 15 |
| `.gaia/tests/lib/plan-reconcile.bats` | 4 |
| `.gaia/tests/lib/spec-reconcile.bats` | 4 |
| `.gaia/tests/hooks/provision-worktree.bats` | 3 |
| `.gaia/scripts/tests/cost-consolidate-absence.bats` | 2 |
| `.gaia/scripts/tests/plan-archive.bats` | 2 |
| `.gaia/tests/lib/plan-ledger-update.bats` | 2 |
| `.gaia/tests/lib/spec-ledger-status.bats` | 2 |
| `.gaia/tests/lib/ledger-title-repair.bats` | 2 |
| `.gaia/tests/hooks/audit-disposition-check.bats` | 2 |
| `.gaia/scripts/tests/main-root-lib.bats` | 1 |
| `.gaia/scripts/tests/main-only-lib.bats` | 1 |

[...]

PR #1203, opened 2026-08-04
debt(tests): seven test-accuracy fixes, one reviewer context (#1035, #1036, #1045, #1046, #1047, #1136, #1190)

### LINKED ISSUE #1035: Distribution scenario 6 duplicates scenario 5's assertion, never asserts the already-latest verdict it names, and overwrites a scenario-1 fixture in place
**Location:** `.gaia/tests/distribution/17-gaia-update-merge-region.sh:199` (Scenario 6, post-update idempotence)

## Failure mode

Two problems in one scenario.

**1. The scenario asserts nothing scenario 5 has not already proven.** Its only assertion is `[ "$S6_FIRST" = "$S6_SECOND" ]` (:219), which is stability of a pure function across two invocations on identical inputs — exactly what scenario 5 establishes for the same function on the same code path. It adds no independent failure mode. The scenario's own comment (:203-207) names the interesting property, that the post-update state resolves to `already-latest`, and then never asserts it. The CLI does return that verdict, so the assertion is available and free.

**2. It mutates a scenario-1 fixture in place.** `cp "$FIXTURES/s1-latest.txt" "$FIXTURES/s1-current.txt"` at :208 overwrites a fixture that scenarios 1 and 7 read. This is safe today only because nothing after :208 reads `s1-current.txt` for its pre-update content. Any scenario later inserted above this line, or any reordering of the file, silently gets post-update content and passes for the wrong reason — a green test proving nothing.

## Suggested fix

Keep the stability check, add a verdict assertion on `S6_FIRST` (`verdict === 'already-latest'`), and write dedicated `s6-*` fixtures instead of overwriting `s1-current.txt`.

Handler: prompt

### LINKED ISSUE #1036: Distribution scenario 3 interpolates $FIXTURES into single-quoted JS literals inside node -e, so a quote or backslash in TMPDIR breaks the eval
**Location:** `.gaia/tests/distribution/17-gaia-update-merge-region.sh:143` (and the two lines below it, :144-145)

## Failure mode

`$FIXTURES` is interpolated by the shell directly into single-quoted JavaScript string literals inside a `node -e` body:

```js
baseline: fs.readFileSync('$FIXTURES/s3-baseline.txt', 'utf8'),
```

A `TMPDIR` whose path contains a single quote or a backslash breaks the JS eval with a syntax error, and the failure surfaces as an unrelated parse error rather than as anything a maintainer can act on.

Narrow in practice: `$FIXTURES` derives from `mktemp -d` under `/var/folders` or `/tmp`, so a hostile or merely unusual path is unlikely. It is still an avoidable eval-injection shape in a test that runs in CI.

## Suggested fix

Pass the paths through the environment instead of interpolating them into the JS source, exactly as the sibling probe at `.gaia/scripts/tests/region-marker-conformance.bats:99` already does with `GAIA_FIXTURE_FILE`.

Handler: prompt

### LINKED ISSUE #1045: Distribution scenarios 1-4 discard the CLI's stderr before the failure branch reads it, so a CI failure reports only that the exit code was non-zero
**Location:** `.gaia/tests/distribution/17-gaia-update-merge-region.sh:65` (same pattern at `:96`, `:132`, `:168`)

## Failure mode

Scenarios 1 through 4 invoke the CLI as:

```sh
  --json 2>/dev/null)" \
  || { fail "scenario 1: gaia update merge-region exited non-zero on staged tree"; exit 1; }
```

The `2>/dev/null` discards the CLI's own diagnostic, and the `||` branch that fires on failure has nothing left to report. A CI failure therefore says only that the exit code was non-zero — the one fact already implied by the branch being taken. Whoever is diagnosing it has to reproduce locally to learn anything at all, and this is a distribution test whose whole job is to catch breakage in an environment nobody is sitting in front of.

Scenario 7 (`:224-231`) genuinely needs the suppression, because it asserts failure and the diagnostic is expected noise. Scenarios 1 through 4 assert success, so any stderr they produce is exactly what a maintainer wants.

## Suggested fix

Drop `2>/dev/null` on scenarios 1 through 4. If keeping stdout clean matters, redirect stderr to a temp file instead and `cat` it inside the `fail` branch.

Handler: prompt

### LINKED ISSUE #1046: The convergence parity test gave up the exit-status floor and put nothing in its place, so an early abort in the checker passes it green
**Location:** `.gaia/scripts/tests/region-regen-convergence.bats:118` (`@test "writer repairs a perturbed region and the check's parity finding for that member is gone"`)

## Failure mode

The test deliberately, and correctly, declines to assert `verify-audit-roster.sh`'s process-wide exit status — the comment at `:145-150` explains why, and the reasoning is sound: that status answers several invariants at once and a single parity finding does not determine it.

But having given up that floor, the test puts nothing in its place. After `run_check "$sb"` at `:144`, the only remaining check is:

```sh
if remit_parity_names_member "$output" "code-audit-region-claim"; then
  return 1
fi
```

which passes whenever `$output` does **not** match the parity pattern.

Concrete bad outcome: if the checker regressed into an early abort — a usage error from a renamed flag, an unparseable config, a crash before any invariant ran — `$output` would carry an unrelated error message, match no parity pattern, and the test would go green having proved nothing whatsoever about convergence. The stronger the regression, the more reliably this test passes.

## Suggested fix

Add a floor before the `if`. Either `[ -n "$output" ]`, or better, assert that the status is the invariant-violation code (`1`) rather than a usage or crash code (`2`), which pins the run to "the checker ran and reported" without re-coupling the test to which invariants failed.

Which floor to use is a real choice the adjacent tests and `verify-audit-roster.sh`'s own exit-code vocabulary settle.

Handler: prompt

### LINKED ISSUE #1047: The byte-identical idempotence test compares via command substitution, which strips trailing newlines, so a trailing-newline regression is invisible to it
**Location:** `.gaia/scripts/tests/region-regen-convergence.bats:235` (`@test "a second writer run is byte-identical to the first (idempotent)"`)

## Failure mode

The test's name promises byte equality, but it compares via command substitution:

```sh
first="$(cat "$f")"
```

Command substitution strips **all** trailing newlines from both sides. A writer regression that only added or dropped a trailing newline is therefore invisible to the one test whose name says it would catch exactly that — and a trailing-newline difference in a marker-delimited generated region is precisely the kind of drift the convergence suite exists to pin.

The same shape appears at `:226-228` in this file (`content_a` / `content_b`) and at `.gaia/scripts/tests/region-marker-conformance.bats:315` (`before` / `after`, guarding that the writer never deletes bytes outside a pair it can identify).

## Suggested fix

`cp` the first state aside and compare the files directly with `cmp -s "$a" "$b"`, which is byte-exact and says so. Apply the same at the two sibling sites named above.

Handler: prompt

### LINKED ISSUE #1136: debt(cli): the __proto__ row in computeMissing's prototype-key test asserts a shape ManifestSchema cannot produce
The `__proto__` row in `computeMissing`'s prototype-key test asserts a manifest shape `ManifestSchema` cannot produce.

## Location

`.gaia/cli/src/release/manifest.test.ts:399`, the `'__proto__'` entry in the `test.each` list under `describe('computeMissing: paths named after Object.prototype keys')`.

## Failure mode

This is a test-accuracy defect, not a production defect. The guard it exercises (`lookupClass`, `.gaia/cli/src/release/manifest.ts:502`) is correct, and the assertion is true as written.

The test hand-builds `ManifestShape` objects, so it never crosses `ManifestSchema`. Verified against this repo's own zod:

| manifest path | survives `z.record` parse | reaches `lookupClass` |
| --- | --- | --- |
| `constructor` | yes | yes |
| `toString` | yes | yes |
| `valueOf` | yes | yes |
| `hasOwnProperty` | yes | yes |
| `__proto__` | **no** | **no** |

For `__proto__`, `Object.hasOwn(parsed.files, '__proto__') === false` and it is absent from `Object.keys`. So a committed manifest entry named `__proto__` never lands in `actual.files`, and `computeMissing` would report that path missing on every run: the drift/answer gate blocks permanently. That fails **closed**, never a silent pass, which is why this is a suggestion and not a correctness bug.

The cost is that the row implies end-to-end coverage the suite does not have. A reader concludes all five prototype names are proven to reach the guard through the real parse path; only four are.

## Suggested fix

Either:

- add one case that round-trips through `ManifestSchema.parse(JSON.parse(serialize(...)))`, proving the four surviving names end-to-end and letting the list stay honest about `__proto__`; or
- drop `__proto__` from this list, with a comment noting that `z.record` strips it before `lookupClass` is reached.

The adjacent `describe('ManifestSchema: regions')` and `describe('byte-identity vs generate-manifest.mjs')` blocks already parse through the schema, so the surrounding conventions settle which option fits.

Scope note: this does **not** apply to the sibling test in `.gaia/cli/src/update/merge-workspace.test.ts`. Those tokens are argv strings that never cross zod, so all five prototype names are live there and the list is correct as written.

Context: raised as a non-blocking Suggestion by `code-audit-maintainer-node` on PR #1124 (merged as `98e2f642`), left unactioned there because folding it in would have voided the earned audit marker and bought a full re-audit round.

Handler: prompt

### LINKED ISSUE #1190: Doc-conformance tests pin a placement's rationale but not the operative clause, so the clause can move with the suite green

stevensacks:
Seven tech-debt issues, one reviewer context, one sentence: **the test asserts less than its name claims.** Each member is proved by running a suite rather than by a reading, which is what qualifies them as a batch.

Closes #1035
Closes #1036
Closes #1045
Closes #1046
Closes #1047
Closes #1136
Closes #1190

## What each member changes

**#1045** — distribution scenarios that assert success captured the CLI's stderr into `/dev/null`, so the `||` branch could report only the exit code, the one fact the branch being taken already implies. They now capture it to a file and print it from the failure branch. Scenario 7 keeps its suppression: it asserts failure, so its diagnostic is expected noise.

**#1036** — scenario 3 interpolated `$FIXTURES` straight into single-quoted JS string literals inside `node -e`. The paths now reach node through the environment and the JS body is single-quoted, so no shell expansion enters it at all.

**#1035** — scenario 6 asserted only stability across two invocations, which scenario 5 already establishes for the same function; it now asserts the `already-latest` verdict that is only observable in the post-update state. It also gets dedicated `s6-*` fixtures instead of overwriting a scenario-1 fixture in place.

**#1046** — the convergence parity test declined the checker's process-wide exit status as a verdict, correctly, and put nothing in its place, so an early abort that never reached an invariant would emit unmatched text and green the test. It now pins status `1`, the ran-and-reported code.

**#1047** — three comparisons went through `$(cat …)`, which strips all trailing newlines from both sides, so a trailing-newline regression was invisible to the test whose name promises byte identity. All three now compare files with `cmp`.

**#1136** — the prototype-key `test.each` list hand-built `ManifestShape` objects, so no row crossed `ManifestSchema` and the list implied end-to-end coverage the suite did not have. Four names now round-trip through `serialize` → `JSON.parse` → schema parse; `__proto__` gets its own case recording that `z.record` drops it, and that the resulting permanent drift-gate block fails closed.

**#1190** — two doc-conformance tests pinned the *rationale* for a placement without the operative clause it justifies, so the clause could move while the rationale stayed put and the suite ran green. Both now pin the clause alongside the reason.

## The step's own defect, found by a mutant rather than by reading

`report_cli_stderr` was first written *after* `fail` in each `|| { … }` branch. `fail` returns 1 rather than exiting, and the group is the last command of its AND-OR list, so `set -e` is **not** suppressed inside it and aborts the script the moment `fail` returns. The report was dead code: a diagnostic that looked present and printed nothing, while all nine scenarios still passed. Only a mutant driving a real non-zero exit exposed it. Fixed in `8d1186ee`, with the mechanism pinned in a comment so the order cannot be undone as a tidy-up.

The trailing `exit 1` in each of those branches is unreachable for the same reason. It is **pre-existing**, and is left in place and documented rather than deleted, per the coding guidelines' pre-existing-dead-code rule.

## Guard proofs

Every new assertion was proved able to fail, and where a pre-fix counterpart existed it was shown blind to the same mutant.

| Mutant | Result |
|---|---|
| post-update state perturbed so the verdict is not `already-latest` | killed |
| scenario 1 `--current` → missing file | killed; branch prints the CLI's own `region_file_missing` payload |
| fixture dir forced to `/tmp/gaia-s17-it's a \dir` | pre-fix dies with `SyntaxError: missing ) after argument list`; fixed passes |
| `run_check` stubbed to an early abort (`exit 2`) | killed; **pre-fix returns `ok`** |
| agent file gains one trailing newline between the two reads | killed; **pre-fix returns `ok`** |
| same, on the reversed-pair byte-identity claim | killed; **pre-fix returns `ok`** |
| both sandboxes given the same roster | killed |
| `roundTrip` bypasses `ManifestSchema.parse` | killed |
| `__proto__` added back to the surviving-four list | killed |
| source edit hoisted above the wiki writes, rationale left in place | killed; **pre-fix returns `ok`** |
| absence check reinstated in the verify arm, rationale left in place | killed; **pre-fix returns `ok`** |

One proof was discarded as hollow rather than counted: an end-to-end run under a hostile `TMPDIR` passed, but **macOS `mktemp -t` ignores `TMPDIR`**, so no fixture ever landed in that directory. The real proof forces the `FIXTURES=` assignment instead. On Linux CI `mktemp -t` does honour `TMPDIR`, so the hazard is real on the runner and simply not reproducible that way on a Mac.

## What the Quality Gate's simplify step changed

Two findings applied:

- The eight identical report-then-fail groups collapse into one `fail_with_stderr` helper, so the load-bearing call order is stated once instead of at every site.
- Scenario 6 now **reads** scenario 1's baseline and latest instead of writing byte-identical copies of them. The contrast with scenario 1 is the point of the scenario, reading those fixtures is what scenarios 5 and 7 already do, and the defect `#1035` names is the **write** over `s1-current.txt`, which is gone either way. Verified structurally: no scenario writes an `s1-*` fixture after scenario 1's own block.

Both mutants covering the reworked code were re-run afterwards and still kill.

One finding was **not** applied: moving `report_cli_stderr` into `lib/lib.sh`. That file's header marks its API frozen, and converging the seven sibling scenarios is not this change's scope. Filed instead, and the file's header comment now records why this scenario departs from the sibling idiom rather than leaving the divergence unexamined.

## New tech-debt filed

- **#1201** — the `$(cat …)` byte-identity defect `#1047` names exists at ~87 further candidate sites across 14 other files. Filed rather than fixed: out of scope, and the deeper fix is a shared `assert_files_identical` helper rather than more hand-rolled `cmp` calls.
- **#1202** — the distribution suite now holds two competing CLI-stderr diagnostic mechanisms and neither lives in `lib/lib.sh`.

Both are cross-remit: non-security, on paths this PR does not change.

## Verification

Quality Gate run rather than recalled, and it **applies** here (`manifest.test.ts` matches the `*.ts` trigger). Typecheck 0 root and CLI; `pnpm lint` 0; `pnpm lint:cli` found 3 and they were fixed; root tests 155 passed, CLI tests 1957 passed, no console warnings; Playwright 8 passed; dev smoke HTTP 200; build 0.

Beyond the gate: `shell-lint` passed; `.gaia/scripts/tests` 1401 ok / 0 not ok and `.gaia/tests/lib` 448 ok / 0 not ok, both under bash 5 so local matches CI; distribution harness 34 PASS / 0 FAIL. `pnpm -C .gaia/cli bundle` was run and the tree stayed clean, so no rebundle is owed — checked rather than assumed, per S9i.

## Scope note

`#1045`'s title names scenarios 1-4, and the fix also covers scenarios 5 and 6. That is the issue's **own** stated discriminator rather than an extension of it: it keeps scenario 7's suppression explicitly *because it asserts failure*, so every success-asserting invocation is in scope. Scenarios 5 and 6 assert success and carried the identical `2>/dev/null`. Fixing four of the six sites while rewriting scenario 6's invocations for `#1035` would have left the rule stated correctly on one surface and wrongly on another, which is this repo's most-repeated defect class.

## CHANGELOG

No entry. Decided from the merge workflow's own worthiness list, whose **Not worthy** clause names this case in as many words: *"test-only changes that alter no shipped behavior"*. All five changed files are test files and no command behaves differently.

### DIFF
--- a/.gaia/cli/src/release/manifest.test.ts
+++ b/.gaia/cli/src/release/manifest.test.ts
@@ -461,15 +461,13 @@ describe('computeMissing: paths named after Object.prototype keys', () => {
     const built = manifestWithFiles({['__proto__']: 'owned'});
     const committed = roundTrip(built);
 
-    // Which stage drops it, named rather than left as "the way back in",
-    // because these two assertions are a canary on that stage's behaviour and a
-    // reader who has to go find it cannot tell a real regression from a
-    // dependency moving underneath. `serialize` emits the key, and `JSON.parse`
-    // KEEPS it as an own property; `ManifestSchema`'s `z.record` is what drops
-    // it, by re-assigning each entry into a fresh object, where a plain
-    // assignment to `__proto__` with a string value is a no-op. So a zod bump
-    // that changes record's `__proto__` handling flips the two lines below and
-    // reds this suite.
+    // Which stage drops it, named because these two assertions are a canary on
+    // that stage and a reader who has to go find it cannot tell a real
+    // regression from a dependency moving underneath. `serialize` emits the
+    // key, `JSON.parse` KEEPS it as an own property, and `ManifestSchema`'s
+    // `z.record` is what drops it, by an explicit guard rather than by any
+    // language-level invariant. So a zod bump that reworks that guard flips the
+    // two lines below and reds this suite.
     expect(serialize(built)).toContain('"__proto__"');
     expect(Object.hasOwn(committed.files, '__proto__')).toBe(false);
     expect(Object.keys(committed.files)).toEqual([]);
--- a/.gaia/cli/src/release/manifest.test.ts
+++ b/.gaia/cli/src/release/manifest.test.ts
@@ -461,8 +461,15 @@ describe('computeMissing: paths named after Object.prototype keys', () => {
     const built = manifestWithFiles({['__proto__']: 'owned'});
     const committed = roundTrip(built);
 
-    // `serialize` emits the key perfectly well; it is the way back in that
-    // drops it, so the loss is invisible to anyone reading the written file.
+    // Which stage drops it, named rather than left as "the way back in",
+    // because these two assertions are a canary on that stage's behaviour and a
+    // reader who has to go find it cannot tell a real regression from a
+    // dependency moving underneath. `serialize` emits the key, and `JSON.parse`
+    // KEEPS it as an own property; `ManifestSchema`'s `z.record` is what drops
+    // it, by re-assigning each entry into a fresh object, where a plain
+    // assignment to `__proto__` with a string value is a no-op. So a zod bump
+    // that changes record's `__proto__` handling flips the two lines below and
+    // reds this suite.
     expect(serialize(built)).toContain('"__proto__"');
     expect(Object.hasOwn(committed.files, '__proto__')).toBe(false);
     expect(Object.keys(committed.files)).toEqual([]);
--- a/.gaia/cli/src/release/manifest.test.ts
+++ b/.gaia/cli/src/release/manifest.test.ts
@@ -391,6 +391,15 @@ const manifestWithFiles = (files: ManifestShape['files']): ManifestShape => ({
   version: '1.0.0',
 });
 
+/**
+ * The path a committed manifest actually takes: written by `serialize`, read
+ * back by `JSON.parse`, then validated by the schema. Hand-built
+ * `ManifestShape` objects skip all three, so a case that never calls this
+ * proves nothing about what survives the trip.
+ */
+const roundTrip = (manifest: ManifestShape): ManifestShape =>
+  ManifestSchema.parse(JSON.parse(serialize(manifest)));
+
 describe('computeMissing: paths named after Object.prototype keys', () => {
   test.each([
     'constructor',
@@ -427,23 +436,23 @@ describe('computeMissing: paths named after Object.prototype keys', () => {
   // shape: `ManifestSchema: regions` and `byte-identity vs
   // generate-manifest.mjs` both assert through a real parse.
   //
-  // Every key here is written computed. A bare `{__proto__: 'owned'}` object
+  // Every key below is written computed. A bare `{__proto__: 'owned'}` object
   // literal sets the PROTOTYPE instead of defining an own property, which
   // would make these cases test nothing at all.
-  const roundTrip = (manifest: ManifestShape): ManifestShape =>
-    ManifestSchema.parse(JSON.parse(serialize(manifest)));
-
   test.each(['constructor', 'toString', 'valueOf', 'hasOwnProperty'])(
     'a manifest path named %s survives the real serialize/parse round trip and is still reported missing',
     (file) => {
       const committed = roundTrip(manifestWithFiles({[file]: 'owned'}));
 
       expect(Object.hasOwn(committed.files, file)).toBe(true);
-      expect(computeMissing(manifestWithFiles({[file]: 'owned'}), committed)).toEqual(
-        []
-      );
       expect(
-        computeMissing(manifestWithFiles({[file]: 'owned'}), manifestWithFiles({}))
+        computeMissing(manifestWithFiles({[file]: 'owned'}), committed)
+      ).toEqual([]);
+      expect(
+        computeMissing(
+          manifestWithFiles({[file]: 'owned'}),
+          manifestWithFiles({})
+        )
       ).toEqual([file]);
     }
   );
--- a/.gaia/cli/src/release/manifest.test.ts
+++ b/.gaia/cli/src/release/manifest.test.ts
@@ -418,6 +418,54 @@ describe('computeMissing: paths named after Object.prototype keys', () => {
 
     expect(computeMissing(expected, actual)).toEqual([]);
   });
+
+  // The rows above hand-build `ManifestShape` objects, so not one of them
+  // crosses `ManifestSchema`. Read alone they imply all five names are proven
+  // to reach `lookupClass` through the path a committed manifest actually
+  // takes. Only four are, and the two cases below are what make the list say
+  // so rather than merely be true. The surrounding convention settles the
+  // shape: `ManifestSchema: regions` and `byte-identity vs
+  // generate-manifest.mjs` both assert through a real parse.
+  //
+  // Every key here is written computed. A bare `{__proto__: 'owned'}` object
+  // literal sets the PROTOTYPE instead of defining an own property, which
+  // would make these cases test nothing at all.
+  const roundTrip = (manifest: ManifestShape): ManifestShape =>
+    ManifestSchema.parse(JSON.parse(serialize(manifest)));
+
+  test.each(['constructor', 'toString', 'valueOf', 'hasOwnProperty'])(
+    'a manifest path named %s survives the real serialize/parse round trip and is still reported missing',
+    (file) => {
+      const committed = roundTrip(manifestWithFiles({[file]: 'owned'}));
+
+      expect(Object.hasOwn(committed.files, file)).toBe(true);
+      expect(computeMissing(manifestWithFiles({[file]: 'owned'}), committed)).toEqual(
+        []
+      );
+      expect(
+        computeMissing(manifestWithFiles({[file]: 'owned'}), manifestWithFiles({}))
+      ).toEqual([file]);
+    }
+  );
+
+  test('a manifest path named __proto__ does not survive the round trip, so it never reaches the guard', () => {
+    const built = manifestWithFiles({['__proto__']: 'owned'});
+    const committed = roundTrip(built);
+
+    // `serialize` emits the key perfectly well; it is the way back in that
+    // drops it, so the loss is invisible to anyone reading the written file.
+    expect(serialize(built)).toContain('"__proto__"');
+    expect(Object.hasOwn(committed.files, '__proto__')).toBe(false);
+    expect(Object.keys(committed.files)).toEqual([]);
+
+    // The consequence, and why this is a suggestion rather than a bug: the
+    // entry is absent from `actual.files` on EVERY run, so `computeMissing`
+    // reports it missing forever and the drift/answer gate blocks
+    // permanently. That fails closed, never a silent pass. It is also why no
+    // end-to-end row is possible for this name, and why the unit row above is
+    // the whole of the coverage it can have.
+    expect(computeMissing(built, committed)).toEqual(['__proto__']);
+  });
 });
 
 describe('computeDrift: regionDrift', () => {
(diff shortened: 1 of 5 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.s8HZ69PJoV

Issue #1205, opened 2026-08-04
The lint rule-set parity guard covers one package, while eight more drift the same way and two already have


stevensacks:
`.gaia/cli/src/lint-pin-parity.test.ts:76` pins the lockfile-resolution parity guard to a single package name, `RULE_PACKAGE = 'typescript-eslint'`, while the mechanism that lets that package drift reaches many more.

**Failure mode.** `typescript-eslint` diverged between the two workspaces because it is a direct dependency of neither, is not pinned by `@gaia-react/lint`, and arrives transitively through `eslint-config-airbnb-extended`'s caret, so each `pnpm-lock.yaml` freezes it independently. That is a property of the arrival path, not of the package, and the same preset reaches at least eight other rule-bearing packages the same way: `eslint-plugin-react`, `eslint-plugin-react-hooks`, `eslint-plugin-import`, `eslint-plugin-import-x`, `eslint-plugin-jsx-a11y`, `eslint-plugin-n`, `@stylistic/eslint-plugin`, and `@next/eslint-plugin-next`. Every one of them can diverge with all three current guards green.

**Two are diverged right now**, measured on the `debt/1152-1155-batch` branch after that branch converged `typescript-eslint`:

| package | root `pnpm-lock.yaml` | `.gaia/cli/pnpm-lock.yaml` |
|---|---|---|
| `eslint-plugin-import-x` | `4.16.2` | `4.17.1` |
| `@next/eslint-plugin-next` | `16.2.6` | `16.2.12` |

Nine further rule-bearing packages were checked and agree, so this is a live gap rather than a hypothetical one. `eslint-plugin-import-x` is the one that matters on today's code: import rules run in both workspaces, so an import rule added, removed, or retuned between `4.16.2` and `4.17.1` applies to one workspace and not the other, both lint green, and nothing reports it. That is the identical consequence #1155 describes, one package over.

[...]

Issue #1163, opened 2026-08-02
80 CLI module docblocks sit below the first import, so each reads as JSDoc for an import; the import sort recreates the class


stevensacks:
**Location:** `.gaia/cli/src/release/manifest.ts:3`, and 79 further files. Nearest sibling
instance: `.gaia/cli/src/release/scrub.ts:3`.

## Failure mode

A module docblock placed below the file's first `import` attaches, as JSDoc, to whatever
import follows it rather than to the module. The file's primary explanation is then
misattributed, and any tooling or reader that resolves JSDoc by adjacency gets the wrong
answer.

This is the same defect `#1034` fixed at `.gaia/cli/src/release/region-registry.ts:1`. That
issue's body asserts it is a lone exception: *"Every sibling module in `release/` and
`update/` leads with its docblock before any import."* That claim is false, and the gap is
large enough to change what the right fix is.

Measured across `.gaia/cli/src/**/*.ts`, counting files that carry both a top-level `/**`
docblock and at least one `import`:

| | files |
|---|---|
| total | 203 |
| docblock leads (correct) | 122 |
| first import leads (orphaned) | 81 |

`region-registry.ts` is one of the 81 and is fixed, leaving **80**. Two of them sit in the
same directory as the file `#1034` cites: `release/manifest.ts:3` and `release/scrub.ts:3`.
Non-test source in the set includes `release/manifest-cli-check.ts`, `update/merge-audit-ci.ts`,
`update/merge-workspace.ts`, `wiki/classify-paths.ts`, `react-perf/reduce.ts`,
`storage/project-id.ts`, and all six `schemas/*.ts`.

## Why it recurs on its own

The mechanism is the one `#1034` guesses at: an import-ordering autofix artifact. Imports
sort alphabetically by specifier, so `js-yaml` sorts above `node:fs`. A docblock written
above what was then the first import is left behind the moment an import that sorts higher
is added. Nothing flags the result, so the count grows by ordinary editing.

The fix does hold once applied: after `#1034`'s move, `pnpm -C .gaia/cli lint` passes with
the docblock at line 1 and the sort rule does not move it back. So this is not a fight with
the linter, it is an absence of one.

## Suggested fix

Two parts, and the second is what makes the first durable:

1. Decide whether docblock-first is the convention. At 122 to 81 it is the majority but not
   an overwhelming one, and `#1034` was filed and accepted on the correctness argument (a
   module docblock should document the module), not on a headcount.
[...]

PR #1206, opened 2026-08-04
debt(cli): harden the .gaia/cli workspace and stop typescript-eslint drifting past the parity guard (#1152, #1155)

### LINKED ISSUE #1152: The .gaia/cli pnpm workspace does not inherit the root workspace's supply-chain hardening

### LINKED ISSUE #1155: typescript-eslint floats transitively, so the two workspaces lint against different rule implementations with the pin-parity guard green
`pnpm-lock.yaml:5115` resolves `typescript-eslint@8.59.2` while `.gaia/cli/pnpm-lock.yaml:3180` resolves `8.65.0`, so the two workspaces lint against different `@typescript-eslint/*` rule implementations. This is #1051's failure mode surviving one layer below the fix that closed it.

**Why the `@gaia-react/lint` pin does not cover this.** `typescript-eslint` is a direct dependency of neither workspace, and the preset does not pin it for consumers. It arrives transitively: `@gaia-react/lint` pins `eslint-config-airbnb-extended` at `3.1.0`, and that package declares `typescript-eslint: ^8.58.0`, a caret. Each workspace's lockfile therefore freezes it independently, at whatever the caret resolved to on the day that lockfile was last written.

**Failure mode.** Every `@typescript-eslint/*` rule implementation and every `tseslint.configs.*` preset that airbnb-extended spreads comes from that package (`typescript-eslint@8.59.2` pins `@typescript-eslint/eslint-plugin@8.59.2` exactly). A rule added, removed, or changed between 8.59.2 and 8.65.0 applies to one workspace and not the other, both lint green, and nothing reports it. That is exactly the silent, self-widening divergence #1051 described, and the parity guard added there (`.gaia/cli/src/lint-pin-parity.test.ts`) passes throughout, because it asserts on the manifest pin.

**Pre-dates the branch that surfaced it.** Before PR #1154 the two resolved `8.59.2` (root) and `8.62.1` (CLI); that PR's install floated the CLI caret to `8.65.0`, which any `pnpm install` in that workspace would have done. The pin alignment #1051 made is correct and this is not a regression from it. It is the residue that alignment could not reach.

**Suggested fix.** Two directions, and the choice is the work:

1. Declare `typescript-eslint` as an exact devDependency in both `package.json` files and extend the existing parity guard to cover it. Matches the convention already in place (exact pins in manifests, a guard asserting on manifests) and needs no lockfile parsing. Costs a direct dependency on a package that is currently transitive, which has to stay in step with what airbnb-extended expects.
2. Assert on the version each `pnpm-lock.yaml` resolves rather than on the manifest pin. Reaches the real value with no new direct dependency, at the cost of a test that reads lockfile internals. The `@typescript-eslint/*` versions are visible in the importers block, in the peer-resolution suffix on the `@gaia-react/lint` entry (`pnpm-lock.yaml:103`, `.gaia/cli/pnpm-lock.yaml:13`).

Either way the guard should fail on divergence rather than leave a docblock as the only record, which is the state today.

Handler: plan

stevensacks:
Drain step **S23**, the `.gaia/cli` supply-chain pair. Both members are one reviewer context: **`.gaia/cli` resolves dependencies under weaker rules than the repository root does, and neither the manifest pins nor the existing parity guard catches it.**

Closes #1152
Closes #1155

## `#1152`, the workspace inherited none of root's hardening

`.gaia/cli` is its own pnpm root, so `minimumReleaseAge` and `trustPolicy` were simply absent there. Both are now set.

**The fix is not settings-only, and the install is what proved it.** Turning the settings on made `pnpm install --frozen-lockfile` fail immediately, naming **nine committed lockfile entries** published inside the 7-day window (`@babel/{generator,parser,traverse,types}@7.29.8`, `baseline-browser-mapping@2.11.9`, `electron-to-chromium@1.5.399`, `enhanced-resolve@5.24.5`, `fast-uri@3.1.5`, `recast@0.23.19`). They were admitted by the last unhardened resolve, when each was two or three days old. Settings alone would therefore have shipped a **red** frozen install, which is worse than the state they replace, so the lockfile is rebuilt under the policy and all nine drop to the newest compliant release.

Note that plain `pnpm install` cannot do this: it verifies *before* resolving, so it fails on the stale lockfile and skips the resolution step that would fix it. `pnpm clean --lockfile` is what the error message itself prescribes.

**Root's exclusion lists are not copied wholesale**, which is the reading the issue asks for. Each entry was checked against this workspace's own closure, and then each was *proved* (see the fixtures below) rather than reasoned about:

| entry | carried over? | why |
|---|---|---|
| `@gaia-react/lint` (release-age) | yes | first-party; present at `.gaia/cli/pnpm-lock.yaml:414` |
| `semver@6.3.1` (trust) | yes | present at `:2867`; its requiring `eslint-plugin-import@2.32.0` at `:1683` |
| `eslint-import-resolver-typescript@3.10.1` (trust) | yes | present at `:1594`; its requiring `eslint-plugin-canonical@5.1.3` at `:1654` |
| `bippy@0.5.42` (release-age) | **no** | zero occurrences; it is the `.playwright/react-perf` harness primitive |
| `chokidar@4.0.3` (trust) | **no** | zero occurrences; it is `@react-router/dev`'s |

## `#1155`, `typescript-eslint` floated past the pin-parity guard

Root resolved `8.59.2`, `.gaia/cli` resolved `8.65.0`: six minors of rule implementations apart, with the existing guard green throughout because it asserts on the manifest and no manifest states this package.

**Direction 2 of the two the issue offers**, chosen from the code rather than by preference. `lint-pin-parity.test.ts`'s own docblock already prescribed it (*"Closing that gap means asserting on the version resolved in each `pnpm-lock.yaml`"*), and direction 1's failure mode is **silent**, which is the property the issue exists to remove: an exact direct pin outside `eslint-config-airbnb-extended`'s range does not error, it installs a second copy while the rules keep coming from airbnb-extended's, leaving the pin decorative and the guard reading it and passing.

**Convergence is up, to `8.65.0`.** That is where `.gaia/cli` already sat, and it is the newest release root's own `minimumReleaseAge` admits (`8.66.0` shipped 2026-08-03 and is inside the window). The root diff is wide, and it was checked rather than accepted on size: **no direct dependency moves**, the `importers` block's only edit is the peer-resolution suffix on `@gaia-react/lint` following its peers, and most of the added entries are ~20 optional `@unrs/resolver-binding-*` platform packages. Root typecheck and lint are clean under the newer rules.

**An override was measured and rejected.** A root `overrides:` entry gives a much narrower lockfile diff (126/123 vs 579/210), so the obvious trick was tried, converge under it and then delete it. It does not work: pnpm records `overrides` in the lockfile, and removing the entry fails the next install with `ERR_PNPM_LOCKFILE_CONFIG_MISMATCH`. Keeping one would be a permanent new mechanism beyond what the issue asks for, and it would make the guard nearly vacuous by preventing the caret float the guard exists to report.

## Two things this change owes itself

1. **A hardening-parity guard.** `#1152` leaves `minimumReleaseAge` and `trustPolicy` existing by hand in two workspace files with nothing holding them together, which is the same shape as the drift that produced #1051. Guarding it is closing a hole this pull request opens, not adjacent debt. The exclusion **lists** are deliberately *not* asserted: each is scoped to its own closure and each file says so, and a "shared subset" is a rule neither file declares.
2. **Three paths-filter entries, not one.** The guards were reachable only on a `.gaia/cli/**` change. Root `pnpm-lock.yaml` and `pnpm-workspace.yaml` now join `package.json` in `cli-tests.yml`'s `code` filter, because each guard has a different root-side trigger: a manifest bump for the pin, a **re-resolve** for the transitive version (which touches no manifest at all), and a settings edit for the hardening. Without them a guard first fires on some later unrelated pull request, which is the #1129 class.

## Verification

Quality Gate run in full: typecheck 0 root / 0 CLI; lint 0 root / 0 CLI (the CLI's first run found **7 real errors**, since fixed, one of which deleted a redundant sort); tests 155 root / **1963** CLI, no console warnings; Playwright 8 passed; dev smoke HTTP 200; build 0. Beyond the gate: `shell-lint` passed, `retrigger-reachability` 23 ok / 0 not ok, `region-marker-conformance` 6 / 0, `verify-required-checks` 14 / 0.

**11 adversarial mutants, 11 killed, 0 survived**, run from a committed tree behind a baseline gate that refuses to report unless the suite is green first. That gate paid for itself on the first run: the harness passed `--reporter=basic`, which Vitest 4 removed, and it died. Without the baseline check it would have reported seven dead mutants as seven kills.

Seven against the assertions: root drifting back to `8.59.2` (the real defect, reproduced), a second `typescript-eslint` in each lockfile, each workspace dropping each hardening setting, the CLI relaxing the window below root's, and `packages:` renamed so the reader must **throw** rather than return `[]` on both sides and pass vacuously.

Four against the pnpm settings themselves, because a setting present in a file is not a setting in force and pnpm caches its verification:

- raising the window to a year → `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` on 367 entries;
- dropping `semver@6.3.1` → `ERR_PNPM_TRUST_DOWNGRADE`, and the same for `eslint-import-resolver-typescript@3.10.1`. **Both exclusions are load-bearing, not copied.**
- dropping `@gaia-react/lint` → **no violation.** Reported as a negative rather than quietly dropped: that entry is **inert today**, because the pinned version already clears the window. It is still correct to carry, since its purpose is the *next* first-party publish. The three exclusions were not all proved the same way, and a reader should not infer that they were.

## Out-of-scope findings

**Filed: #1205** (`severity:suggestion`, `difficulty:hard`). The guard covers one package; the same preset reaches at least eight more the same way and **two diverge today**, measured on this branch: `eslint-plugin-import-x` is `4.16.2` at root against `4.17.1` in `.gaia/cli`, and `@next/eslint-plugin-next` is `16.2.6` against `16.2.12`. Filed rather than waived although the touched-file term licensed a waive, under DRAIN rule 4's spread override on *observed* spread: this is #1051's class on its third and fourth package. Not folded in here because widening the constant is not the work; deciding **which** packages earn parity is, and the `@next` divergence shows the answer is not "all of them" in a repo that runs none of its rules.

**Recorded, not filed** (both non-security, both on paths this pull request does not change):

- `wiki/decisions/Quality Gate.md` lists `pnpm-lock.yaml` as gate-affecting config but anchors its quick-check grep at `^`, so `.gaia/cli/pnpm-lock.yaml` reads as skippable. The gate was run here on substance rather than on that grep's literal text.
- Root's `pnpm-workspace.yaml:25` excludes `chokidar@4.0.3` from `trustPolicy`, but root now resolves `chokidar@5.0.0`, so that exception has been inert for some time.

`lint-pin-parity.test.ts`'s module docblock sits below the first import. That is #1163's class exactly, already filed and already scheduled; it is deliberately **not** fixed here, because #1163 records that the import sort recreates the class and owns the sweep.

## Audit rounds

**Round 1** dispatched `code-audit-frontend`, `code-audit-github-workflows`, `code-audit-maintainer-node`. Zero Critical, **two Important**, two Suggestions, two cross-remit.

`code-audit-maintainer-node` did not take this PR's 11 mutants on faith: it built an isolated fixture repo with its own `.git` and synthetic manifests, ran the **unmodified** test file against it, and swept 13 mutants of its own. **Three survived**, and both Importants are the same root cause, a class this PR's own sweep never constructed:

- `toBeDefined()` is only `!== undefined`, so the hardening block passed on `minimumReleaseAge: 0`, `trustPolicy: none`, and a YAML null left by deleting a value and keeping its key. Every one of those is the unhardened state #1152 removes, reached by a one-character edit, and parity could not see them either because a symmetric weakening keeps both sides equal. The sweep here dropped each setting and never **weakened** one.
- Omitting the exclusion lists left the escape hatch for the setting beside them unguarded, so exempting a package in `.gaia/cli` alone was invisible. The justification written for omitting them was falsified by `.gaia/cli/pnpm-workspace.yaml` itself, which states the relation outright.

Both were fixed in `eb9fdfa7`: floor-and-value assertions per side, and **containment** rather than equality for the lists, which is the direction the file states. The Suggestion in the same round (`typeof [] === 'object'`, so an array-shaped `packages` produced a bare length failure instead of the throw the comment promises by name) was fixed too, since it landed on original code rather than on a previous round's repair.

Seven further fixtures against the repair, all killed, including all four of the member's survivors. One method correction worth recording: the first asymmetric control raised `minimumReleaseAge` in `.gaia/cli` above root's and **appeared** to survive. It had not: raising that workspace's own window makes pnpm's deps-status check refuse before vitest runs, because the committed lockfile then violates the wider window, so the mutant killed the harness rather than the test. Re-run from the root side it fails correctly. A mutant that prevents the suite from running proves nothing about the suite.

**Round 2** dispatched `code-audit-maintainer-node` alone. Confirmed by re-running `resolve-audit-spawn.sh` rather than predicted: the repair touched only that member's content, so the other two markers stayed valid and two member rounds were not re-paid.

- `pnpm-lock.yaml:5347` — dedupable duplicate resolutions grew from 10 at the base to 37 at HEAD (`pnpm dedupe --check`). `unrs-resolver` now sits at both 1.11.1 and 1.12.2 with 19 duplicated native bindings, and a second `@typescript-eslint/*` island at 8.61.1 sits beside 8.65.0; both islands have live consumers, so this is un-consolidated peer-keyed resolution rather than dead weight. Deliberately **not** auto-fixed: a trial `pnpm dedupe` in an isolated worktree removed 46 packages but moved seven **production-reachable** resolutions (`postcss`, `nanoid`, `prettier` among them) and swept in the 10 duplicates predating this branch, which is not surgical and belongs in a deliberate `/update-deps` pass. No CI gate on `pnpm dedupe` exists, so nothing currently fails on it. Worth carrying to that pass: the retained `brace-expansion@5.0.6` is the sole trigger for high advisories 1123898 and 1130591, and consolidating onto the 5.0.8 this PR already introduces would close both. Dedup key `v1 class=holistic/unclassified path=pnpm-lock.yaml line=5347`.

- `.github/workflows/cli-tests.yml:236` — the **`distribution-harness`** job's `code` filter omits root `pnpm-lock.yaml` and `pnpm-workspace.yaml`, which its own scenarios 04 and 05 install from after both ship into the staging tarball. Pre-existing, on the other job, and outside this PR's changed line ranges. **Recorded on #1129 as a comment rather than filed here**, because #1129 is *"the distribution-harness paths filter covers none of the files the marker-strip assertions inspect"*, the same job and the same filter, so a new issue would duplicate it. That makes four instances of this class in one file counting the two this PR fixes, which is an argument for a check that every path a job's tests read appears in that job's filter.

**Also filed: #1207** (`severity:suggestion`, `difficulty:medium`), from `code-audit-maintainer-node`'s cross-remit finding. `.gaia/cli/pnpm-workspace.yaml` is matched by **no** Code Audit Team member, verified mechanically via `audit_owner_for_path` rather than read off the globs, so a PR whose only change is lowering `minimumReleaseAge` there dispatches an empty member set and merges unaudited. Filed rather than waived because the path it fixes (`.gaia/audit-ci.yml`) is one this PR does not change. It is a recurrence, not a novel shape: #1008 was the same defect for `.husky/**` and #806 a routing variant on the same file, both closed. The new parity test is a partial mitigation and should not be overread as closing it, since it guards the four values it names and cannot review an added key or a changed `allowBuilds` entry.

**Rounds 3 and 4** each dispatched `code-audit-maintainer-node` alone, and each found real defects in the guard rather than in its subject.

Round 3 **falsified this PR's own terminality claim**, which is what it was asked to do. Two more weakening paths, both measured against real installs: a leading `!` is treated as **inversion** by pnpm's single-pattern matcher, so `'!zzz-not-a-real-package'` matches every package and exempts everything while never reading as a wildcard; and the list reader collapsed anything non-array to `[]`, so writing the list as a scalar (`minimumReleaseAgeExclude: '*'`, which pnpm iterates character by character) exempted everything while **four assertions passed vacuously at once**. The structural lesson is the durable one: **a blacklist can only enumerate the ways in**, which is why each round found one more. The fix replaced it with a positive predicate for a legal npm specifier.

Round 4 could not falsify the new claim, and verified it independently rather than resting on this PR's fixtures: it transcribed pnpm 11.9.0's matcher pipeline from source, confirmed only `*` and a leading `!` broaden a name, ran both real closures (965 distinct names), and swept every 1-to-4-character string over the predicate's own class — **zero accepted entries match more than one real package**. A permissive version half is safe because pnpm runs each version through `semver.valid` and fails the install on anything inexact.

It also found **two false positives in this PR's own guard**, both of which would have red a declared-required context on a valid config:

- pnpm **writes** `name@1.0.0 || 2.0.0` into `pnpm-workspace.yaml` itself when it merges exemptions, and the predicate rejected any entry containing a space — so the guard would have failed on a list pnpm authored.
- A present-but-empty key parses to YAML `null`, not `undefined`, so the reader threw. pnpm treats that exactly as absent and builds no exclude policy, meaning the workspace exempts **nothing** — the guard would have failed on the most hardened configuration available.

Both are fixed, along with a latent third (the name class was lowercase-only, while legacy names such as `JSONStream` remain installable). Six fixtures confirm all three now accept and the three earlier weakenings still die.

**Also filed: #1208** — and round 5 then **falsified its premise**, so read it with the correction below. As filed, it claimed neither workspace sets `minimumReleaseAgeStrict` and that its default is the **weak** value, making the 7-day window self-waiving. The first half is true and the second is not.

---

**Round 5** dispatched `code-audit-maintainer-node` alone, owed only because the round-4 repair rotated its digest. **Clean pass: zero Critical, zero Important, one Suggestion, zero cross-remit.**

It converted round 4's empirical closure into a **structural** one. pnpm's `parseVersionPolicyRule` splits at the first `@` after index 0 and hands **only the name half** to `createMatcher`, so the permissive version half can never reach the matcher at all. The predicate's captured name half was proved equal to pnpm's derived `packageName` on all 92 accepted cases, and codepoints `0..0x2FFFF` were swept for any broadening character reachable in the name half: **zero**. Both round-4 widenings are therefore breadth-neutral by construction — uppercase adds letters and no metacharacter is a letter, and the union form lives entirely in the version half. The 965-name sweep re-ran anyway and agreed across 3,894 candidates. Neither widening has a false-negative twin: the regex uses the identical atom class on the first and every subsequent atom, and `value == null → []` matches both pnpm consumers, which are falsy on null. The ReDoS fix measured as genuine rather than cosmetic — the naive form runs 0.23 → 0.93 → 14.5 → 235 ms across 10/14/18/22 atoms where the shipped form stays flat at ~0.002 ms.

**Correcting #1208.** Round 5 reported that pnpm defaults `minimumReleaseAgeStrict` to **true** whenever `minimumReleaseAge` is explicitly set. That contradicts an issue this PR filed, so it was checked against the pinned bundle rather than taken on the return: `pnpm@11.9.0/dist/pnpm.mjs:146622-146623` reads `if (explicitlySetKeys.has("minimumReleaseAge") && minimumReleaseAgeStrict == null) minimumReleaseAgeStrict = true`. Both workspace files set it explicitly, so **strict mode is already in force in both**, and `createMinimumReleaseAgeHandler` (`:195630`) throws on an immature resolution under CI and prompts on a TTY. The hard install failure #1208 proposed adding is the behavior this PR already ships. What survives is narrower and stays open on the issue: `pickManifestUpdates` sits outside the `if (!strictMode) return` guard, so the exemption **write** is genuinely ungated by strict mode, and a config that explicitly sets `minimumReleaseAgeStrict: false` does degrade to the described behavior. #1208 is corrected by comment and rescoped rather than closed. Round 4 was wrong on a fact, round 5 caught it, and the correction is recorded rather than dropped.

**Accepted-and-noted, then filed as #1209** (`severity:suggestion`, `difficulty:medium`) — `.gaia/cli/src/lint-pin-parity.test.ts:307` and `:376`. The round-4 commit taught the *exactness* predicate pnpm's merged union spelling but left the two *containment* assertions doing literal string membership. Root's closure is a superset, so root can legitimately need two versions of a shared package where `.gaia/cli` needs one; pnpm then writes root's as `semver@6.3.1 || 6.3.2` while `.gaia/cli` keeps `semver@6.3.1`, and the containment assertion reds on a genuine subset. Verified through the repo's own `expect`.

**Not fixed, and the reasoning is recorded because it was eligible to be.** A stopping rule written before round 1 permits a fix for a finding showing the guard failing a valid configuration, and this meets that arm in principle. It was declined on four grounds: it does not fire today (every shared entry is byte-identical) and is not a regression (before the round-4 commit the same pair redded on the exactness test instead); when it fires it is a loud red naming the entries, never a silent pass, so whoever meets it has better information than a blind fix now; the repair means reproducing pnpm's spec expansion, which is the hand-rolled-parser shape the merge workflow names as the canonical multi-round trap; and the write path is **less** automatic than the finding implies, since under strict mode the union is written only after a maintainer approves at a prompt. Digest economics agree independently: this PR is clean and the member is marked, so folding it in buys a full round 6 solely to re-earn the marker. Filed rather than left as a body note because the union-spelling class has now been *observed* at two sites in one file rather than predicted at one.

Five rounds, five earned clearances, and the round-over-round finding rate on this predicate is now closed by grammar rather than by enumeration.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.D5C15Gznug

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Q3GP22AsUM

PR #1210, opened 2026-08-04
debt(storybook): the preview carries no env values, and the test suite stops reading .env (#1020, #1027, #1022)

### LINKED ISSUE #1020: Storybook preview env shim is tree-shaken out of every built preview
The Storybook preview env shim is tree-shaken out of every built preview, so the whole `define` pipeline in `.storybook/main.ts` only ever runs under `storybook dev`.

**Location:** `.storybook/preview.ts:6` (`import './env';`), interacting with `"sideEffects": false` in the root `package.json`.

**Failure mode**

`.storybook/env.ts` is pulled in as a bare side-effect import. The root `package.json` declares `"sideEffects": false`, so the bundler is free to elide it, and it does: after `pnpm build-storybook` (what the Chromatic workflow runs) and after a plain `storybook build`, the built output contains zero occurrences of `window.process`, `API_URL`, `COMMIT_SHA`, `MSW_ENABLED`, or `npm_package_version`, while other `preview.ts` content (the `gaiareact.com/docs` brand URL, the `chromatic/index.ts` module) is present. `window.process.env` therefore stays `{}`, the value `preview-head.html` seeds it with.

Concretely: a story that renders a component calling `getBaseUrl()` (`app/services/api/utils.ts:108`, `if (window.process.env.API_URL) return window.process.env.API_URL;`) resolves the configured API base under `storybook dev` and silently falls through to the `''` fallback in the published Chromatic snapshot. The snapshot then passes against the wrong state, and the dev/CI divergence is invisible. No story reads it today, so the defect is latent rather than active.

The second cost is dead configuration: `.storybook/main.ts` maintains a `previewEnvKeys` list, a `loadEnv` call, and a generated `define` map, all type-pinned to `clientSchema`, that have no effect on any artifact the build publishes.

**Suggested fix**

Decide which of these the preview should be, then make the code say it:

1. The built preview *should* carry env values: stop relying on a bare side-effect import. Export an initializer from `.storybook/env.ts` and call it from `preview.ts`, or add `.storybook/**` to a `sideEffects` array in `package.json`. Verify by grepping the build output for `window.process`.
2. The built preview *should not* carry env values (this is the security-favorable reading, and matches the note in the `CHANGELOG.md` entry for #996 that nothing reached the published build): delete `.storybook/env.ts`, the `previewEnvKeys` list, the `loadEnv` call, and the `define` block, and have stories that need env values receive them as props or story args instead.

Either way, add a check that fails when the two drift again, since the current type-pin only guards the key set, not whether the shim survives the bundle.

Handler: plan

### LINKED ISSUE #1022: vitest.config.ts merges every .env value into process.env and forwards it to every test worker
## Failure mode

`vitest.config.ts:12`

```ts
export default defineConfig(({mode}) => {
  Object.assign(process.env, loadEnv(mode, process.cwd(), ''));
```

`loadEnv(mode, cwd, '')` passes an empty prefix, so it returns every key of every `.env` file rather than a prefixed subset, and `Object.assign` writes all of them into the Vitest main process's `process.env`. `vitest.config.ts:42` then sets `env: process.env`, forwarding that merged object into every test worker's environment.

Any test file, and any transitive dependency loaded in that process, can therefore read the developer's real `.env` values from `process.env`. The committed `.env.example` includes `SESSION_SECRET`, so a developer whose local `.env` holds real values has them readable by every package that loads in the test process.

This is byte-for-byte the pattern removed from `.storybook/vite.config.ts` in #1018, for the reason that PR's own CHANGELOG entry gives: it stops merging every `.env` value into the build process's environment, where any plugin, addon, or transitive dependency in that process could read it.

Scope, stated honestly:

- **CI is unaffected.** `.env` is gitignored, so it does not exist in a CI checkout. `loadEnv` returns only `process.env` there and the merge is a no-op. The exposure is local-developer-machine only.
- **It raises an already-present local risk rather than creating one.** Code running in the Vitest process runs as the developer and could read `.env` from disk directly; the merge saves an attacker a file read rather than granting new reach. It still matters, because generic exfiltration payloads harvest `process.env` far more readily than they read a named dotfile. That is the same defense-in-depth reasoning that made the pattern worth removing from the Storybook config.

## Suggested fix

Mirror the Storybook change: read the loaded env into a local binding and forward only the keys the test environment actually needs, instead of merging everything into `process.env`.

The merge may already be largely redundant. `test/setup.ts` supplies `??=` fallbacks for `API_URL`, `SESSION_SECRET`, and `SITE_URL`; Vitest supplies `NODE_ENV`; pnpm supplies `npm_package_version` (#1021 covers the one gap in that set). Confirm the suite passes with the merge removed before removing it, since a developer may be relying on a real `API_URL` from their `.env`.

If the conclusion is that tests should keep seeing the developer's real environment, that is a deliberate tradeoff rather than a defect. Record it in a comment on the line so a later audit does not re-raise it.

## Related

- #1018 removed the same pattern from `.storybook/vite.config.ts`.
- #1000 (closed) is the Storybook half of the same secret-exposure concern.
- #1021 covers the `npm_package_version` fallback gap in `test/setup.ts` that this fix interacts with.

Handler: plan

### LINKED ISSUE #1027: Storybook preview env shim replaces process.env wholesale inside the Vitest worker
The Storybook preview env shim replaces `process.env` wholesale inside the Vitest worker, so the test environment silently loses every variable the shim does not name.

**Location:** `.storybook/env.ts:6` (`window.process.env = {`), reached from `test/setup.ts:3` (`import * as globalStorybookConfig from '../.storybook/preview'`) via `.storybook/preview.ts:6` (`import './env';`).

## Failure mode

`.storybook/env.ts:6` **assigns** a fresh object literal to `window.process.env` rather than merging into it:

```ts
window.process.env = {
  API_URL: import.meta.env.API_URL,
  COMMIT_SHA: import.meta.env.COMMIT_SHA,
  MSW_ENABLED: import.meta.env.MSW_ENABLED,
  NODE_ENV: import.meta.env.NODE_ENV,
  npm_package_version: import.meta.env.npm_package_version,
} satisfies Record<keyof Window['process']['env'], unknown>;
```

That is written for the browser, where `window.process` is a shim seeded by `preview-head.html`. In the Vitest happy-dom worker it is not a shim: `window.process` **is** Node's `process`, so the statement discards the real environment object and installs a five-key literal in its place. `test/setup.ts` is the global `setupFiles` entry, so this runs once for every test file in the suite, before the `??=` defaults on `test/setup.ts:13-16` execute.

Measured inside a test on the current tree:

```
window.process===process: true | typeof window.process: object
envKeyCount: 7 | hasPATH: false
npm_package_version: 1.6.1 | MSW_ENABLED: true
```

Seven keys: the five above, plus `SESSION_SECRET` and `SITE_URL` added afterwards by `test/setup.ts`. `PATH` is gone, and so is everything else Node started with.

Consequences, in order of how close they are to biting:

1. **Already wrong, currently masked.** `test/mocks/ping.ts:5` reads `process.env.REMIX_DEV_ORIGIN`, which is not in the surviving set, so it can never see a configured value and always takes its `?? 'http://localhost:3001'` fallback.
2. **Test/production fidelity gap.** `app/env.server.ts:21` runs `schema.parse(process.env)`. Under test it validates a five-to-seven-key object that cannot occur at runtime, so the parse is not exercising the shape it exists to guard.
3. **Latent.** No test currently spawns a child process, but any future `execSync`/`spawn` that inherits `process.env` gets no `PATH`. Any new `.env` key an adopter adds is invisible to tests unless it is also added to the `previewEnvKeys` list in `.storybook/main.ts`.

Nothing currently fails, because every present consumer either reads one of the surviving keys or has a `??` fallback. This is a landmine rather than an active break, which is why it is filed at `severity:suggestion` rather than higher.

## Interaction with #1020

\#1020 reports that this same shim is **tree-shaken out of built previews** (`"sideEffects": false`), so it never runs in a published Chromatic snapshot. Taken together, the Vitest worker and `storybook dev` are the only places the shim actually executes today, and in the Vitest worker its main observable effect is this clobber.

The two issues must be resolved consistently:

- \#1020's option 2 (delete `.storybook/env.ts` and the `define` pipeline) resolves this issue as a side effect.
- \#1020's option 1 (keep the shim, make it survive bundling) does **not** resolve this issue, and should be paired with the fix below.

## Suggested fix

Merge instead of replace, so the statement adds the inlined keys without discarding the host environment:

```ts
Object.assign(window.process.env, {
  API_URL: import.meta.env.API_URL,
  // ...
} satisfies Record<keyof Window['process']['env'], unknown>);
```

The `satisfies` pin that `.storybook/main.ts:14` depends on is preserved by keeping the object literal as the `Object.assign` source. Alternatively, guard the assignment so it only constructs the object when `window.process?.env` is absent, which keeps the browser path identical and makes the Node path a no-op.

Whichever is chosen, add a test asserting that a variable outside the inlined key set survives into the worker, since nothing currently detects this.

Handler: prompt

stevensacks:
Drains the env/Storybook trio as one unit, because they are one decision rather than three fixes.

Closes #1020
Closes #1027
Closes #1022

## The decision

`#1020` offers two directions and they are mutually exclusive: either the built preview *should* carry env values (fix the tree-shaking) or it *should not* (delete the pipeline). Taken here: **it should not.** Grounds, each checked against the repo:

1. This repo has already chosen the security-favorable reading on this exact surface twice, in `#1000` and `#1018`, and the `#996` CHANGELOG entry records that nothing reached the published build. A Chromatic snapshot is a public artifact. Option 1 would newly push deployment values into it, which is a regression against the state the repo believes it is in.
2. Zero consumers. The only readers of `window.process.env` are `app/entry.client.tsx`, which Storybook never loads, and `getBaseUrl()` in `app/services/api/utils.ts`, which already guards before use. No story reads a client env value.
3. `app/utils/environment.ts` reads `process.env.*`, substituted by the **root** `vite.config.ts` define, a config Storybook does not use. The two never met, so the pipeline was not serving the app's own env readers either.
4. It resolves `#1027` outright, per that issue's own **Interaction with #1020** section, rather than leaving a shim that has to stay correct.

Option 1's mechanics were weighed and declined too: adding `.storybook/**` to a `sideEffects` array edits `package.json`, which is `manifest: shared` and changes bundling for the app build as well; and exporting an initializer re-introduces module-order fragility, which is exactly `app/utils/environment.ts`'s shape.

## Premises verified rather than taken

- **`#1020`'s tree-shaking claim is confirmed, and the obvious alternative explanation was excluded.** A build showed 0 hits for `API_URL`, `COMMIT_SHA`, `MSW_ENABLED` and `npm_package_version` while control strings from the same `preview.ts` were present. That alone does not distinguish "tree-shaken" from "the define map was empty because the key was unset", so a second build ran with those keys set to unique sentinel values in the build process's own environment, which `loadEnv(mode, cwd, '')` includes because an empty prefix matches every `process.env` key. All three sentinels were absent and the control was present.
- **`#1022`'s exposure is live, not theoretical.** Measured by key name only: on a developer machine with a real `.env`, the removed merge injected four keys into every test worker that the shell did not have, `SESSION_SECRET` among them. A green suite after a deletion proves nothing about whether the exposure existed, so this was measured rather than inferred.
- **`#1027`'s clobber reproduced exactly.** Before the fix the worker's `process.env` held precisely the seven keys the issue reports and no `PATH`.

## What ships

- `.storybook/env.ts` is deleted, along with `previewEnvKeys`, the `loadEnv` call and the `define` block in `.storybook/main.ts`, and the bare `import './env'` in `preview.ts`. `preview-head.html` still seeds `window.process = {env: {}}`, so a client env read is `undefined` rather than a `TypeError`.
- `vitest.config.ts` stops merging `.env` into `process.env` and stops forwarding it via `env: process.env`. CI was already unaffected, since `.env` is gitignored and absent there, so this makes a local run match a CI one rather than changing what CI does.
- `test/preview-env.test.ts` pins the worker half: the preview must not replace `process.env`. Proved red before the fix.
- `app/env.server.ts` records that Storybook is deliberately exempt from the `Window['process']['env']` contract that `app/root.tsx` satisfies, so a story author does not trust a type the runtime no longer honors there.

## Quality Gate

Typecheck 0; `pnpm lint` 0 errors / 0 warnings; unit tests 30 files, 156 passed, 1 skipped, no console warnings; Playwright 8 passed, 1 skipped; dev smoke HTTP 200; `pnpm build` exit 0. `pnpm lint:cli` correctly skipped, no `.gaia/cli/**` file changed.

The Simplify pass produced one head-on contradiction between two angles about the surviving `viteFinal`: one called it a pure duplicate of `.storybook/vite.config.ts`, the other called it load-bearing. Broken with the build rather than split. `viteFinal` is gone and `build-storybook` still exits 0 with every `~` alias resolving. An intermediate half-edit redded that same build, which is what shows the green run is not vacuous.

## Notes

- `/distribution-audit` answered `test/preview-env.test.ts` as `owned`, matching every other file under `test/`. Regenerating also dropped the manifest entry for the deleted `.storybook/env.ts`.
- **Adopter-visible.** Every file touched is `manifest: owned`. A story reading `window.process.env` under `storybook dev` now reads `undefined`, and a test depending on a value from a developer's own `.env` now sees `test/setup.ts`'s default. The CHANGELOG entry carries the **Action required:** line.
- Pre-existing and deliberately untouched: `pnpm build` warns that production source maps are enabled, and `app/utils/environment.ts` has no importers.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Dvu13QsRtr

PR #1212, opened 2026-08-04
debt(audit): refuse a Code Audit Team pass whose review scope is dirty (#1105)

### LINKED ISSUE #1105: code-audit members review working-tree bytes for files their marker covers at HEAD

stevensacks:
Closes #1105

## What

A Code Audit Team member's clearance marker attests to a per-member content digest computed over tracked files **at HEAD** (`git ls-tree HEAD`, `.claude/hooks/lib/audit-digest.sh`), while the member reviews a file by `Read`ing it, which returns **working-tree** bytes. On a dirty tree those disagree, so a pass can write a marker certifying content nobody read. The list side was aligned to HEAD already; this is the read side.

All five members now resolve their review scope and then check `git status --porcelain` over that scope, refusing the pass when it is dirty: no marker, a findings sidecar naming each dirty path, and a re-dispatch once the operator commits or reverts.

## The design decision, and why this posture

The issue names three candidate postures and prescribes none.

- **Warn and proceed** was declined because it does not close the defect the issue states. The complaint is that the marker's guarantee is unconditional while the thing backing it is not; a warning changes the record, not the guarantee, and the gate still clears.
- **Read everything through `git show HEAD:<path>`** was declined as the wrong instrument at this size. It rewrites how every member reads every file, and it collides with the self-heal pass, which genuinely needs working-tree access. Its scoped form is what `code-audit-frontend.md` already advises, and this issue exists **because that advisory is insufficient**: one sentence, in one member, conditional on the member's own judgment, whose failure mode when the judgment lapses is silent.
- **Refuse** is what this repository has already chosen for the same class of divergence. Marker precondition 0 holds that *a marker only ever attests committed content*, so a self-heal pass writes no marker however clean the tree then looks. Self-heal is the dirty-**because-of-me** case; this is the dirty-**when-I-arrived** case. Extending the existing rule beats standing a second mechanism up beside it, and it fails legible rather than silent.

**The issue's stated complication dissolves on timing and scope, which is what makes refusal affordable.** It says a blanket refusal must distinguish "dirty when I arrived" from "dirty because of me". It does not: the check sits at **scope resolution**, which precedes any self-heal in every member's run order, so dirty-because-of-me cannot exist yet. And scoping it to the member's **own resolved `changed` set** rather than the whole tree keeps a concurrently-dispatched sibling's legitimate self-heal in another remit from refusing this member's pass, which is the one real deadlock risk a whole-tree check would carry.

## Verification

- New suite `.gaia/scripts/tests/audit-dirty-tree-posture.bats`, **7/7** under bash 5.
- **The snippet was executed, not just written.** Six cases in a scratch repo: a clean scope reports nothing; a dirty in-scope file is reported; a dirty file *outside* the resolved scope is not; a path containing a space is reported; an empty `changed` neither crashes nor reports; and the issue's own repro (commit a change, revert it in the working tree) is detected.
- **Three adversarial mutants against the real member files**, each redding the right assertion: dropping the check from one member, drifting one member's check by one flag, and moving the check above the derivation it reads.
- **One hollow proof was caught rather than counted.** The first mutant run reported all-green because the mutation never applied, not because the guard was weak; re-run with an occurrence count taken before and after, it reds correctly.
- Neighbouring suites **110/110** (`audit-guard-structural`, `audit-base-agreement`, `check-audit-base-derivation`, `audit-machinery-complete`, `write-audit-remits`), and `bash .gaia/tests/shell-lint.sh` passes.
- `release manifest --check` clean: no newly-shipping file, so no distribution answer is owed.

## Notes

- **Two of the five member definitions ship** (`code-audit-frontend.md`, `code-audit-github-workflows.md`); the three `maintainer-*` members are maintainer-only. The adopter-observable surface is the shipped pair, and the fix lands on all five for parity.
- This edits the audit agent definitions, so **every member's content digest rotates** and any clearance already earned on an open pull request is invalidated.

---

## Audit gate: five rounds, clean on round 5

`code-audit-maintainer-shell` was the only dispatched member (it owns both `.claude/agents/code-audit-*.md` and `.gaia/**/*.bats`, which is the whole diff bar `CHANGELOG.md`). It refused four times and cleared on the fifth. Every finding was real and every one but the last round's was a defect in this PR's own work.

| Round | Findings | Outcome |
|---|---|---|
| 1 | 1 Important, 1 Suggestion | a hollow test assertion; C-quoted paths defeating the new guard |
| 2 | 2 Important, 3 Suggestions | four tautological non-vacuity proofs; the other half of the quoting hole |
| 3 | 2 Important, 4 Suggestions | the guard computed a result nothing could read; a filterable fail-closed sentinel |
| 4 | **1 Critical** | the posture applied byte-identically to a member whose charter forbids withholding |
| 5 | 0 Critical, 0 Important, 7 Suggestions | **clean, marker earned** |

Two things worth carrying out of it. **Round 4's Critical was a round-0 design defect, not a repair-of-a-repair**, which is why the S7 runaway test said keep going: the audit reached back past three rounds of repair to the original choice rather than re-litigating its own prose. And **the suggested fix is a finding, not a specification**: round 2's proposed one-liner would have dropped the default member's `-- '*.ts' '*.tsx'` pathspec and made it refuse on files it never reviews, so the `-z` went onto the existing derivation instead.

## Out-of-scope findings, dispositioned

- **#1213 filed** — the four specialists' `full_changed` self-skip derivation carries the same C-quoting fail-open, with a worse consequence (an entirely unaudited merge rather than a missed dirty check). Pre-existing and a different variable, so out of scope here. Filed rather than waived because the class was observed on two derivations in one round while the default member had already avoided it, which is observed spread.
- **#1214 filed** — two load-bearing lines of this posture are pinned by nothing, so the suite greens while the mechanism is gone. Filed rather than waived on the same spread rule: this is the second instance of that class in this file. Filed rather than folded in because the reporting member had already earned its marker, and editing the file rotates its digest and buys a full re-dispatch solely to re-earn it.

Per a stopping rule pre-committed before each round rather than decided after seeing its findings. All need a filename shape this repository does not contain, and none has a live failure mode here:

- **`changed` paths reach `git status` as pathspecs, which are glob-matched.** A tracked path containing `[`, `*` or `?` would match something other than itself, and a non-matching pathspec exits 0 silently. Zero tracked paths carry a metacharacter. The fix if it ever bites: `--pathspec-from-file=- --pathspec-file-nul` (git 2.25+).
- **`-z` is neutralized by the round trip back through newlines**, so a literal newline in a path still splits into two bogus pathspecs. The `-z` closes the C-quoting case, which was the live one. `changed` has to stay newline-delimited for the remit filter, so such a path is already unrepresentable in the review scope.
- **13 `SC2016` shellcheck infos** on the suite's single-quoted literal pins, where non-expansion is the entire point. Below `shell-lint.sh`'s `warning` floor for `.bats` by policy rather than oversight; the gate passes.
- **A trailing space, and three advisory-member pins without a non-vacuity proof** (`mutate_member` hardcodes the shell member, so it structurally cannot produce one). The helper gap is recorded in #1214.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.L9XbLYCHSY

Issue #1214, opened 2026-08-04
Two load-bearing lines of the dirty-tree posture are unpinned, so its suite greens while the mechanism is gone


stevensacks:
## Two load-bearing lines of the dirty-tree posture are unpinned, so the suite greens while the mechanism is gone

`.gaia/scripts/tests/audit-dirty-tree-posture.bats:78`

The suite exists to keep five agent definitions from drifting apart on the dirty-tree posture. Two parts of that posture are asserted by nothing, and both were demonstrated by mutation against a scratch copy of the tree.

### Failure mode 1: the fail-closed sentinel assignment is pinned nowhere

`SENTINEL_CARVEOUT` pins the *prose* sentence that says the sentinel is never remit-filtered. `CHECK_LINE` pins only the `if` line. The assignment that actually produces the sentinel is pinned by neither:

```
dirty_in_scope="dirty-scope check failed"
```

It lives at `code-audit-maintainer-shell.md:102`, `code-audit-maintainer-prose.md:86`, `code-audit-github-workflows.md:89`, `code-audit-maintainer-node.md:92`, and `code-audit-frontend.md:718`.

Deleting that line from all five members leaves the suite reporting **18/18 ok**, test 7 included, while the mutated block against a broken `git status` prints "check could not run" and then reports a clean tree. The prose keeps promising unconditional withholding after the code has stopped producing anything to withhold on. That is precisely the suite's own stated anti-goal.

### Failure mode 2: the advisory-drift guard only catches the byte-identical copy-paste

`.gaia/scripts/tests/audit-dirty-tree-posture.bats:192` asserts the advisory member does **not** carry the gating withhold contract, using an exact-string absence check. Leaving the exemption paragraph intact and *adding a reworded* withhold clause ahead of `code-audit-maintainer-prose.md`'s "There is no withhold path here" restores the original contradiction in effect, and all 18 tests stay green. The control case confirms the guard does catch the byte-identical paste it was written for; the gap is the reworded variant.

### Why this is filed rather than waived

The touched-file waive term applies cleanly: the cited file is one PR #1212 added. It is filed anyway because **this class has now been observed twice in the same file**, which is the spread condition that overrides a licensed waive. The first instance was an assertion that matched a string every member already carried in unrelated sections, so the obligation it named was pinned by nothing; it was repaired mid-review. These two are the same shape reached from a different direction, and a suite whose whole purpose is preventing drift is the worst place for a pin that cannot fail.

[...]

PR #1215, opened 2026-08-04
debt(cli): state regen-regions write confinement on the axis the code uses (#1179)

### LINKED ISSUE #1179: regen-regions --help states write confinement as a scope binary, but five of the six reported arms are in-scope
`.gaia/cli/src/update/regen-regions.ts:88` (the `HELP_TEXT` block)

Surfaced as a cross-remit finding by `code-audit-maintainer-prose` during PR #1178's round 1. Out of scope there: the pull request changes `.claude/skills/update-gaia/SKILL.md` and `CHANGELOG.md` only, so this path is not in its changed set and the touched-file waive term does not apply.

## Failure mode

`gaia update regen-regions --help` states write confinement as a scope binary:

> Writes outside a region's declared paths are reverted (inside the region's own directories) or reported (anywhere else); a regular file or symlink the snapshot could examine is never silently kept.

The reverted-inside / reported-anywhere-else split is not what the code does. `sweepScope` emits `reported` from six strictly **in-scope** arms:

- an unenumerated ancestor, so an adopter's own file cannot be told from a spawn creation (`:1172`)
- a path that does not resolve lexically, on the creation side (`:1178`) and again on the restore side (`:1220`)
- a failed unlink (`:1189`)
- a non-restorable pre-image, which includes an unreadable file, link, or directory and a node with nothing to read at all (`:1210`)
- a failed restore (`:1260`)

The trailing hedge ("a regular file or symlink the snapshot could examine is never silently kept") narrows the silently-kept claim but leaves the binary itself standing, and it does not cover the node kinds that hold nothing to read.

The consequence is an adopter's search: `--help` tells them a `reported` row means a write that landed somewhere other than the region's own directories, so they look for it outside those directories. For five of the six arms above it is inside them. The same line also implies every `reported` entry is a write, when an in-scope node with nothing to read (a FIFO, socket, or device node) is recorded present with no pre-image, is never `isUntouched`, and so takes a `reported` row on **every** run, including runs the regeneration never touched it (the module states this itself at `:1143`).

This is the same defect PR #1178 removed from the skill's Step 7d and Step 9 item 10. Two shipped surfaces state one rule and only one of them has been corrected, so the binary now survives in the surface an adopter reaches from their own terminal.

## Suggested fix

Restate the help text on the axis the skill now uses: scope decides whether a revert is **attempted**, and inside the snapshot a path the runner cannot put back, or cannot establish it would put back safely, is reported too. The module docblock at `:45`-`:49` is already accurate and can supply the wording, so the fix is a restatement rather than a new claim. Keep it terse; `--help` does not need the six arms enumerated, only the binary dropped.

Handler: prompt

stevensacks:
Closes #1179

## What was wrong

`gaia update regen-regions --help` stated write confinement as a scope binary:

> Writes outside a region's declared paths are reverted (inside the region's own directories) or reported (anywhere else)

`sweepScope` emits `reported` from six strictly **in-scope** arms (`regen-regions.ts:1172`, `:1178`, `:1189`, `:1210`, `:1220`, `:1260`), so an adopter reading a `reported` row searched outside the region's own directories for a write that is inside them. The same line implied every `reported` row is a write, when an in-scope node with nothing to read takes one on **every** run, including runs the regeneration never touched it (`:1143-1146`).

## What changed

Both surfaces are restated on the axis `.claude/skills/update-gaia/SKILL.md:622` already uses: scope decides whether an undo is *attempted*, and a path the runner cannot undo, or cannot establish it would undo safely, is reported too. The wording is a restatement, not a new claim; the module docblock supplied it. `--help` does not enumerate the six arms.

**Two surfaces, not one.** `wiki/concepts/Update Workflow.md:80` stated the identical binary and is `wiki-owned` in the manifest, so it ships. It is corrected here rather than filed: it is the same claim, text-only and zero-behavior. `wiki/concepts/Generated Regions.md` was checked and needs no change, since `:40` already documents the in-scope `reported` cases explicitly.

Both CLI bundles are rebuilt and committed, since `HELP_TEXT` is a string constant rather than a stripped comment. The bundle diff is confined to the `HELP_TEXT` block, and the rendered `--help` was verified from the rebuilt binary rather than from the source.

## No test added, deliberately

No suite pins `HELP_TEXT` (`regen-regions.test.ts` has zero help assertions), confirmed rather than assumed. No test can assert prose accuracy, and a string pin is the presence-only weakness already filed as #1214: a reworded sentence reintroducing the same false binary passes it undetected. Whether this restatement is true is settled by reading `sweepScope`'s six arms, which is why the row is `opus`-authored.

## Accepted and noted, not fixed here

- **"anywhere else it reports the write" is not unconditional.** A write reaching outside the repository through an in-scope symlinked directory is reported by nothing, and the whole-root `git status` half misses a path git cannot see. Both are pre-existing and carried verbatim from the old text, and neither is #1179's subject. The retained hedge ("a regular file or symlink the snapshot could examine is never silently kept") is what keeps the paragraph from reading as an exhaustive partition; enumerating the limits in a usage screen is altitude creep.
- **No cross-reference comment naming the sibling surfaces was added**, though a review angle proposed one and the idea is sound. It makes the artifact richer rather than smaller, and a four-path list nothing checks is a new claim that rots on the first rename.

## Verification

Quality Gate run in full, both workspaces: typecheck clean root and CLI; `pnpm lint` and `pnpm lint:cli` at zero errors and zero warnings; root tests 156 passed / 30 files, CLI tests 1967 passed / 131 files; Playwright 8 passed; dev smoke HTTP 200; `pnpm build` clean; `bash .gaia/tests/shell-lint.sh` passed. `release manifest --check` clean and no file newly ships, so no `/distribution-audit` is owed. Wiki-style and repo-relative-path audits clean on the changed page.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.T2evfdC3Cv

PR #1217, opened 2026-08-04
debt(cli): graduate the hand-written Unreleased block, not the generated one (#1146)

### LINKED ISSUE #1146: graduateChangelog concatenates the generated block with the hand-written Unreleased block, so a release ships duplicate section headings
**Location:** `.gaia/cli/src/release/changelog.ts:365` (the `newSection` assembly inside `graduateChangelog`)

## Failure mode

`graduateChangelog` puts the **generated** block and the **hand-written** `## [Unreleased]` content under the same dated version heading, so a release ships a version block carrying `### Changed` and `### Added` twice.

The splice at `changelog.ts:365-383`:

```ts
const newSection = [UNRELEASED_HEADING, '', versionHeading, '', ...blockLines, ''];
const head = lines.slice(0, unreleasedIndex);   // everything before `## [Unreleased]`
const tail = lines.slice(unreleasedIndex + 1);  // the hand-written entries, then older versions
const graduated = [...head, ...newSection, ...tail].join('\n');
```

`blockLines` is `renderBlock(groupCommits(commits))` (`changelog.ts:469-470`), the block generated from conventional-commit subjects. `tail` begins with whatever was written under `## [Unreleased]` by hand. Nothing separates them, so both land under the one new `## [X.Y.Z] - DATE` heading, generated first. `renderBlock` emits `### Added` / `### Changed` / `### Fixed` from its fixed order array (`changelog.ts:182`), and the hand-written block carries the same headings, so the released block carries each duplicated category once per source.

This is by construction and does not depend on the state of `CHANGELOG.md`. It is a distinct cause from #1029, which was one hand-written block carrying `### Changed` twice on its own; fixing that (PR for #1029) does not touch this.

**The real release path is the non-draft one.** `.claude/commands/gaia-release.md` step 5 runs `release changelog --draft` for human review, then bare `release changelog` to apply, which is the splice above.

**Why no test catches it.** `changelog.test.ts`'s `describe('graduateChangelog')` `TEMPLATE` has an **empty** `## [Unreleased]` block:

```
# Changelog

## [Unreleased]

## [1.0.0] - 2026-01-01

- Old entry
```

With nothing under the heading, `tail` starts at the older version and the concatenation cannot show up. Every `graduateChangelog` assertion runs against that template.

## What has to be settled first

Two in-tree statements of intent disagree, so this is not a mechanical fix:

- the module docblock (`changelog.ts:8`) says that without `--draft`, "the rendered block is graduated", i.e. the generated block is the release content; and
- `gaia-release.md`'s own step-5 comment says `# graduate Unreleased → vX.Y.Z`, i.e. the hand-written block is the release content and the draft is an aid.

GAIA's practice is the second: `.claude/rules/pr-merge.md` makes a hand-written `## [Unreleased]` entry a mandatory maintainer gate on every merge, and those entries are adopter-facing prose that a conventional-commit subject line cannot stand in for.

## Suggested fix

Decide the contract, then implement it. Candidate directions:

1. **Merge, do not concatenate.** Parse the hand-written `## [Unreleased]` body into sections, merge the generated block into it by category, and emit one heading per category in `renderBlock`'s order. Most faithful to both documents, most code.
2. **Graduate the hand-written block only**, and make the generated block draft-only. Matches how the release is actually run, and reduces the non-draft path to a move plus the link-reference rewrite.
3. **Reject a non-empty Unreleased block in the non-draft path**, forcing the maintainer to hand-merge the draft first. Cheapest, but it makes every release a two-step manual edit.

Whichever direction is taken, give `graduateChangelog` a test whose `TEMPLATE` has a **populated** `## [Unreleased]` block. That absence is what let this stand.

Handler: plan

stevensacks:
Closes #1146

## What

`gaia-maintainer release changelog` (non-draft) spliced the commit-derived block above the hand-written `## [Unreleased]` body under one dated version heading, so a release shipped `### Added` / `### Changed` / `### Fixed` once per source, and the duplication then entered permanent history. Reproduced as a unit test before the fix: two `### Changed` in the released block.

The hand-written block is now the release content. The non-draft path renames `## [Unreleased]` in place to `## [X.Y.Z] - DATE`, opens a fresh empty `## [Unreleased]` above it, and maintains the reference-link block. `--draft` still renders the commit-derived block to stdout, unchanged, as the review aid the runbook already presents.

## The contract decision

The issue named three candidate directions and asked for the contract to be settled first. Direction 2 (graduate the hand-written block, make the generated one draft-only), settled from the code and the in-tree documents rather than from preference:

1. **Practice and the merge gate both name the hand-written block.** `.claude/rules/pr-merge.md` makes an `## [Unreleased]` entry a merge gate, so those entries are adopter-facing prose written per PR; `gaia-release.md` says `graduate Unreleased → vX.Y.Z`. The module docblock was the only statement on the other side, and it is the one that described the defect.
2. **Direction 3 (refuse a non-empty Unreleased) would have fired on every real release**, because the merge gate is what makes the block non-empty.
3. **Direction 1 (merge by category) does not remove the redundancy it targets**, only its headings: each change would still appear once as prose and once as its own commit subject, and nothing can dedup those. It also needs a markdown section parser inside `graduateChangelog`.
4. **Direction 2 removes a surface rather than adding one.**

Two consequences of that contract, both deliberate:

- **An empty `## [Unreleased]` body is refused** (`empty_unreleased_section`, exit 1) rather than dated. Direction 2 deletes the only thing that was filling an empty block, so without this a release could silently ship a version heading with nothing under it, and the GitHub release fallback body with it.
- **The non-draft path no longer reads git at all.** `block` is not an input to `graduateChangelog` any more, so collecting commits outside `--draft` was dead work. `--draft` keeps `defaultRunner` and its 64 MiB bound untouched.

## Verification

- **The defect was reproduced before it was fixed**: a `graduateChangelog` TEMPLATE with a populated Unreleased block, which is the absence the issue identifies as what let this stand. Every prior assertion in that describe ran against an empty one.
- **Three mutants, all killed.** Neutering `hasUnreleasedEntries` reds 2 tests; injecting generated-shaped content into the released block reds 2; leaking content into the fresh Unreleased reds 1.
- **End-to-end on a sandbox copy of this repo's real `CHANGELOG.md`**: one of each section heading under the dated version, an empty new Unreleased, the compare link repointed and a release-tag definition added.
- Quality Gate green: typecheck, lint, `pnpm test --run`, `pnpm -C .gaia/cli vitest run`, Playwright, dev smoke (200), build. `pnpm -C .gaia/cli lint` found three (a hoistable test helper, prettier formatting, and `run` passing its cognitive-complexity budget); all fixed.
- Both bundles rebuilt. Only `gaia-maintainer` moves; the adopter `gaia` binary is byte-identical, which is the expected shape for a maintainer-only namespace. `--help` verified from the rebuilt binary rather than the source.

## Documentation surfaces

The old contract was stated in four places besides the code. All corrected here, because they describe the behavior this change moves: the module docblock, `changelog.ts`'s `--help`, the `release` namespace help summary, `.claude/commands/gaia-release.md` step 5, and `wiki/concepts/Release Workflow.md`. The runbook additionally documented the duplication as expected ("can carry duplicated sub-sections"), which is now false.

## Considered and noted

- **`hasUnreleasedEntries` is fence-blind, and it is safe here.** A `## ` line inside a fenced code block in the Unreleased body would terminate the scan, but it can only be reached after the fence's own opening line, which is non-blank and returns `true` first. The first-non-blank-wins ordering is what makes the blindness unreachable.
- **One Simplify finding was declined against an oracle rather than an opinion.** One angle called the `GRADUATE_REFUSALS` table indirection for a two-case union and wanted the inline `if` blocks back; another independently called it a reactive dedup rather than speculative infrastructure. Restoring the inline form re-trips `sonarjs/cognitive-complexity` at 16 against a budget of 15, which is what the table exists to stay inside. Verified by making the change and running the linter, not asserted.
- **One efficiency finding declined.** `hasUnreleasedEntries` slices to the end of the file before scanning, where an index loop would allocate nothing. It runs once per release cut on an array already fully in memory, and the alternative reads worse for no measurable gain.
- **The graduation contract is now stated in five places** (docblock, two help surfaces, runbook, wiki page), each for a different reader. Normal for documentation, but five places to keep in sync if this behavior moves again.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.HhIboHRk3E

PR #1222, opened 2026-08-05
debt(cli): enumerate every router subcommand in both help summaries (#1166)

### LINKED ISSUE #1166: Both CLI top-level help summaries drift from their routers: 21 subcommands omitted across 9 lines
**Location:** `.gaia/cli/src/index.ts:34` (the adopter binary's `HELP_TEXT`), and
`.gaia/cli/src/index.maintainer.ts:30` (the maintainer binary's).

## Failure mode

Both binaries print a top-level summary with one line per namespace, enumerating that
namespace's subcommands. Nine of those lines name fewer subcommands than their router
dispatches, for a total of **21 omissions across two files**.

Adopter summary (`index.ts`), after the `update` line is corrected:

| Line | Lists | Router has | Omits |
|---|---|---|---|
| `wiki` (`:38`) | 10 | 14 | `chain`, `diff-size`, `empty-sections`, `frontmatter` |
| `automation` (`:42`) | 2 | 4 | `install-audit-workflow`, `render-workflows` |
| `update-deps` (`:47`) | 1 | 2 | `decline` |
| `init` (`:48`) | 6 | 8 | `bootstrap-env`, `configure-automation` |
| `setup-ci` (`:50`) | 10 | 13 | `check-audit-drift`, `check-drift`, `write-isolation-policy` |

Maintainer summary (`index.maintainer.ts`) carries the same `wiki`, `automation`,
`update-deps`, and `init` drift, from the same four routers. Its `release`, `setup`, and
`scaffold` lines are accurate, as are the adopter binary's `sandbox`, `scaffold`,
`react-perf`, `fitness`, and `ci-*` lines.

**Nothing dispatches incorrectly, and every namespace's own `--help` is complete and
accurate.** The rot is confined to the two top-level summaries. The bad outcome is a
reader consulting the narrower surface: an operator who runs `gaia --help` to find out
what the tool can do is told about 21 fewer things than it does, and some of the omitted
ones are commands GAIA's own tooling runs.

**The class recurs on its own, which is what makes it worth a guard rather than a sweep.**
A new subcommand is added to its router's `SUBCOMMAND_HANDLERS` and to that router's own
`HELP_TEXT`, both of which sit in the same file the author is editing. The top-level
summary lives in a different file that nothing points at from there, so it is never
updated. That mechanism has produced 21 omissions and will produce more.

It has also already been fixed twice, one line at a time, which is the evidence that
one-at-a-time is not converging: `#1042` corrected the maintainer binary's `update` line,
and `#1164` corrected the adopter binary's. Both were filed as isolated defects. Neither
filing noticed the other eight lines.

## Suggested fix

Two halves, and they should be sequenced rather than merged:

1. **A guard.** `command-reachability.test.ts` already enumerates every
   `SUBCOMMAND_HANDLERS` leaf across both entrypoints and every domain router
   (`extractHandlerKeys` / `enumerateLeafCommands`), which is most of the machinery this
   needs. An assertion that each summary line names every key its router dispatches would
   fail today for all nine lines and would catch the next one at the moment it is added.
2. **The sweep**, once the guard defines what "correct" means.

A sweep without a guard is a nine-line diff that starts decaying on the next subcommand
added, which is exactly how the current 21 accumulated.

## What the code does not settle

**Must a summary line be exhaustive?** The code answers ambiguously. `sandbox seed` is
listed in `INTERNAL_COMMANDS` in `command-reachability.test.ts` as invoker-less by design,
yet it appears in the adopter summary's `sandbox` line, so the current text is not applying
a "user-facing commands only" rule. But `wiki`'s router at 14 keys would produce a summary
line well past any reasonable width, so "list all of them" has a real cost of its own. A
guard needs an answer, plus an allowlist shape for whatever is deliberately unlisted, and
neither is derivable from the current code: the nine drifted lines are the residue of an
absent rule, not the expression of one.

Two secondary questions the fix should settle while it is there. `update-deps` still
carries flags (`run --emit-updates <path>`) where every other multi-subcommand line
enumerates instead, which is the sub-defect `#1042` and `#1164` each fixed on their own
line. And whether the two binaries' shared namespace lines should be generated from one
source rather than hand-maintained twice, since four namespaces are currently duplicated
between them and drift identically.

## Scope note

Surfaced while verifying `#1164`'s premise before authoring its fix, by extracting every
router's `SUBCOMMAND_HANDLERS` keys and diffing them against both summaries. `#1164`'s own
claim is correct and its fix is unaffected; only its framing as a single-line defect was
incomplete. Not folded into that pull request: `#1164` is a one-line correction with a
settled answer, and this needs a design decision first, so batching them would hold the
mechanical fix behind the larger question.

This is the same shape as `#1163` (80 CLI module docblocks below the first import, also
surfaced by checking a narrower issue's premise, also mechanical-sweep-plus-guard, also
graded `hard` for the convention question rather than the edit).

Handler: plan

stevensacks:
Closes #1166

## What

Nine namespace lines in the two CLI top-level help summaries named fewer subcommands than their routers dispatch, 21 omissions in total. This adds a guard that pins each line to its router's `SUBCOMMAND_HANDLERS` keys, then sweeps both summaries, then rebundles the two committed binaries so the fix reaches an adopter.

| Line | Was | Now | Added |
|---|---|---|---|
| `wiki` (both binaries) | 10 | 14 | `chain`, `diff-size`, `empty-sections`, `frontmatter` |
| `automation` (both) | 2 | 4 | `install-audit-workflow`, `render-workflows` |
| `update-deps` (both) | 1 | 2 | `decline` |
| `init` (both) | 6 | 8 | `bootstrap-env`, `configure-automation` |
| `setup-ci` (adopter) | 10 | 13 | `check-audit-drift`, `check-drift`, `write-isolation-policy` |

Nothing dispatched incorrectly and every namespace's own `--help` was already complete; only the two top-level summaries had rotted. Subcommand order now mirrors each namespace's own `HELP_TEXT`, so the two surfaces correspond line for line.

## The design decision

The issue asks whether a summary line must be exhaustive, and says the code does not settle it: `sandbox seed` sits in the reachability guard's `INTERNAL_COMMANDS` as invoker-less by design, yet it is listed, so no "user-facing only" rule was in force.

Settled as **exhaustive enumeration with no allowlist**. A curated list plus a reasoned allowlist was declined because it adds a second hand-maintained list that itself needs curating; dropping subcommands from the summary entirely in favour of each namespace's own `--help` was declined because it costs one-screen discovery, and the reported defect is that the summary omits things, not that it should not exist. If a command ever genuinely needs to be unlisted, the allowlist gets designed then.

Two secondary questions the issue raises, settled from the code:

- `update-deps` drops its `--emit-updates <path>` flag text and becomes `run|decline`. Every other multi-subcommand line enumerates rather than showing flags; `update-deps` was the sole exception, and only because its second subcommand was missing. `gaia update-deps --help` still documents the flags.
- The two binaries stay hand-maintained rather than generated from one source. `index.maintainer.ts`'s own docblock says the mirroring is deliberate and that a shared abstraction would obscure the contract more than it saves; the guard is what makes that duplication safe.

## The guard

Added to `command-reachability.test.ts`, which already enumerated every router's keys, and sharing that one scan. Three assertions:

- each namespace line's subcommand list **set-equals** its router's keys, in both directions
- every dispatched command has a summary line
- every summary line is for something the binary dispatches

Scope boundary, stated in the docstring: only lines whose first token is a `SUBCOMMAND_HANDLERS` domain router are compared key-for-key. `ci-revert`, `harden-ledger` and `scaffold` declare their verbs inside a single handler or via `if (subcommand === '...')`, so for those the guard checks only that the command is documented and dispatched.

## Verification

The guard was committed red before the fix (`4304fc55`), failing on exactly the 9 lines and 21 omissions the issue reports, which is an independent re-derivation of the premise rather than a restatement of it.

Six mutations, each firing its own assertion, **re-run in full twice** after the Simplify pass and the lint fixes rewrote the parser:

| Mutation | Fires |
|---|---|
| line names a subcommand the router lacks | `drifted` |
| line omits a subcommand the router has | `drifted` |
| dispatched command with no summary line | `unlisted` |
| summary line for something not dispatched | `undispatched` |
| line pattern indent 2 → 3 (matches nothing) | the anti-vacuity pin |
| subcommand capture neutered to always-empty | `drifted` |

That fifth one is why the anti-vacuity pin exists: every assertion in the block is a set difference, and an empty set difference is what passing looks like, so a parser that matched nothing would green all three.

Quality Gate green as written: typecheck (root and `.gaia/cli`), lint, `lint:cli`, `pnpm test --run`, the CLI's own 1976 tests, Playwright, dev smoke, build. `shell-lint.sh` passes and `release manifest --check` is clean.

## Notes for the reviewer

- `pnpm lint:cli` failed on the first run with 8 errors, all in the new code, fixed in `01888a7c`. One of them was substantive rather than stylistic: with the subcommand capture written as `(...)?` the group types as `string` while being `undefined` at runtime, and the optional chain covering that gap read as an unnecessary condition. The optionality now sits inside the capture so the group always participates.
- `.gaia/cli/gaia` and `.gaia/cli/gaia-maintainer` are committed build artifacts, so the source sweep does not reach an adopter until they are rebuilt. The bundle diff is confined to the nine help lines; the template copies that step also refreshes came back byte-identical.

## Audit rounds

**Round 1** clean: 0 Critical, 0 Important, 1 Suggestion, 0 cross-remit. The member re-derived the guard independently in a scratch harness and ran seven mutations through it, including three I had not (dropping the first, middle and last subcommand of a line separately, and a stray two-space-indented prose line). All caught.

Its Suggestion was **fixed rather than accepted**, against the digest-economics default, because it falsified a claim this PR makes about itself: `HANDLER_KEY_PATTERN` matches only a `run<Pascal>` value, so an entry written any other way is dropped from the key set silently, and a subcommand missing from *both* its router's keys and its summary line reads as agreement. The block's own anti-vacuity test is named "guards against parser rot", so a parser-rot path it did not cover made that name false.

It was **proved before it was fixed**: with `chain: handleChain` in the wiki router and `chain` dropped from both summaries, the entire suite passed green. The repair (`d3996521`) is not the one suggested. Per-namespace key-count pins would add a second hand-maintained list of numbers; requiring every map-body line to be an entry, a comment, or blank needs no maintenance and covers both guards in the file. Widening the pattern was rejected on the ground it already fails on: the next unanticipated value shape fails the same silent way.

Three proofs after the fix: the reproduced gap now fails loudly; **neutering the new assertion brings the silent green back** (which is what shows it is load-bearing rather than merely present); and an inline-arrow entry is caught too, so the repair is not fitted to the one shape reported.

**Round 2** clean: 0 Critical, 0 Important, 1 Suggestion, 0 cross-remit. The member independently mutated seven more shapes through the new logic (two-line entries, non-conforming key shapes, a numeric-leading quoted key, prose inside a block comment) and confirmed the deliberately-quiet case, an entry commented out, stays quiet.

## Accepted residual (recorded, not fixed)

Round 2's Suggestion lands on round 1's own repair, which is the tell this repo's merge workflow names, and the stopping rule for round 2 was written into the step's checkpoint before the round ran. Accepting it under that rule:

- `.gaia/cli/src/command-reachability.test.ts:146` — `handlerMapBody`'s `endOffset === -1` fallback returns the rest of the file, so a `SUBCOMMAND_HANDLERS` map that does not close with a column-0 `};` reports every following non-comment line as unparsed. Rewriting the terminator to `} as const;` took the count from 0 to 19. **The failure stays loud, which is the property that matters**; only the diagnosis is noisier than it needs to be. No live failure: every map in the tree closes conventionally, and the reporting member graded it advisory and reasonable to decline.

Nothing was waived and no tech-debt was filed: both rounds returned zero cross-remit findings.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.RoJOFYyLGD

PR #1223, opened 2026-08-05
debt(audit): make every membership diff quote-safe (#1213)

### LINKED ISSUE #1213: A C-quoted path defeats a specialist audit member's self-skip, so the file merges unaudited
## A C-quoted path defeats a specialist member's self-skip, so the file merges unaudited

`.claude/agents/code-audit-github-workflows.md:52`

The same defect sits at `.claude/agents/code-audit-maintainer-node.md:55`, `.claude/agents/code-audit-maintainer-prose.md:49`, and `.claude/agents/code-audit-maintainer-shell.md:65`. All four carry the identical derivation:

```bash
full_changed=$(git -C "$AUDIT_ROOT" diff --name-only "${FULL_BASE}...HEAD" 2>/dev/null || true)
```

### Failure mode

`git diff --name-only` C-quotes any path containing non-ASCII or control bytes under git's default `core.quotePath=true`, emitting the surrounding double quotes as literal characters in the output. A tracked file named `été.ts` comes back as the 20-character token `"\303\251t\303\251.ts"`, not as `été.ts`.

`full_changed` decides **whether the member runs at all**: each specialist filters it against its own remit globs and self-skips when nothing matches. A C-quoted token matches no glob, so a pull request whose only in-remit change is a non-ASCII-named file produces an empty filtered set, the member self-skips cleanly, writes no marker, and `resolve-audit-spawn.sh` never names it. The merge then proceeds with nobody having reviewed that file. The self-skip is the fail-open: it is indistinguishable from a genuine no-match, and the member reports a clean skip rather than an error.

Verified in a scratch repository against the review-scope sibling of this exact derivation: the quoted token reached `git status --porcelain --` as a pathspec, matched nothing, and the pipeline returned empty at exit 0 while an unscoped `git status --porcelain` printed ` M "\303\251t\303\251.ts"` for the same file.

### Why this is filed rather than fixed in place

PR #1212 fixes this class on the **review-scope** derivation (`changed=…`) in all five members, because that derivation feeds the dirty-tree check that pull request adds. This issue is the **membership** derivation (`full_changed=…`), a different variable feeding a different decision with a different consequence: a missed dirty-tree check versus an entirely unaudited merge. It is pre-existing on all four specialists and outside that pull request's subject.

### Suggested fix

Add `-c core.quotePath=false` to the four specialists' `full_changed` derivation, matching what PR #1212 landed on the review-scope derivation beside it:

```bash
full_changed=$(git -C "$AUDIT_ROOT" -c core.quotePath=false diff --name-only "${FULL_BASE}...HEAD" 2>/dev/null || true)
```

The default member already avoids the defect by a different route at `.claude/agents/code-audit-frontend.md:739`, using `diff --name-only -z … | tr '\0' '\n'`, since `-z` disables quoting. Either route is correct; the flag is the one the adjacent `changed` derivation now uses, so it is what keeps the two lines in one member reading the same way. Whichever is chosen should be applied to all four, and `.gaia/scripts/check-audit-base-derivation.sh` is the natural place to pin it if the parity is worth enforcing.

### Reachability

Low. It needs a repository with a tracked path carrying non-ASCII or control bytes, and this repository has none today. Filed because the self-skip is silent by construction: when it does fire, the member reports a clean skip and nothing anywhere records that a file went unreviewed.

Handler: plan

stevensacks:
Closes #1213

## What this fixes

Every derivation that answers "which Code Audit Team members does this pull request owe" read `git diff --name-only` with git's default `core.quotePath` left on. That C-quotes any path carrying non-ASCII or control bytes, emitting the surrounding double quotes as literal characters, and the resulting token matches no member's remit glob.

Two consequences, and they are not the same one:

- **The membership resolver is the silent failure.** `.gaia/scripts/resolve-audit-members.sh` names no owner for a quoted token, answers with an empty set, and `resolve-audit-spawn.sh` reads that as "nobody owns anything here" and falls through to its ownerless probe, which spawns the default member. The file *is* reviewed, just never by the specialist whose remit it falls in, and the merge completes looking audited.
- **A member's own self-skip is the loud one.** When membership did name the member, it derives `full_changed` itself, sees the quoted token, self-skips and writes no marker the gate still demands.

## Two corrections to the issue's analysis, both from re-reading the code

1. **The suggested fix rests on a false premise.** `#1213` proposes `-c core.quotePath=false` on the ground that it is "what PR #1212 landed on the review-scope derivation beside it". `core.quotePath` appears nowhere in `.claude/agents/`; #1212 landed `--name-only -z … | tr '\0' '\n'`, so the read-alike argument actually points at `-z`. Independently, `-z` is the stronger fix: `core.quotePath=false` only stops treating bytes at or above 0x80 as unusual, so a path containing a double quote, a backslash, or a control byte is still C-quoted, and the issue's own failure statement names "non-ASCII **or control bytes**".
2. **`resolve-audit-spawn.sh` was in scope for a different reason than the issue implies, and "the file merges unaudited" is an overclaim.** The issue reads the spawn script's silence as a consequence of the member self-skipping; it derives its own list independently. Its ownerless probe nonetheless fails *closed* on a quoted token (the token is not allowlisted, so the default member is spawned), so what quoting costs there is accuracy rather than coverage. Fixing only the four members would have closed the issue without retiring the consequence in its own title, which is why the resolvers are in this diff rather than filed.

## Scope

In, because each fails **open** on the membership question: the four specialists' `full_changed`, `resolve-audit-members.sh`, `resolve-audit-spawn.sh`.

Out, and recorded here so a later reader does not re-find them as misses: `.claude/hooks/pr-merge-audit-check.sh:485` and `:533` carry the same shape and fail **closed** — a quoted path is neither allowlisted nor classified out-of-scope, so the marker becomes mandatory and the self-modification is denied.

Folded in: `wiki/concepts/PR Merge Workflow.md`'s copy of the derivation, because that page states the orchestrator reuses the fork point "in the same spelling the audit machinery already computes", which this change would otherwise falsify.

## The parity pin

This defect exists only because #1212 fixed one of two sibling derivations in the same fence, in five files, with nothing to notice the other. `check-audit-base-derivation.sh` gains a fourth assertion requiring every base-consuming `diff --name-only` in the roster to carry `-z`. Assertions 3 and 4 share one candidate net and one per-call window walk, parameterized by the token each requires, rather than opening a second copy of a drift-prone derivation inside the file whose subject is exactly that.

Its limits are stated in its own docblock rather than left implicit: it cannot require the NUL-to-newline conversion (the net cannot tell a fence from prose, and one member legitimately states the command inside a code span with no pipeline), and it reports `git diff -z --name-only`, which is correct but puts the flag outside the call's own window.

## Verification

- **Six adversarial mutations, all correct.** Dropping `-z` from one member reds the static check (assertion 3 staying green, so the two are independent), the real-repo bats pin, and the behavioural probe. Dropping it from the membership resolver reds that resolver's suite. Neutering assertion 4's token test reds its fixtures. Neutering the new real-repo `-z` loop assertion takes that one test green under the same defect, which is what proves the loop load-bearing rather than merely present.
- **The behavioural probe carries its own non-vacuity.** It commits one non-ASCII in-remit path per specialist, asserts each member's whole-PR list comes back raw and resolves to its owner through the real classifier, then re-asks the same question with the pre-fix spelling, which must quote and must leave every specialist ownerless. A fixture whose paths turned out to be plain ASCII cannot pass it.
- **Every existing assertion-3 fixture gained `-z`**, so each fixture carries exactly one defect and a red names one assertion.
- 308 bats tests across every touched path, 0 failures, run under bash 5. `bash .gaia/tests/shell-lint.sh` passes. The Quality Gate skips by its own documented rule (no source or gate-affecting config in the diff), verified with the page's own grep rather than from memory.

## Note for reviewers

This changes the audit agent definitions, so any clearance already earned on an open pull request is invalidated; re-run its audit members before merging.

---

## Audit gate: five rounds, four clean, one withheld

`code-audit-maintainer-shell` was the whole resolved member set every round, and it is the right one: it owns `.claude/agents/**`, `.gaia/**/*.sh` and `.gaia/**/*.bats`, while the two `wiki/` pages and `CHANGELOG.md` classify to no member.

| Round | Result | What it found |
|---|---|---|
| 1 | clean | 4 Suggestions. Two repaired here, two filed (below) |
| 2 | clean | 2 Suggestions, both repaired |
| 3 | **withheld** | 1 Important, repaired |
| 4 | clean | 3 Suggestions, all repaired |
| 5 | clean | 1 Suggestion, accepted and noted (below) |

Round 3's Important is the one worth reading. Round 2 changed assertion 4's token in **two** ways at once, adding a leading space and an anchor, and the fixture written for it rejected on the space alone, so dropping the anchor left the whole suite green. The mechanism shipped unproven, and the mutation meant to prove it had reverted both changes together, which is precisely the trap `wiki/concepts/PR Merge Workflow.md` names: *a change that adds two of them needs both mutated separately or the second one ships unproven.* Each mechanism now has its own fixture and its own mutant, and each reds a distinct set.

## Out-of-scope findings, filed rather than fixed

Both were on paths this pull request does not change, so no waive was available and filing is the disposition the cross-remit rule names. Both are the same C-quoting class this branch closes, on derivations that answer different questions:

- **#1224** `.github/audit/resolve-audit-base.sh:240` — the RT-006 machinery base reset silently does not fire on a C-quoted machinery path.
- **#1225** `.claude/hooks/worthiness-presence-check.sh:164` — the worthiness presence gate passes on exactly the input it exists to hold.

`.claude/hooks/pr-merge-audit-check.sh:485` and `:533` carry the same shape and were checked rather than assumed: both fail **closed**, so neither is a defect and neither was filed.

## Accepted residual, recorded not fixed

Round 5's single Suggestion, under a stopping rule written into this step's checkpoint before that round ran, which routes any prose/comment/count/test-hygiene Suggestion to accept-and-note unconditionally.

`.gaia/scripts/tests/check-audit-base-derivation.bats:719` — the mutation-coverage comment maps fixtures to mutants **correctly**, and the mapping is the part a later editor would act on, but two of its three bullets state the outcome polarity inverted. `.gaia/scripts/tests/check-audit-base-derivation.bats:774` carries the same inversion. The corrections, in full, for whoever next touches the file:

- bullet 1: "the only test that goes green when the anchor alone is dropped" → **reds**
- bullet 3: "survives both single mutants and turns green only on the both-absent pre-fix spelling" → **reds** only on the both-absent spelling
- `:774`: "this is the test that goes green while the pathspec fixture beside it stays red" → that test **reds** while the pathspec fixture stays **green**

Measured, not inferred: anchor dropped reds test 33 alone; leading space dropped reds 17, 22, 23, 30, 36, 39; both dropped reds 32 and 33.

Not fixed because a sixth member re-dispatch for two words is the runaway the rule exists to stop, and not filed because it is a comment in a maintainer-only test file with no adopter surface, which the scope gate would mark `maintenance` on arrival.

## One operational note

Every member dispatched in this session had its self-skip snippet rendered from the **pre-fix** copy of its own definition, so this branch's members were running the very derivation it repairs. It changed nothing here (no non-ASCII path in the diff), and the tracked files at HEAD are correct. Noted because the same thing happened on #1194.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.keaMiO0Kkr

PR #1227, opened 2026-08-05
debt(audit): make the RT-006 reset and the worthiness gate quote-safe (#1224, #1225)

### LINKED ISSUE #1224: A C-quoted machinery path stops the RT-006 base reset from firing
## A C-quoted machinery path stops the RT-006 base reset from firing

`.github/audit/resolve-audit-base.sh:240`

```bash
_delta="$(git -C "$repo_root" diff --name-only "$candidate" "$head_sha" 2>/dev/null || true)"
```

### Failure mode

Under git's default `core.quotePath`, `git diff --name-only` C-quotes any path carrying non-ASCII or control bytes, wrapping it in double quotes and backslash-escaping the offending bytes, and emits those quotes as literal characters.

`_delta` feeds `audit_delta_has_machinery`, and `audit-machinery.sh` matches `.claude/rules/**`, `.claude/hooks/lib/**` and `.github/audit/**` as literal prefixes. A quoted token begins with `"`, so it prefix-matches nothing.

The consequence is that the machinery base reset **silently does not fire**. That reset exists to widen every Code Audit Team member back to full scope when gate machinery moved between the cleared commit and HEAD; when it does not fire, the members stay on the incremental base and review only the delta since the last clean round, so a machinery change carried by a non-ASCII path is never re-reviewed at full scope. Nothing reports that the reset was skipped, because a machinery-free delta is the ordinary case and looks identical.

Verified empirically during the `#1213` audit round: `audit_delta_has_machinery` returns detected for `.claude/rules/regle.md` and not-detected for the C-quoted form of the same path spelled with an accent.

### Suggested fix

```bash
_delta="$(git -C "$repo_root" diff --name-only -z "$candidate" "$head_sha" 2>/dev/null | tr '\0' '\n' || true)"
```

This is the spelling PR #1223 standardized across the membership derivations and the one the whole Code Audit Team roster already uses. `-z` rather than `-c core.quotePath=false`, because the flag only stops treating bytes at or above 0x80 as unusual and still quotes a path containing a double quote, a backslash, or a control byte. Leave the here-string feed below and its documented SIGPIPE reasoning untouched.

`.gaia/scripts/check-audit-base-derivation.sh` assertion 4 pins this spelling, but its scan is `.claude/agents/code-audit-*.md`, so it does not reach this file.

### Reachability

Low. It needs a tracked path carrying non-ASCII or control bytes, and this repository has none today. Filed rather than left unrecorded because the skipped reset is silent by construction: when it does fire, nothing anywhere records that a machinery change went re-reviewed only at incremental scope.

### Provenance

Found by the `code-audit-maintainer-shell` round-1 audit on PR #1223 (`#1213`), which closes the same defect class on the derivations that decide audit membership. This one is a different consumer (scope reset) on a path that pull request does not change, so no waive was available and filing is the disposition the merge workflow's cross-remit rule names.

Handler: prompt

### LINKED ISSUE #1225: A C-quoted path makes the worthiness presence gate pass the test it exists to hold
## A C-quoted path makes the worthiness presence gate pass the test it exists to hold

`.claude/hooks/worthiness-presence-check.sh:164`

```bash
changed=$(git diff --name-only "${base}...HEAD" 2>/dev/null || true)
```

### Failure mode

Under git's default `core.quotePath`, `git diff --name-only` C-quotes any path carrying non-ASCII or control bytes, emitting the surrounding double quotes as literal characters.

The quoted token is handed to `classify_emergent`, which classifies a path that does not exist, returns nothing, and never lands in `offenders`. The gate then reports no offenders and **passes**, on exactly the input it exists to catch.

This is the fail-open half of a split worth stating, because the sibling shape is not a defect: `.claude/hooks/pr-merge-audit-check.sh:485` and `:533` read the same quoting list and fail **closed**, since a quoted path is neither allowlisted nor classified out-of-scope, so the marker becomes mandatory and the self-modification is denied. Only this consumer turns the same bad input into a pass.

### Suggested fix

```bash
changed=$(git diff --name-only -z "${base}...HEAD" 2>/dev/null | tr '\0' '\n' || true)
```

The spelling PR #1223 standardized across the membership derivations. `-z` rather than `-c core.quotePath=false`, because the flag only stops treating bytes at or above 0x80 as unusual and still quotes a path containing a double quote, a backslash, or a control byte.

### Reachability

Low. It needs a tracked path carrying non-ASCII or control bytes, and this repository has none today. Filed rather than left unrecorded because a gate that passes on the input it exists to hold reports nothing at all: the failure is indistinguishable from having nothing to flag.

### Provenance

Found by the `code-audit-maintainer-shell` round-1 audit on PR #1223 (`#1213`), which closes the same defect class on the derivations that decide audit membership. This one is a different gate on a path that pull request does not change, so no waive was available and filing is the disposition the merge workflow's cross-remit rule names.

Handler: prompt

stevensacks:
Closes #1224
Closes #1225

## What

Two consumers of `git diff --name-only` read the file names git prints and match them literally, and both left git's default `core.quotePath` on. A path carrying non-ASCII or control bytes comes back C-quoted (wrapped in literal double quotes, offending bytes backslash-escaped), so it matches nothing either consumer is looking for, and **each turns the input it exists to catch into a pass**:

- **`.github/audit/resolve-audit-base.sh:240`** — the quoted token prefix-matches no machinery path, so RT-006's base reset does not fire. The round then reviews only `<candidate>..HEAD` under machinery rules that moved underneath it, and nothing anywhere records the skip, because a machinery-free delta is the ordinary case.
- **`.claude/hooks/worthiness-presence-check.sh:164`** — the quoted token matches none of the emergent-surface case patterns, drops out of the loop, and the gate reports no offenders and passes.

Both take the `--name-only -z … | tr '\0' '\n'` spelling that PR #1223 standardized across the membership derivations and that the whole audit roster already uses. `-z` rather than `-c core.quotePath=false`, because the flag only stops treating bytes at or above 0x80 as unusual and still quotes a path containing a double quote, a backslash, or a control byte.

Neither script's other behaviour is touched. In particular `resolve-audit-base.sh` keeps its here-string feed to `audit_delta_has_machinery` and the documented SIGPIPE reasoning behind it: `tr` drains stdin fully, so the new pipe cannot reintroduce that failure.

## Oracle

One test per member, each in that script's own existing suite, each committed red before the fix:

- `.github/audit/tests/resolve-audit-base.bats` case 19 — a machinery commit on `.claude/rules/règle.md` between the version-matching base and HEAD must reset to full scope. Pre-fix it emitted the candidate.
- `.gaia/tests/hooks/worthiness-presence-check.bats` — an emergent component test at `app/components/Café/tests/index.test.tsx` with no ledger line must be denied. Pre-fix it was allowed.

**Guard proofs, one mutant per mechanism.** Reverting `resolve-audit-base.sh` alone reds the RT-006 test and leaves the worthiness test green; reverting `worthiness-presence-check.sh` alone does the mirror image. Each guard fires on its own mechanism and neither vouches for the other.

## Verification

- `.github/audit/tests/` — 165 ok / 0 not ok
- `.gaia/tests/hooks/` — 1290 ok / 0 not ok
- `.gaia/scripts/tests/{audit-base-agreement,check-audit-base-derivation,resolve-audit-members,resolve-audit-spawn,audit-machinery-complete}.bats` — 163 ok / 0 not ok
- Quality Gate skips by its own rule (`wiki/decisions/Quality Gate.md`, **When to skip the gate**): the four changed files are two `.sh` and two `.bats`, none of which typecheck / lint / tests / build inspect.

## Scope notes

- **A single static sweep was considered and declined.** A check requiring `-z` on every `diff --name-only` in the audit shell would also flag `.claude/hooks/pr-merge-audit-check.sh:485` and `:533`, which S32 examined and found are **not** defects: a quoted path there is neither allowlisted nor classified out-of-scope, so both fail closed. Greening such a check would mean either editing those two sites, outside these two issues, or building an exemption mechanism.
- **`check-audit-base-derivation.sh` assertion 4 is not extended.** Its `consumes` net keys on `resolve-audit-base.sh` / `BASE_REF` / `BASE_SHA` / `FULL_BASE` appearing inside the call window, and neither call here names any of them, so widening its scan to shell would not reach them without rewriting the net. Both issue bodies note the scan does not reach these files; that stands as an accurate observation.
- This PR changes gate machinery, so RT-006 resets the review base to full scope for every audit round on it. That is the correct behaviour, not a fault.

---

## Audit round 1: clean, with dispositions

`code-audit-maintainer-shell` was the whole dispatched set (`resolve-audit-spawn.sh`, exit 0) and it cleared: no Critical, no in-scope Important, marker earned. It reproduced both guards failing against the base scripts in its own scratch sandbox before accepting them.

### Accepted and noted, not fixed

Both new comments describe the derivation as quoting a path carrying "non-ASCII or control bytes". A newline is a control byte, and `tr '\0' '\n'` re-splits on it, so a newline-bearing path still arrives split rather than whole. The precise correction, for a later editor:

- `.github/audit/resolve-audit-base.sh` and `.claude/hooks/worthiness-presence-check.sh` — the comments describe the **defect** (git C-quotes such paths) rather than claiming the fix's coverage, and what `-z … | tr` actually closes is the quoting, not every control byte.
- There is no live failure mode on either side: in the hook, a split fragment fails the `[ -f "$rel" ]` guard and continues; in `resolve-audit-base.sh`, a split fragment can only cause a spurious full-scope reset, which reviews more rather than less.
- `.gaia/scripts/resolve-audit-members.sh` carries the same wording, so the phrasing is repo-consistent rather than newly introduced here.

Accepted rather than fixed on the digest economics in `wiki/concepts/PR Merge Workflow.md` (*Applying the audit's own Suggestions*): this PR is clean and the member is already marked, so editing either script buys a full re-dispatch to re-earn the marker, for two words in a comment.

### Filed, not waived

Three out-of-scope findings. No waive was available for any of them: none of the three paths is gate machinery by `audit_path_is_machinery`, and this PR changes none of them, so neither cross-remit eligibility term holds.

- **#1228** `.claude/hooks/distribution-preflight-check.sh:270` — the same defect class in a third gate: a C-quoted path never intersects the manifest checker's raw `missing` set, so the hook allows `gh pr create` for a newly-shipping file with no ship-or-withhold answer.
- **#1230** `.github/workflows/distribution-audit-pr.yml:118` — the CI half of the same defect, and the backstop #1228's own comment claims. Filed separately because the filing recipe keys dedup on one `path=`+`line=` per body, and because its Code Audit Team owner is `code-audit-github-workflows`.
- **#1229** `.gaia/scripts/check-audit-base-derivation.sh:557` — assertion 4's candidate net is scoped to `.claude/agents/code-audit-*.md`, so no deterministic check has ever been able to see a shell call site. The class has now been found and fixed four times in four files, each time by a human or a member reading the code. Filed rather than folded in because the fix is a real design fork: widening the net requires redefining its `consumes` predicate for shell, and a blanket shell lint reds `pr-merge-audit-check.sh:485`/`:533`, which fail closed and are not defects.

All three are `severity:suggestion` to match #1224/#1225, the identical class at identical reachability; the member's own **Important** grading of #1228 is recorded in that issue so a triager can overturn the call rather than rediscover it.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.rP4sDxq0yj

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.pfxTD81q3C

PR #1231, opened 2026-08-05
debt(tests): make three assertions check what their names claim (#1201, #1209, #1214)

### LINKED ISSUE #1201: Byte-identity assertions built on $(cat …) strip trailing newlines, so a trailing-newline regression greens the tests named for byte identity

### LINKED ISSUE #1209: The lint-pin-parity containment assertions compare exemption entries literally, so pnpm's own merged union spelling reds a valid asymmetric pair
`.gaia/cli/src/lint-pin-parity.test.ts:307` and `.gaia/cli/src/lint-pin-parity.test.ts:376`

## Failure mode

The two containment assertions compare exemption entries as **literal strings**:

```ts
expect(rootSettings.minimumReleaseAgeExclude).toEqual(
  expect.arrayContaining(cliSettings.minimumReleaseAgeExclude)
);
```

pnpm writes its own exemption entries into `pnpm-workspace.yaml`, and when a package needs more than one exempted version it merges them through `mergePackageVersionSpecs` into a single union entry: `semver@6.3.1 || 6.3.2`.

Root's dependency closure is a superset of `.gaia/cli`'s, so the two workspaces can legitimately need different version counts of the same shared package. Once that happens, root carries `semver@6.3.1 || 6.3.2` while `.gaia/cli` carries `semver@6.3.1`. That is a genuine subset relationship, and it is exactly what the assertion is written to permit, but literal string membership does not see it: `arrayContaining` looks for the exact string `semver@6.3.1` in root's list, finds only the merged union, and throws.

The result is a **red required check on a valid configuration** — the same class of false positive that was fixed one test over (in the exactness predicate, which was taught pnpm's union spelling), left unfixed in the containment assertions beside it.

Verified rather than argued: the pair was run through the repo's own `expect`, both sides return `[]` from `inexactEntries` (so the exactness tests pass), the `.gaia/cli` exemption set is a genuine subset, and the containment assertion throws anyway.

## Reachability, stated honestly

This does not fire today: every entry shared by the two files is currently byte-identical. Reaching it needs pnpm to write a merged union to one workspace and not the other.

That write path is real but gated. Because both workspace files set `minimumReleaseAge` explicitly, pnpm coerces `minimumReleaseAgeStrict` to `true` (`pnpm@11.9.0/dist/pnpm.mjs:146622-146623`), so an immature resolution **fails the install** under CI (non-TTY) and **prompts for approval** on a TTY. The union entry is therefore written only after a maintainer approves at that prompt, not silently on an ordinary install.

When it does fire it is a loud, self-describing red naming the entries, never a silent pass.

## Suggested fix

Expand both sides into `name` / `name@version` atoms before comparing, so the union spelling is normalized away on both sides rather than compared literally. pnpm ships this as `expandPackageVersionSpecs`, the inverse of the `mergePackageVersionSpecs` that produces the union, so the shape is the vendor's rather than invented — prefer reusing pnpm's own spelling over hand-rolling a split on `||`, which is how the sibling predicate accumulated four rounds of shape-by-shape repair.

Apply it to both assertions (`:307` and `:376`); they have the same defect.

Handler: prompt

### LINKED ISSUE #1214: Two load-bearing lines of the dirty-tree posture are unpinned, so its suite greens while the mechanism is gone

stevensacks:
Three assertions that did not check what their names claimed. Each is proved by mutating the mechanism it guards and watching the suite go red.

Closes #1201
Closes #1209
Closes #1214

## `#1214` — two load-bearing lines of the dirty-tree posture were unpinned

`SENTINEL_CARVEOUT` pinned the prose promising the fail-closed sentinel is never remit-filtered, and `CHECK_LINE` pinned only the `if`. The assignment that actually produces the sentinel was pinned by neither, so deleting `dirty_in_scope="dirty-scope check failed"` from all five members left the suite reporting 18/18 ok while the failure arm produced nothing to withhold on. It now carries its own pin across every member.

The advisory-drift guard compared one bolded sentence literally, so it only ever caught the byte-identical copy-paste. It now scans for withhold-shaped clauses by meaning (`withhold(s|ing)? (this|your|the) (pass|clearance|marker)`, case-insensitive), excluding only the exemption's own negation and anchoring that negator to the verb. Two spellings are deliberately out of the pattern because the advisory member carries them legitimately: `write no marker` is its self-skip arm, and an unqualified `withhold` covers its own exemption prose. Newlines are folded before matching, since `grep` is line-scoped. The fixtures are derived from the gating members' real phrasings rather than invented — see the audit-gate section below for why that inversion is the point.

**Proof.** Deleting the sentinel assignment from all five members reds test 6. Inserting a *reworded* withhold clause into the advisory member, exemption paragraph intact, reds test 12 — and the old exact-string guard was confirmed green on that same mutant.

## `#1209` — a valid asymmetric exemption pair red a required check

pnpm merges multiple exempted versions of one package into a single union entry, `name@1.0.0 || 2.0.0`, and writes that back itself. Root's dependency closure is a superset of `.gaia/cli`'s, so the two workspaces can legitimately need different version counts of a shared package: root carries the union, `.gaia/cli` carries the single version. Literal `arrayContaining` membership cannot see that as containment.

Both assertions now expand each side into the `name` / `name@version` atoms pnpm's own `expandPackageVersionSpecs` produces, following its scope-aware parse. Mirrored rather than imported: pnpm is this repo's `packageManager`, not a dependency of either workspace. One divergence is written into the comment — no `semver.valid` folding of `v1.2.3` — rather than left implicit.

**Proof.** Root carrying `semver@6.3.1 || 6.3.2` against `.gaia/cli`'s `semver@6.3.1` reds the pre-fix test file and passes after. It stays catchable both ways: `.gaia/cli` exempting a version root does not, and `.gaia/cli` exempting *all* versions where root exempts one, each still red.

## `#1201` — byte-identity assertions built on `$(cat …)`

Command substitution strips all trailing newlines from both sides, so a regression that only adds or drops one compared equal in the very tests whose names promise byte identity.

The issue's regex returns 87 candidates; they are candidates, not instances. Sweeping for comparisons where **both** sides are file contents yields 49, in two shapes the regex only half-covers (`[ "$(cat f)" = "$before" ]` and `[ "$before" = "$after" ]`). Two of those 49 are not the class and are untouched: `plan-allocator.bats:151` and `spec-allocator-remote.bats:98` compare allocator IDs a script *printed* into a file, where the strip is what you want. The remaining **47** now compare files, through a shared helper rather than hand-rolled `cmp` per call site, so the next byte-identity test written reaches for a primitive instead of `$(cat …)` out of habit.

`.gaia/tests/helpers/files.sh` is new and sourced from all three bats directories. It gets its own entry in `audit-ci-tests.yml`'s `code:` paths filter: it sits at the top of `.gaia/tests/`, matching none of the `.gaia/tests/*/` globs, so without it an edit weakening the helper alone would report `code=false`, skip every bats step, and green the job.

**Proof.** A real writer regression planted in `.gaia/scripts/cost-reprice.sh` (append a trailing newline on the no-op arm, the exact path those tests call "left byte-identical") reds **six** tests on the new assertions and **zero** on the old ones.

## Verification

Quality Gate run in full, all green: typecheck, lint, `vitest`, and `build` at the root; typecheck, lint and 1980 tests in `.gaia/cli`; Playwright (8 passed, 1 skipped) and the dev smoke test. `shell-lint.sh` passes over 195 scripts and 189 suites. Every mutation above was run after committing, so `git checkout --` could take nothing else with it, and every one was reverted to a clean tree.

**No CHANGELOG entry.** All 15 changed files were checked against `.gaia/manifest.json` and none is in it, so nothing here ships; all three issues are `surface:maintainer`.

## Accepted residuals (audit round 2, `code-audit-maintainer-node`)

Both are Suggestions on an already-marked member whose digest nothing else rotates, so folding
either in buys a full re-dispatch for a comment. Recorded here instead, per the digest economics in
`wiki/concepts/PR Merge Workflow.md`.

- **A name-only entry is not contained by a versioned one, in either direction.** Root exempting
  every version of a package (`semver`) while `.gaia/cli` exempts one (`semver@6.3.1`) reds the
  containment assertion even though root is strictly broader. Real, but **inert today** (the only
  name-only entry, `@gaia-react/lint`, is name-only on both sides), **unchanged from the pre-diff
  literal comparison**, and faithful to pnpm's own `expandPackageVersionSpecs`, which likewise
  yields a single atom for a name-only spec. Changing it would diverge from the vendor parse this
  helper deliberately mirrors.
- **Two of the four `exemptionAtoms` tests stay green under an identity mutation.** True: only the
  merged-union and scoped-union cases red when the helper is replaced by `list => list`. The other
  two are kept deliberately, because they pin a different property, that expansion does not make
  containment *vacuous* in the direction that would hide real drift.

## Audit gate

Three rounds. Round 1 dispatched all three resolved members; `code-audit-github-workflows` cleared
with zero findings and has held a valid marker since. Rounds 2 and 3 re-dispatched only the members
whose content digests had rotated.

Every finding raised against this PR was **fixed**, not accepted, except the two residuals above.
The two Important findings were both real defects in this PR's own new guards, and both are worth
naming because they are the failure this batch exists to fix, reappearing one level up:

1. The drift guard's negation exclusion matched `does not` / `never` / `cannot` anywhere in the
   context window rather than against the verb. Since `never` alone appears 33 times in the guarded
   file, two house-style withhold clauses passed it. Fixed by anchoring the negator to the verb.
2. The drift pattern required the determiner `this` or `your`, so it missed `withhold **the**
   marker`, which is the verbatim handshake sentence three gating members carry, i.e. the likeliest
   copy-paste vector of all. Fixed by admitting `the`, and, more importantly, by **deriving the
   fixture set from the gating members themselves** so a phrasing the scan cannot reach reds when a
   sibling adopts it rather than after it is pasted.

The second repair is deliberately not a third alternation patch: two consecutive rounds finding the
regex one shape short is the enrichment loop, and the derived-fixture inversion is what ends it.

## Accepted residuals (audit round 3, `code-audit-maintainer-shell`)

Round 3 came back clean: no Critical, no Important, three Suggestions. Shell is marked and nothing
else rotates its digest, so folding any of these in buys a full re-dispatch of the roster's most
expensive member for comment-level improvements. Recorded here instead.

- **`gating_withhold_phrases` dedupes case-sensitively while extracting case-insensitively.** So
  `withhold the marker` and `Withhold the marker` both survive `sort -u`: the five extracted phrases
  are **four** distinct ones case-insensitively, and the `>= 4` anti-vacuity floor therefore carries
  one entry of slack. **The comment beside that floor saying "five distinct phrasings today" is
  inaccurate on this point** and is corrected here rather than in the file. The primary assertion is
  unaffected: every extracted phrase must still be caught, and all five are. The fix, when someone
  is next in that file, is `| tr '[:upper:]' '[:lower:]' | sort -u`.
- **The shared helper is sourced by four different path spellings.** Three lib suites use
  `$BATS_TEST_DIRNAME/../helpers/files.sh` on the line after they set `HELPERS` to a *different*
  directory one component away, and `cost-reprice.bats` uses a third spelling where its siblings use
  `$REPO_ROOT`. Consistency, not a defect; standardizing on `$REPO_ROOT/.gaia/tests/helpers/files.sh`
  needs `REPO_ROOT` derived in three suites that lack it.
- **A test name reads "every member assigns the fail-closed sentinel it withholds on"** while
  looping all five members, including the advisory one that never withholds. The loop bound is
  correct (that member does carry the assignment, and records rather than withholds on it); only the
  wording invites a later "fix" to `GATING` that would be wrong.

None is filed as tech debt. Each sits on a path this PR already changes, which is the waive term in
the merge workflow's cross-remit rule, and none has been observed to spread.

### DIFF
--- a/.gaia/cli/src/lint-pin-parity.test.ts
+++ b/.gaia/cli/src/lint-pin-parity.test.ts
@@ -198,6 +198,67 @@ const readHardeningSettings = (
   };
 };
 
+// Exemption entries compared as ATOMS rather than as literal strings, because
+// pnpm does not store one exemption per entry. When a package needs more than
+// one exempted version it merges them into a single union entry,
+// `name@1.0.0 || 2.0.0`, and writes that back into `pnpm-workspace.yaml`
+// itself. Root's dependency closure is a superset of `.gaia/cli`'s, so the two
+// workspaces can legitimately need different version counts of one shared
+// package, at which point root carries the union and `.gaia/cli` carries the
+// single version. That is exactly the containment the two tests below permit,
+// and literal membership cannot see it: `arrayContaining` looks for
+// `semver@6.3.1`, finds only the union, and reds a required check over a valid
+// configuration.
+//
+// The shape is pnpm's own (`expandPackageVersionSpecs`, the inverse of the
+// `mergePackageVersionSpecs` that writes the union), mirrored here rather than
+// imported: pnpm is this repo's `packageManager`, not a dependency of either
+// workspace, so there is nothing to import from. Its parse is followed exactly,
+// including the scope-aware `@` split that keeps `@scope/name` from splitting on
+// its own leading `@`.
+//
+// One deliberate divergence, stated because a wrong normalization here is the
+// same false-positive class this exists to fix: pnpm runs each version through
+// `semver.valid`, which also folds spellings like `=1.2.3` and `v1.2.3` onto
+// `1.2.3`. Trimming is all that pnpm's OWN output needs (its union separator
+// leaves a leading space), and `semver` is a dependency of neither workspace. A
+// hand-written `v1.2.3` on one side and `1.2.3` on the other would therefore
+// still red. Add the dependency and use `semver.valid` if that ever happens; do
+// not grow a second normalizer by hand, which is how the sibling exactness
+// predicate accumulated four rounds of shape-by-shape repair.
+//
+// A name-only entry stays one atom, so exempting EVERY version of a package in
+// `.gaia/cli` is not contained by exempting one version of it at root. That is
+// asymmetric hardening and stays red, which is pnpm's reading too.
+//
+// A non-string entry is the exactness tests' finding, not this one's. It passes
+// through unexpanded so it still compares, rather than being dropped into a
+// silently smaller list on the containing side.
+const exemptionAtoms = (list: unknown[]): unknown[] => {
+  const atoms = new Set<unknown>();
+
+  for (const entry of list) {
+    if (typeof entry === 'string') {
+      const atIndex =
+        entry.startsWith('@') ? entry.indexOf('@', 1) : entry.indexOf('@');
+
+      if (atIndex === -1) {
+        atoms.add(entry);
+      } else {
+        const packageName = entry.slice(0, atIndex);
+
+        for (const version of entry.slice(atIndex + 1).split('||')) {
+          atoms.add(`${packageName}@${version.trim()}`);
+        }
+      }
+    } else {
+      atoms.add(entry);
+    }
+  }
+
+  return [...atoms];
+};
+
 describe('@gaia-react/lint pin parity', () => {
   const repoRoot = resolveRepoRootFromImportMeta(import.meta.url);
   const rootPin = readPin(path.join(repoRoot, 'package.json'));
@@ -303,72 +364,14 @@ describe('supply-chain hardening parity', () => {
   // Containment, not equality. Root's lists are supersets by design, so equality
   // would fail on `bippy` and `chokidar`; but an entry present ONLY in `.gaia/cli`
   // exempts a package from the setting above it in one workspace and not the other,
-  // which is the asymmetric hardening this describe block exists to catch.
-  // Compared as ATOMS rather than as literal strings, because pnpm does not
-  // store one exemption per entry. When a package needs more than one exempted
-  // version it merges them into a single union entry, `name@1.0.0 || 2.0.0`, and
-  // writes that back into `pnpm-workspace.yaml` itself. Root's dependency closure
-  // is a superset of `.gaia/cli`'s, so the two workspaces can legitimately need
-  // different version counts of one shared package, at which point root carries
-  // the union and `.gaia/cli` carries the single version. That is exactly the
-  // containment these tests exist to permit, and literal membership cannot see
-  // it: `arrayContaining` looks for `semver@6.3.1`, finds only the union, and
-  // reds a required check over a valid configuration.
-  //
-  // The shape is pnpm's own (`expandPackageVersionSpecs`, the inverse of the
-  // `mergePackageVersionSpecs` that writes the union), mirrored here rather than
-  // imported: pnpm is this repo's `packageManager`, not a dependency of either
-  // workspace, so there is nothing to import from. Its parse is followed exactly,
-  // including the scope-aware `@` split that keeps `@scope/name` from splitting
-  // on its own leading `@`.
-  //
-  // One deliberate divergence, stated because a wrong normalization here is the
-  // same false-positive class this fixes: pnpm runs each version through
-  // `semver.valid`, which also folds spellings like `=1.2.3` and `v1.2.3` onto
-  // `1.2.3`. Trimming is all that pnpm's OWN output needs (its union separator
-  // leaves a leading space), and `semver` is a dependency of neither workspace.
-  // A hand-written `v1.2.3` on one side and `1.2.3` on the other would therefore
-  // still red. Add the dependency and use `semver.valid` if that ever happens;
-  // do not grow a second normalizer by hand, which is how the sibling predicate
-  // below accumulated four rounds of shape-by-shape repair.
-  //
-  // A name-only entry stays one atom, so exempting EVERY version of a package in
-  // `.gaia/cli` is not contained by exempting one version of it at root. That is
-  // asymmetric hardening and stays red, which is pnpm's reading too.
-  const exemptionAtoms = (list: unknown[]): unknown[] => {
-    const atoms = new Set<unknown>();
-
-    for (const entry of list) {
-      // A non-string is the exactness tests' finding, not this one's. Passed
-      // through unexpanded so it still compares, rather than being dropped into
-      // a silently smaller list on the containing side.
-      if (typeof entry !== 'string') {
-        atoms.add(entry);
-        continue;
-      }
-
-      const atIndex = entry.startsWith('@')
-        ? entry.indexOf('@', 1)
-        : entry.indexOf('@');
-
-      if (atIndex === -1) {
-        atoms.add(entry);
-        continue;
-      }
-
-      const packageName = entry.slice(0, atIndex);
-
-      for (const version of entry.slice(atIndex + 1).split('||')) {
-        atoms.add(`${packageName}@${version.trim()}`);
-      }
-    }
-
-    return [...atoms];
-  };
-
+  // which is the asymmetric hardening this describe block exists to catch. The
+  // two sides are compared as atoms; see `exemptionAtoms` above for why literal
+  // membership cannot see a legitimate containment.
   test('.gaia/cli exempts nothing from the release-age window that root does not', () => {
     expect(exemptionAtoms(rootSettings.minimumReleaseAgeExclude)).toEqual(
-      expect.arrayContaining(exemptionAtoms(cliSettings.minimumReleaseAgeExclude))
+      expect.arrayContaining(
+        exemptionAtoms(cliSettings.minimumReleaseAgeExclude)
+      )
     );
   });
 
--- a/.gaia/cli/src/lint-pin-parity.test.ts
+++ b/.gaia/cli/src/lint-pin-parity.test.ts
@@ -375,19 +375,36 @@ describe('supply-chain hardening parity', () => {
     );
   });
 
-  test('a merged union contains the single version it was merged from', () => {
+  // The four below are unit tests of `exemptionAtoms` over constructed entries,
+  // not assertions about either live `pnpm-workspace.yaml`. Named for the helper
+  // so they cannot be read as live parity coverage: the two tests that do assert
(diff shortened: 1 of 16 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.uXFpClupwM

Issue #1232, opened 2026-08-05
The concurrency suite hand-rolls the hook-invocation idiom 14 times and cannot reach the shared helper


stevensacks:
**Location:** `.gaia/tests/concurrency/concurrency.bats:107`, and 13 further copies in the same file.

`.gaia/tests/hooks/helpers/run-hook.sh` now holds the hook-invocation idiom for the `.gaia/tests/hooks/` suites, which all reach for it. The concurrency suite hand-rolls the same delivery 14 times and reaches for nothing:

- **1 plain copy**, `run bash -c 'printf %s "$1" | bash "$2"' _ "$json" "$hook"` (line 107), which `invoke_hook` expresses exactly;
- **11 wrapped in `run run_in "$B" -- bash -c …`**, the worktree-isolation wrapper this suite needs;
- **2 wrapped in `run env HOME=… bash -c …`**.

## Failure mode

State this precisely, because the obvious reading is wrong: **all 14 copies are already the quote-safe positional form, so no payload is being mangled today.** This is not the defect `#1219` describes; it is the *condition* that produced it.

A maintainer adding a hook invocation to this file has no shared primitive to reach for. The next one either copy-pastes a neighbour or writes a plausible variant (`bash "$HOOK" <<<"$json"`, or a payload spliced into an outer single-quoted string), and a fixture carrying a quote of its own then delivers something other than what it spells. The hook denies for the wrong reason, or never parses the payload, and the test greens having proved nothing about the case it is named for. That is exactly how the sibling directory acquired the drift `#1219` documents: 29 hand-copies, two incompatible deny contracts, three different matchers.

The suite is also the one place where a mis-delivered payload is hardest to notice, since its assertions are about *which tree* a guard attributes a write to rather than about the payload text.

## Suggested fix

Give the concurrency suite the same primitive. The helper as written calls bats' `run` itself, so the 13 wrapped call sites cannot adopt it as-is; that is the real work here and it is a genuine choice, not a mechanical port:

- add a variant that composes with a caller-supplied runner (`run_in "$B" --`, `env HOME=…`) rather than calling `run` itself, and have both directories use it; or
- restructure the wrapped call sites so the wrapper moves inside the helper; or
- keep the helper hooks-local and give `.gaia/tests/concurrency/lib/` its own, accepting two implementations of one idiom for two genuinely different harnesses.

The middle option is the one that most easily makes the artifact bigger for no behavioural gain, so weigh it against the third rather than assuming consolidation wins.

[...]

PR #1233, opened 2026-08-05
debt(tests): one shared primitive per suite for CLI-stderr diagnostics and hook invocation (#1202, #1219)

### LINKED ISSUE #1202: The distribution suite has two competing CLI-stderr diagnostic mechanisms and neither lives in lib/lib.sh

### LINKED ISSUE #1219: The hook bats suites hand-roll one quote-safe invocation idiom 29 times and their own deny assertion 16 times, and the copies have drifted
The hook suites under `.gaia/tests/hooks/` each hand-roll the same two pieces of harness, and the copies have already drifted.

`.gaia/tests/hooks/block-invalid-yaml-write.bats:70` is a representative copy of the invocation idiom:

```bash
run bash -c 'printf %s "$1" | bash "$2"' _ "$json" "$HOOK_ABS"
```

That exact line, or a near-copy of it, appears in **29** of the 46 suites, and **16** suites define their own `assert_denied` / `assert_blocked` / `assert_allowed` pair. A `helpers/` directory already exists beside them and is already the established place for shared harness (`mock-hook-input.sh` and `tmp-git-repo.sh` are sourced by five suites), but neither of these two idioms lives there.

**Failure mode.** The invocation idiom is not arbitrary: passing the payload and the hook path as positional arguments is what keeps a fixture carrying quotes of its own from being re-wrapped and mangled before it reaches the hook. Hand-copying it means a suite can write a plausible variant (`run bash "$HOOK" <<<"$json"`, or an inline `jq -n` program with the payload spliced into single quotes) that delivers a *different* payload than the one the fixture spells. The hook then denies for the wrong reason, or never parses the payload at all, and the test greens while proving nothing about the case it is named for. That is not hypothetical: authoring PR #1218 hit this shape twice in its own reproduction harnesses, each time caught only because a control case that should have denied also passed.

**The assertion copies have measurably drifted.** Two incompatible deny mechanisms are asserted across the suites with no shared vocabulary saying which hook uses which:

- `[ "$status" -eq 2 ]` plus a `BLOCKED` substring, in `block-bare-test.bats:33`, `serena-code-search-guard.bats:66`, `block-eslint-config-edit.bats:111`;
- `[ "$status" -eq 0 ]` plus a `"permissionDecision": "deny"` substring, in eleven others.

Within the second group the copies disagree on how they match: `grep -qF`, a bare `[[ ]]`, and a `jq -r` field read all appear. `block-secrets-write.bats:106` carries an extra negative guard none of the others has, a per-suite improvement that never propagated because there was nothing to propagate it to.

**Suggested fix.** Add `.gaia/tests/hooks/helpers/run-hook.sh` next to the two helpers already there, exporting the quote-safe invocation and one deny/allow assertion pair per mechanism (named for the mechanism, so a suite picks the one its hook actually uses rather than a generic `assert_denied` that hides which contract is under test), and source it from the suites. Keep the assertion forms bash-3.2 safe per `.claude/rules/bats-assertions.md`, since consolidating is a chance to fix the bare `[[ ]]` copies rather than preserve them.

Found while adding one more instance of the pattern in PR #1218, which does not originate it.

Handler: plan

stevensacks:
Closes #1202
Closes #1219

Two test suites had grown their own copies of one primitive each, and in both cases the copies had drifted. This gives each directory a shared home for its primitive and converges every call site on it.

## `#1202`, the distribution suite's CLI-stderr diagnostic

Every scenario asserting a CLI call *succeeds* has to keep that call's stderr off stdout, because stdout is the JSON it is about to parse. Nine scenarios solved that two different ways and neither lived in `lib/lib.sh`: seven hand-copied a block that re-runs the identical command unsuppressed on failure, one captured stderr to a file.

`lib/lib.sh` now carries the mechanism as `capture_cli_stderr` / `run_cli` / `fail_with_stderr`, and all nine scenarios use it.

- **Mechanism B won, and the reason is substantive rather than aesthetic.** A re-run is a second, different execution: it can succeed and leave the failure unexplained, fail differently, or double the side effects on the one path already going wrong. Capturing reports the bytes the *failing* run emitted. This is also the fix `#1045` named.
- **The capture path is the caller's**, declared once per scenario. A library-allocated temp file would need a library-installed `EXIT` trap, and every scenario sets its own `trap` afterwards, which would silently replace it. Scenarios with a scratch dir that is not the tree under assertion use it; the rest `mktemp` a file and add it to the trap they already have.
- **Extending the frozen API was a deliberate call, which is why `#1202` exists as an issue.** The header's `API (frozen; see plan README)` points at the plan document for the suite's original pull request, which no longer exists in `.gaia/local/plans`, so the freeze is undocumented and unenforceable. The `frozen` marker is dropped rather than left pointing nowhere, and the header's API list gains the three functions.
- `14-gaia-update-deps.sh` had a second success-asserting call (`decline --clear`) with no diagnostic at all. It gets one for free.

## `#1219`, the hook suites' invocation and assertions

`.gaia/tests/hooks/helpers/run-hook.sh` is new, beside the two helpers already there. It holds `invoke_hook` / `invoke_hook_in` and **two assertion pairs named for the mechanism they check**, because these suites test hooks with two incompatible deny contracts:

| contract | deny | allow |
|---|---|---|
| exit code carries the verdict | `assert_blocked_by_exit` (exit 2 + `BLOCKED`) | `assert_allowed_by_exit` (exit 0, silent) |
| JSON on stdout carries the verdict | `assert_denied_by_json` | `assert_allowed_by_json` |

A generic `assert_denied` hides which contract is under test, which is how `block-eslint-config-edit.bats` came to call its exit-2 assertion `assert_denied`.

33 suites converted. Every form in the helper is bash-3.2 safe per `.claude/rules/bats-assertions.md`; four suites' copies were not (`[[ ]]` matchers in `block-manifest-write` / `block-main-destructive-git` / `block-no-verify`, a `!`-negation in `block-secrets-write` and `block-env-read`), and those fix themselves by converging.

**Nothing was weakened to fit the shared form.** Three suites assert more than the pair does and keep the extra:

- `block-spec-plan-chain.bats` keeps its own pair entirely: its deny reads the decision as a JSON *field* rather than as a substring, and its hook is silent on an allow. Both are narrower than the shared pair, so trading would lose coverage. The reason is now a comment in the file.
- `distribution-preflight-check.bats` composes: `assert_allowed_by_json` **plus** its own empty-output assertion.
- `block-secrets-write.bats` composes: its `assert_denied` / `assert_denied_cap` keep the cap-message guard that never propagated to its siblings.

**Two things the conversion changed that are worth naming:**

1. **Seven suites gained an `is executable` test.** They reached the hook by direct execution (`| '$HOOK_ABS'`), so the exec bit was under test only incidentally; the shared invocation runs `bash "$hook"`. Rather than smuggle the property back through the delivery form, each suite that had no such test now names it. Eight suites already did.
2. **Four teardowns are fixed.** `[ -n "${REPO:-}" ] && rm -rf "$REPO"` is an AND-list that returns 1 when there is no scratch repo to remove, failing an innocent test. Every existing test happened to create one; the new exec tests do not, which surfaced it. Fixed with the `return 0` the sibling suites already use.

Ten suites opened with a paragraph restating the positional-argument mechanism line by line. Each keeps the part still about that suite (what its own payloads carry) and names the helper for the rest.

## What is deliberately left alone

- **`pr-merge-audit-check.bats:832` and `:852`** merge stderr into `$output` with `2>&1`, which the tests' own comments say is deliberate. The helper does not express that and should not.
- **`block-rm-rf.bats:1346` and `:1357`** `source` the hook to probe a helper and to prove it leaks no shell options. Not deliveries.
- **`block-eslint-config-edit.bats:170`** feeds a 1MB payload from a file on purpose.

## Verification

- **19-case adversarial proof for `run-hook.sh`**, run against throwaway hooks in a scratch tree. Each assertion is proved to **fail** when the property it names is false: a `BLOCKED` message on the wrong exit code, a silent exit-2 block, a noisy allow, a deny on a non-zero exit, a missing deny. Each is also proved to fail **from a non-final line**, which is the hollow-assertion trap `.claude/rules/bats-assertions.md` exists for. Delivery is proved byte-identical for a payload carrying single quotes, double quotes and a newline, through both `invoke_hook` and `invoke_hook_in`, plus the cwd and the env-prefix behaviour that several suites depend on.
- **12-case adversarial proof for the `lib.sh` capture**: the failing run's stderr is reported, stdout stays clean, a silent failure says so, an earlier call's stderr is never reported as this one's, and a scenario that forgot `capture_cli_stderr` fails closed instead of running the command. Plus an end-to-end mutation of scenario 13 (a missing `--baseline`) that printed the real CLI error inside the fenced block ahead of the FAIL line.
- Every touched distribution scenario run individually; `shell-lint.sh`; the bats suites under bash 5.

**The Quality Gate skips by its own criterion.** Its grep over the 44 changed files returns nothing: no `.ts|tsx|js|jsx|mjs|cjs|css`, no gate-affecting config. `simplify` is step 1 of that gate and skips with it. The verification this diff owes is the shell gate and the suites, above.

**No CHANGELOG entry is owed**, verified rather than assumed: all 44 changed files were checked against `.gaia/manifest.json` and none is in it, so nothing ships. Both issues are `surface:maintainer`.

## Out-of-scope findings

- **Filed as #1232**: `.gaia/tests/concurrency/concurrency.bats` hand-rolls the same delivery idiom 14 times and cannot reach the new helper, because 13 of the copies wrap it in `run_in` or an env prefix while the helper calls bats' `run` itself. Filed rather than waived because that file is outside this pull request. Stated honestly in the issue: all 14 copies are already the quote-safe form, so the mangling failure mode is not present there; what is present is the condition that produced the drift here, no shared primitive to reach for.

## Audit gate

Round 1, `code-audit-maintainer-shell` (the only member the roster dispatches for an all-`.bats`/`.sh` diff). **Clean: no Critical, no Important, two Suggestions.** The member verified rather than read: `@test` count 1088 → 1095 with no test deleted and no assertion loosened, every removed local assertion replaced by an equal-or-stronger shared one, all five helper forms mutation-proved in non-final position on bash 3.2.57 *and* bash 5.3.15, and the full hooks directory green at 1297/1297.

Both Suggestions are **in-scope and accepted rather than folded in**, per the digest economics in `wiki/concepts/PR Merge Workflow.md`: the member is already marked and nothing else is rotating its digest, so folding either in buys a full re-dispatch to re-earn the marker, and neither is consequential enough to be worth it.

**1. `helpers/run-hook.sh:44`, interpreter selection moved from shebang to PATH.** `invoke_hook` runs `bash "$hook"`, so a hook's own shebang no longer picks the interpreter. Two of the eight affected hooks (`debt-sentinel-touch.sh`, `debt-session-reconcile.sh`) pin `#!/bin/bash` where the other six pin `#!/usr/bin/env bash`; verified. Accepted, and the reason is not only cost: **the fidelity this trades away is fidelity the repo tells maintainers not to rely on.** CI runs these suites on `ubuntu-latest`, where `/bin/bash` is bash 5 either way, so the difference only ever existed on a maintainer's local macOS run, and `.claude/rules/bats-assertions.md` directs local bats runs through the bash-5 guard for exactly that reason. The member confirmed no live failure: both suites pass 20/20 under either interpreter. Normalizing the two shebangs would change which bash runs an adopter's hook, which is a production change that deserves its own review rather than a ride-along on a test refactor.

**2. `pr-merge-audit-check.bats:832` and `:852` stay on the pre-refactor idiom**, because they merge stderr with `2>&1` and the helper has no variant for that. Accepted; the exemption is named in "What is deliberately left alone" above so the next reader does not read it as a miss. Both payloads are jq-built from a fixed command with no quote of their own, which the member confirmed.

The member wrote its marker and findings sidecar, stamped the trailer, pushed, and posted the `GAIA-Audit` success status. The dispatch classified `real` against both artifacts.

### DIFF
--- a/.gaia/tests/distribution/07-gaia-init-strip-branding.sh
+++ b/.gaia/tests/distribution/07-gaia-init-strip-branding.sh
@@ -28,7 +28,10 @@ require_cmd rsync "rsync required for adopter-flow scaffold copy"
 
 STAGING="$(mktemp -d -t gaia-dist-init-stage-XXXXXX)"
 SCAFFOLD="$(mktemp -d -t gaia-dist-init-scaffold-XXXXXX)"
-trap 'rm -rf "$STAGING" "$SCAFFOLD"' EXIT
+# Outside the scaffold, which this scenario asserts on file by file.
+CLI_STDERR_FILE="$(mktemp -t gaia-dist-init-stderr-XXXXXX)"
+trap 'rm -rf "$STAGING" "$SCAFFOLD" "$CLI_STDERR_FILE"' EXIT
+capture_cli_stderr "$CLI_STDERR_FILE"
 
 "$HERE/lib/build-staging.sh" "$STAGING" \
   || { fail "build-staging failed"; exit 1; }
@@ -52,15 +55,8 @@ fi
 # own pwd; never `cd "$SCAFFOLD"` at scenario scope, since other
 # scenarios sourced or invoked from `run-all.sh` may rely on $PWD.
 TITLE="Test Project"
-STDOUT="$(cd "$SCAFFOLD" && "$SCAFFOLD/.gaia/cli/gaia" init strip-branding --title "$TITLE" 2>/dev/null)" || {
-  # Re-run with stderr unsuppressed for diagnosis. The `fail; exit 1`
-  # below runs unconditionally; the diagnostic re-run's exit code is
-  # intentionally ignored (`|| :`).
-  log "gaia init strip-branding exited non-zero; rerunning with stderr:"
-  ( cd "$SCAFFOLD" && "$SCAFFOLD/.gaia/cli/gaia" init strip-branding --title "$TITLE" ) || :
-  fail "gaia init strip-branding exited non-zero on staged tree"
-  exit 1
-}
+STDOUT="$(cd "$SCAFFOLD" && run_cli "$SCAFFOLD/.gaia/cli/gaia" init strip-branding --title "$TITLE")" \
+  || fail_with_stderr "gaia init strip-branding exited non-zero on staged tree"
 
 if [ -n "$STDOUT" ]; then
   log "unexpected stdout from strip-branding (contract: no stdout on success):"
--- a/.gaia/tests/distribution/08-gaia-init-cli-sequence.sh
+++ b/.gaia/tests/distribution/08-gaia-init-cli-sequence.sh
@@ -54,7 +54,10 @@ require_cmd rsync "rsync required for adopter-flow scaffold copy"
 
 STAGING="$(mktemp -d -t gaia-dist-init-stage-XXXXXX)"
 SCAFFOLD="$(mktemp -d -t gaia-dist-init-scaffold-XXXXXX)"
-trap 'rm -rf "$STAGING" "$SCAFFOLD"' EXIT
+# Outside the staged tree, which these scenarios assert on file by file.
+CLI_STDERR_FILE="$(mktemp -t gaia-dist-init-stderr-XXXXXX)"
+trap 'rm -rf "$STAGING" "$SCAFFOLD" "$CLI_STDERR_FILE"' EXIT
+capture_cli_stderr "$CLI_STDERR_FILE"
 
 "$HERE/lib/build-staging.sh" "$STAGING" \
   || { fail "build-staging failed"; exit 1; }
@@ -109,15 +112,8 @@ export GAIA_TELEMETRY_PING_DISABLE=1
 run_step() {
   local label="$1"; shift
   local stdout
-  stdout="$(cd "$SCAFFOLD" && "$GAIA" "$@" 2>/dev/null)" || {
-    # Re-run with stderr unsuppressed for diagnosis. The `fail; exit 1`
-    # below runs unconditionally; the diagnostic re-run's exit code is
-    # intentionally ignored (`|| :`).
-    log "gaia $* exited non-zero; rerunning with stderr:"
-    ( cd "$SCAFFOLD" && "$GAIA" "$@" ) || :
-    fail "gaia $* exited non-zero on staged tree (step: $label)"
-    exit 1
-  }
+  stdout="$(cd "$SCAFFOLD" && run_cli "$GAIA" "$@")" \
+    || fail_with_stderr "gaia $* exited non-zero on staged tree (step: $label)"
   if [ -n "$stdout" ]; then
     log "unexpected stdout from gaia $* (contract: no stdout on success):"
     printf '%s\n' "$stdout" >&2
--- a/.gaia/tests/distribution/10-gaia-scaffold-templates.sh
+++ b/.gaia/tests/distribution/10-gaia-scaffold-templates.sh
@@ -33,7 +33,10 @@ require_cmd rsync "rsync required for adopter-flow scaffold copy"
 
 STAGING="$(mktemp -d -t gaia-dist-scaffold-stage-XXXXXX)"
 SCAFFOLD="$(mktemp -d -t gaia-dist-scaffold-scaffold-XXXXXX)"
-trap 'rm -rf "$STAGING" "$SCAFFOLD"' EXIT
+# Outside the staged tree, which these scenarios assert on file by file.
+CLI_STDERR_FILE="$(mktemp -t gaia-dist-scaffold-stderr-XXXXXX)"
+trap 'rm -rf "$STAGING" "$SCAFFOLD" "$CLI_STDERR_FILE"' EXIT
+capture_cli_stderr "$CLI_STDERR_FILE"
 
 "$HERE/lib/build-staging.sh" "$STAGING" \
   || { fail "build-staging failed"; exit 1; }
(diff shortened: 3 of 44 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.LBHSoHTmcC

PR #1234, opened 2026-08-06
debt(tests): gate git auto-maintenance in the commit-classify sandbox (#1216)

### LINKED ISSUE #1216: commit-classify's test sandbox is repacked mid-run despite gc.auto=0, corrupting HEAD and reddening an unrelated PR
`.gaia/cli/src/wiki/commit-classify.test.ts:84` (the `gc.auto` line of `setupSandbox`)

Surfaced by a CI failure during PR #1215, not by an audit member. That pull request changed a help-text string constant, a wiki sentence, the two rebuilt CLI bundles, and `CHANGELOG.md`, none of which this suite reads.

## Failure mode

The `Vitest (.gaia/cli)` job failed on `src/wiki/commit-classify.test.ts` at `wiki commit-classify > health signal > a large healthy sample reports neither inert nor a warning`. The identical tree passed the same job on a plain re-run, and passed locally in a full run (1967 passed / 131 files), so it is intermittent.

The suite's own diagnostic (`describeSandbox`) is what makes this more than a flake report. It shows the per-test sandbox repository is **corrupt**, not merely contended:

```
Command failed: git commit -q -m docs: prose 23
fatal: could not parse HEAD

  HEAD ref: refs/heads/main
  HEAD sha: 831ef6cbc9a57824c43918f2fa40e5936c59e5ae
  objects:  count: 1 size: 4 in-pack: 93 packs: 2 size-pack: 11 prune-packable: 0 garbage: 0
  fsck:     <Command failed: git fsck --no-progress --connectivity-only>
```

Three facts together: `rev-parse HEAD` resolves the ref while `git commit` cannot parse the object it names, `git fsck --connectivity-only` fails outright, and the sandbox holds **two packfiles and 93 in-pack objects** roughly 24 commits into a repository created seconds earlier by `mkdtempSync`. A repack ran in that sandbox and the commit HEAD points at is no longer reachable in the object store.

That contradicts the setup's own stated defence. `setupSandbox` sets `gc.auto=0` at `:84` specifically to stop the `git maintenance run --auto --quiet --detach` that each `git commit` otherwise spawns, and the comment above it records a measurement ("3 spawns per commit by default, 0 with this set"). Two packfiles say something packed this repository anyway, so the suppression does not hold under CI's parallel load, whatever held when it was measured. `gc.auto` governs only the automatic gc heuristic; it does not disable the maintenance task set, and a detached maintenance process racing the next `git commit` in the same repository is consistent with every symptom above.

The consequence is a false red: a pull request whose diff cannot reach this suite gets a failing required check, and the operator either re-runs on a hunch or spends a triage on someone else's fixture. This is the second known intermittent failure of this shape in the repository, after #1204 (`token-rollup.bats` under concurrent load), and a different suite with a different mechanism.

## Suggested fix

Establish the mechanism first, then harden the sandbox; the fix should not be guessed from the symptoms above.

The likely direction is to suppress the maintenance task set rather than only the gc heuristic: set `maintenance.auto=false` and `gc.autoDetach=false` alongside the existing `gc.auto=0`, and consider running the sandbox's git under `GIT_OPTIONAL_LOCKS=0`. Whatever is chosen, prove it the way the existing comment claims to have been proven: count the spawned git processes for a 25-commit scenario, under parallel load rather than in isolation, since isolation is the condition the current measurement holds under and CI is the one it fails under.

Worth checking while there: `describeSandbox`'s `probe` reports only the first line of a failure (`error.message.split('\n', 1)[0]`), which is why the `fsck` row above names no cause. Keeping git's stderr on that row would have made this diagnosable from the first failure rather than the second.

Handler: prompt

stevensacks:
Closes #1216

## What was wrong

The `commit-classify` sandbox's only defence against git's own background maintenance was `gc.auto=0`. That governs the gc heuristic, not the maintenance task set, so it says nothing about the `loose-objects` and `incremental-repack` tasks, and those are the ones that pack.

CI found the sandbox holding two packfiles and 93 in-pack objects roughly 24 commits into a repository created seconds earlier, with a HEAD whose commit object `git commit` could no longer parse. A detached maintenance run had repacked the repository underneath the suite and reddened a pull request whose diff cannot reach this file.

## Mechanism, as far as the evidence establishes it

Measured rather than inferred, on git 2.55:

- `gc.auto=0` **does** suppress the spawn outright here, which is what the comment it carried claimed. The defence is one version-dependent heuristic, and the CI evidence proves it did not hold there.
- `maintenance.auto=false` is the documented gate on the spawn itself and covers the whole task set.
- Tracing the entire suite with `GIT_TRACE2_EVENT` showed exactly **one** unsuppressed `git maintenance run --auto --quiet --detach` per run today, from the decoy repository in the hermeticity test, which set no suppression at all and which `rmSync` then deletes out from under. That is the same hazard class, presently reproducible, and it is now closed.

## What changed

- Suppression is stated in the spelling each git generation reads: `maintenance.auto` for the task set, `gc.auto` for git predating it, and `gc.autoDetach` so a run that starts anyway finishes before its triggering command returns rather than outliving the test or racing the next commit. `maintenance.autoDetach` is deliberately absent: every git new enough to read it also honors `maintenance.auto=false`.
- Every repository the fixture creates now goes through one initializer, so none can be the one that quietly skips it.
- `describeSandbox`'s probe rows keep git's own explanation. They truncated to the first line, which is why the CI report says `fsck` failed while naming no cause, and why this took a second failure to diagnose rather than the first.

## Verification

- Full CLI suite green: 131 files, 1982 passed. Root typecheck, root lint, CLI typecheck, CLI lint, root tests (156), and `pnpm build` all green.
- The new guard was proven falsifiable on **both** arms, per the pre-dispatch rule:
  - removing the gates from the shared config list fails the subject arm (`expected [ …(5) ] to deeply equal []`, one spawn per commit);
  - giving the control repository the full suppression fails the control arm (`expected 0 to be greater than 0`).
- The control exists because a bare "no maintenance was spawned" assertion passes just as well against a trace that recorded nothing. It proves in the same run that git still spawns maintenance on commit and that the trace captures it.

Playwright was not run: the diff is one file in the `.gaia/cli` workspace, which no `app/` code imports and no browser flow can reach.

## Known follow-up (out of scope here)

Seventeen other `.gaia/cli/src` suites build git sandboxes the same way and carry no suppression. Their exposure is lower (5-12 commits per sandbox, against 25+ here, and the maintenance auto-conditions are threshold-based), but the class is real. Promoting this config to a shared fixture touches all of them and is its own change, filed separately rather than folded into a single-issue debt fix.

## Audit rounds

Four rounds of `code-audit-maintainer-node`, each finding addressed:

1. **Important** the control repository established its ungated state by omission, so a machine whose `~/.gitconfig` carries `gc.auto = 0` (a common dotfile) would give the control the very gate it must lack and redden the suite against a fixture behaving perfectly. Fixed: the gates now carry git's own defaults beside them and the control writes those back. Two Suggestions on prose accuracy and the trace filter's spelling coverage were applied in the same round.
2. **Important** the same defect one key over: the control's *foreground* state was still inherited. Modern git resolves `maintenance.autoDetach` ahead of `gc.autoDetach`, so a global `maintenance.autoDetach = true` detached the control's run and the cleanup then deleted its directory out from under a live git. Fixed by restoring `maintenance.autoDetach`, with the docblock now recording why it is not the redundant key it looks like.
3. **Suggestion** the control arm asserted a spawn count and never the argv, so round 2's hazard had prose behind it and nothing else. Applied: the arm now also asserts no spawn carries `--detach`, proven to fail by dropping the key under that global.

### Accepted and noted, not fixed

Round 4 was clean (no Critical, no Important) with one Suggestion the member itself recommended no change for, recorded here instead:

- `.gaia/cli/src/wiki/commit-classify.test.ts:903` the `--detach` assertion reads the child's argv, so it detects detachment only on a git that spells it there. Git predating the maintenance task set daemonizes inside `git gc --auto` with no argv flag, so that arm passes because the flag is absent from argv, not because that git cannot detach. Reaching it needs both an ancient git and a future edit dropping `gc.autoDetach`, and no such git was installed to test against.

### DIFF
--- a/.gaia/cli/src/wiki/commit-classify.test.ts
+++ b/.gaia/cli/src/wiki/commit-classify.test.ts
@@ -892,7 +892,19 @@ describe('commit-classify sandbox fixture', () => {
         controlRoot
       );
 
-      expect(autoMaintenanceSpawns(controlTrace).length).toBeGreaterThan(0);
+      const controlSpawns = autoMaintenanceSpawns(controlTrace);
+      expect(controlSpawns.length).toBeGreaterThan(0);
+      // The control's own run must stay in the foreground, or the `finally`
+      // below deletes `controlRoot` out from under a live background git. A
+      // count alone cannot see that: the parent records the child either way,
+      // so this test would pass green while leaking the orphan the whole
+      // fixture exists to prevent. Asserted as an absence to stay portable,
+      // git predating the task set spawns `git gc --auto`, which carries no
+      // detach flag at all and passes, while modern git catches a
+      // `maintenance.autoDetach` that a later edit dropped as redundant.
+      expect(controlSpawns.some((line) => line.includes('"--detach"'))).toBe(
+        false
+      );
 
       const sandboxTrace = path.join(traceRoot, 'sandbox.json');
       vi.stubEnv('GIT_TRACE2_EVENT', sandboxTrace);
--- a/.gaia/cli/src/wiki/commit-classify.test.ts
+++ b/.gaia/cli/src/wiki/commit-classify.test.ts
@@ -95,12 +95,14 @@ const sandboxGit = (root: string, args: string[]): string => {
  * when `git maintenance` replaced `git gc --auto`, and the git the fixture
  * runs under is whatever the machine or the CI runner supplies:
  *
- * - `maintenance.auto=false` is the gate on modern git, and it covers the whole
- *   task set. `gc.auto` governs only the gc heuristic, so it says nothing about
- *   the `loose-objects` and `incremental-repack` tasks, and those are the ones
- *   that pack.
- * - `gc.auto=0` is the same gate for git predating that task set, which reaches
- *   the repository through `git gc --auto` instead.
+ * - `maintenance.auto=false` is the documented gate on the spawn, and it covers
+ *   the whole task set, including the `loose-objects` and `incremental-repack`
+ *   tasks, which are the ones that pack.
+ * - `gc.auto=0` is the gate for git predating that task set, which reaches the
+ *   repository through `git gc --auto` instead. On git 2.55 it also suppresses
+ *   the modern spawn on its own, so what failed in CI is not established to be
+ *   this key being the weaker one; what is established is that a single key was
+ *   not enough on the git that ran there.
  * - `gc.autoDetach=false` is the fallback for both: a run that starts anyway
  *   finishes before the command that triggered it returns, so it can neither
  *   outlive the test nor race the next `git commit`. `maintenance.autoDetach`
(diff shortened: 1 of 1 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.A3puRqeOa3

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.JaxGxzONcO

Issue #1229, opened 2026-08-05
No deterministic check reaches a shell call site of diff --name-only, and the quoting class has now spread four times


stevensacks:
## No deterministic check reaches a shell call site of `diff --name-only`, and the class has now spread four times

`.gaia/scripts/check-audit-base-derivation.sh:557` (the candidate net; assertion 4 consumes it at `.gaia/scripts/check-audit-base-derivation.sh:568`)

```bash
diff_candidates="$(git -C "$repo_root" grep -nIF "$GAIA_AUDIT_DIFF_CALL" -- '.claude/agents/code-audit-*.md' 2>/dev/null)"
```

### Failure mode

Assertion 4 exists to stop a `diff --name-only` from letting git C-quote the paths it prints. Its candidate net is scoped to `.claude/agents/code-audit-*.md`, so it has never been able to see a `.sh` or `.yml` call site. The check prints `changed-file diffs that let git C-quote a path: 0` and passes, whether or not any shell consumer in the repository quotes.

That is not hypothetical drift. The same defect has now been found and fixed **four times in four different files**, each time by a human or an audit member reading the code rather than by a check:

- `.gaia/scripts/resolve-audit-members.sh` and the roster derivations (`#1213`, PR #1223)
- `.github/audit/resolve-audit-base.sh` (`#1224`, PR #1227)
- `.claude/hooks/worthiness-presence-check.sh` (`#1225`, PR #1227)
- `.claude/hooks/distribution-preflight-check.sh` and `.github/workflows/distribution-audit-pr.yml` (filed alongside this issue)

`#1213`'s own repair landed on one of two sibling derivations in a single file and nothing noticed the other; assertion 4 was added in response, and it still cannot see any of the sites above.

### The design fork, stated rather than hidden

This is not a one-token widening, which is why it is filed rather than folded into PR #1227:

1. **Widen assertion 4's net to shell.** The blocker is that `_gaia_keep_diff_matches_missing`'s `consumes` test keys on `resolve-audit-base.sh` / `BASE_REF` / `BASE_SHA` / `FULL_BASE` appearing inside the call window, and **none of the four shell sites above names any of them**. So the net cannot simply be pointed at `.sh`; the predicate for "this call's output is matched against path-shaped data" has to be redefined for shell.
[...]

Issue #1165, opened 2026-08-02
Local distribution pre-flight mirrors only the missing-intersection arm, so region drift still surfaces as a red check after push


stevensacks:
**Location:** `.claude/hooks/distribution-preflight-check.sh:269`

## Failure mode

The local pre-flight hook exists so a maintainer learns about a distribution-gate failure at
`gh pr create` time rather than from a red required check after the push. It mirrors the
`missing`-intersection arm of `.github/workflows/distribution-audit-pr.yml` and **only** that
arm: `grep -c regionDrift .claude/hooks/distribution-preflight-check.sh` returns `0`.

The CI gate has a second, independent failure condition. When the committed
`.gaia/manifest.json`'s region declaration disagrees with a fresh scan, the workflow fails on
the whole symmetric difference, deliberately not intersected with the PR's changed files.
The hook does not evaluate that condition at all.

Concrete bad outcome: with region drift present, `gh pr create` passes the local gate, the
branch pushes, and the required `Distribution Audit` context goes red, which is exactly the
after-the-fact redness the hook's denial message says it exists to prevent.

**Two comments overstate the coverage**, which is what makes this more than a gap:

- `:269` "Matches distribution-audit-pr.yml exactly so the two gates never disagree."
- `:313` "This gate mirrors .github/workflows/distribution-audit-pr.yml deliberately."

Both read as a completeness claim. A maintainer debugging a red check that passed locally
reads them and concludes the CI gate misfired.

## Suggested fix

Two directions, and the choice is a real one rather than an oversight to correct:

1. **Add the arm.** The hook already parses the same `release manifest --check --json`
   output the workflow does, so reading `.regionDrift` is a small addition. The cost is that
   region drift is a repository-wide condition with no per-branch scoping, so it would deny
   every `gh pr create` in the clone until the manifest is regenerated, including PRs that
   have nothing to do with regions. That may be correct, or it may be far too blunt for a
   local pre-flight.
2. **Keep the narrow scoping and fix the comments.** If per-PR scoping is the deliberate
   design for the local hook, say so at both sites: name the second condition, say the hook
   does not evaluate it and why, and drop the "matches exactly" / "mirrors" phrasing.

Direction 2 is the smaller change and is correct regardless, since the comments are wrong
under either outcome.

## Scope note

Pre-existing: the region arm and this hook's omission of it both predate PR #1162, whose only
change to this area is a header comment in the workflow naming the second condition. That
[...]

PR #1237, opened 2026-08-06
debt(distribution): NUL-delimit the changed-file diff on both distribution gates (#1228, #1230)

### LINKED ISSUE #1228: A C-quoted path makes the distribution pre-flight allow a file with no ship-or-withhold answer
## A C-quoted path makes the distribution pre-flight allow a file with no ship-or-withhold answer

`.claude/hooks/distribution-preflight-check.sh:270`

```bash
changed=$( cd "$audited_root" && git diff --name-only --diff-filter=ACMR "${base_rev}...HEAD" 2>/dev/null \
  | LC_ALL=C sort -u || true)
```

### Failure mode

Under git's default `core.quotePath`, `git diff --name-only` C-quotes any path carrying non-ASCII or control bytes, wrapping it in double quotes and backslash-escaping the offending bytes, and emits those quotes as literal characters.

`changed` is intersected with `missing` two lines below via `comm -12`. `missing` comes from the manifest checker's JSON through `jq -r` (`missing: {expected, file}[]`, `.gaia/cli/src/release/manifest.ts`), which emits raw bytes. So the two sides are in different encodings for exactly the paths at issue: the quoted token never matches its raw counterpart, `offenders` comes back empty, and the hook **allows `gh pr create`** for a newly-shipping file that has no ship-or-withhold answer in the manifest.

The comment at `:269` names `.github/workflows/distribution-audit-pr.yml` as the gate this one mirrors, so a maintainer would expect CI to catch what the hook missed. **That mirror carries the identical un-`-z` spelling at `:118`**, so there is no backstop; it is filed separately because the two live in different files with different Code Audit Team owners.

### Suggested fix

```bash
changed=$( cd "$audited_root" && git diff --name-only -z --diff-filter=ACMR "${base_rev}...HEAD" 2>/dev/null \
  | tr '\0' '\n' | LC_ALL=C sort -u || true)
```

This is the spelling PR #1223 standardized across the audit membership derivations and PR #1227 applied to two more consumers. `-z` rather than `-c core.quotePath=false`, because the flag only stops treating bytes at or above 0x80 as unusual and still quotes a path containing a double quote, a backslash, or a control byte. Land it together with the workflow half so the two gates do not disagree.

### Reachability

Low. It needs a newly-shipping tracked path carrying non-ASCII or control bytes, and this repository has none today. Filed rather than left unrecorded because the miss is silent by construction: an empty `offenders` set is also what a correctly-answered branch produces, so the failure is indistinguishable from the ordinary pass.

### Severity note

The reporting member graded this **Important** under its own rubric, on the ground that it is a fail-open in a gate. It is filed `severity:suggestion` to match `#1224` and `#1225`, which are the identical defect class at identical reachability. Reachability is what decides it: all three need a tracked path carrying non-ASCII bytes and the repository has none. The member's grading is recorded here so a triager can overturn this call rather than having to rediscover it.

### Related

`#1165` cites `:269` in this same file, one line above, for a different defect (the hook mirrors only the `missing`-intersection arm and not the region-drift arm). The two are independent but adjacent, and whoever drains either should look at both.

### Provenance

Found by the `code-audit-maintainer-shell` round-1 audit on PR #1227 (`#1224` + `#1225`), which closes the same defect class on two other consumers. This is a third consumer on a path that pull request does not change, and the path is not gate machinery by `audit_path_is_machinery`, so neither cross-remit waive term applies and filing is the disposition the merge workflow's cross-remit rule names.

Handler: prompt

### LINKED ISSUE #1230: A C-quoted path makes the Distribution Audit check pass on an unanswered file
## A C-quoted path makes the Distribution Audit check pass on an unanswered file

`.github/workflows/distribution-audit-pr.yml:118`

```yaml
          git diff --name-only --diff-filter=ACMR "${BASE_SHA}...${HEAD_SHA}" \
            | LC_ALL=C sort -u > /tmp/pr-changed.txt
```

### Failure mode

Under git's default `core.quotePath`, `git diff --name-only` C-quotes any path carrying non-ASCII or control bytes, wrapping it in double quotes and backslash-escaping the offending bytes, and emits those quotes as literal characters.

That list is intersected with the manifest checker's `missing` set, which reaches the step as raw bytes through `jq -r`. The two sides are therefore in different encodings for exactly the paths at issue, the intersection comes back empty, and the required **Distribution Audit** check passes for a newly-shipping file that carries no ship-or-withhold answer in `.gaia/manifest.json`.

This is the CI half of `#1228`, and it is the half that matters most: the local hook's own comment names this workflow as the gate it mirrors, so the hook is documented as the cheap early warning and this as the authority. Both fail open on the same input, so there is no backstop on either side.

### Suggested fix

```yaml
          git diff --name-only -z --diff-filter=ACMR "${BASE_SHA}...${HEAD_SHA}" \
            | tr '\0' '\n' | LC_ALL=C sort -u > /tmp/pr-changed.txt
```

This is the spelling PR #1223 standardized across the audit membership derivations and PR #1227 applied to two more consumers. `-z` rather than `-c core.quotePath=false`, because the flag only stops treating bytes at or above 0x80 as unusual and still quotes a path containing a double quote, a backslash, or a control byte. Land it together with the local-hook half so the two gates do not disagree.

**Filed as its own issue rather than folded into the hook's**, on two grounds: the filing recipe keys dedup on a single `path=`+`line=` and one body carries exactly one key line, so bundling would leave one of the two undedupable; and this file's Code Audit Team owner is `code-audit-github-workflows`, not the shell member that owns the hook.

### Reachability

Low. It needs a newly-shipping tracked path carrying non-ASCII or control bytes, and this repository has none today. Filed rather than left unrecorded because a green required check is exactly what a correctly-answered branch produces, so the failure is indistinguishable from the ordinary pass.

### Severity note

The reporting member graded the sibling hook finding **Important** under its own rubric, on the ground that it is a fail-open in a gate; the same reasoning applies here. It is filed `severity:suggestion` to match `#1224` and `#1225`, the identical defect class at identical reachability. A triager who weighs the fail-open above the reachability should raise it rather than re-derive the question.

### Provenance

Surfaced as a cross-remit finding by the `code-audit-maintainer-shell` round-1 audit on PR #1227 (`#1224` + `#1225`), which closes the same defect class on two other consumers. The path is outside that member's remit and outside that pull request's diff, and it is not gate machinery by `audit_path_is_machinery`, so neither cross-remit waive term applies and filing is the disposition the merge workflow's cross-remit rule names.

Handler: prompt

stevensacks:
Closes #1228
Closes #1230

The two halves of the distribution gate, the local `gh pr create` pre-flight and the required CI check it mirrors, both read the changed-file list from `git diff --name-only` with git's default `core.quotePath` left on. A path carrying non-ASCII or control bytes comes back C-quoted, while the manifest checker's `missing` set arrives raw through `jq -r`, so the two sides never intersect for exactly those paths. `offenders` comes back empty and a newly-shipping file with no ship-or-withhold answer sails through: the hook allows the PR to be opened and the required check reports green.

Neither failure announces itself. An empty `offenders` set is also what a correctly-answered branch produces, so the miss is indistinguishable from an ordinary pass on both sides.

## What changed

Both call sites take `-z` plus `| tr '\0' '\n'`, the spelling PR #1223 standardized across the audit membership derivations and PR #1227 applied to two more consumers. `-z` rather than `-c core.quotePath=false`: the flag only stops treating bytes at or above 0x80 as unusual and still quotes a path containing a double quote, a backslash, or a control byte.

**They land together deliberately.** The hook's own comment names the workflow as the gate it mirrors, so the local pre-flight is documented as the cheap early warning and the workflow as the authority. Fixing one leaves the two disagreeing about the same input.

## Oracles

Each half is pinned by a behavioural test that was proven red against the pre-fix spelling before the fix landed, and each carries its own non-vacuity assertion, because a fixture whose paths were quietly all-ASCII would pass while proving nothing.

- `.gaia/tests/hooks/distribution-preflight-check.bats` gains one test to the existing suite: the hook is driven for real, with a non-ASCII path committed on the feature branch and reported as `missing`.
- `.gaia/scripts/tests/distribution-audit-pr-gate.bats` is new. It **executes the gate step's real `run:` body**, extracted from the workflow YAML, against a sandbox repository with a mocked `gaia-maintainer`, the technique `.github/audit/tests/ci-status-member-gate.bats` already uses on `code-review-audit.yml`. Three of its five tests exist to make the fourth mean something: they prove the harness reaches a real verdict (fails on an unanswered file, passes on an answered one, passes on an inherited backlog it did not touch). The fifth proves the YAML extraction guard fires rather than silently capturing an empty body, which would read as a passing gate.

`core.quotePath` is pinned to `true` in both fixtures. It is git's default and is the condition under test, so a maintainer who turns it off in `~/.gitconfig` would otherwise green against the bug itself.

All four guards were mutation-proved: reverting the hook fix reds its test; an all-ASCII fixture fires each non-vacuity assertion by name; breaking the extraction anchor fires the extraction guard.

`.github/workflows/audit-ci-tests.yml` gains one paths-filter entry naming the new suite's *source*, `distribution-audit-pr.yml`. It is redundant with the `.github/workflows/**` catch-all already in that output, and is listed deliberately anyway as documentation of the test-to-source pairing, so narrowing the catch-all later cannot silently drop a guarded source. See the round-1 dispositions below: this entry shipped with a comment claiming it was load-bearing, which the audit correctly caught.

## Verification

- `bash .gaia/tests/shell-lint.sh` clean
- `bats .gaia/scripts/tests/` 1448 tests, zero failures
- `bats .gaia/tests/hooks/` zero failures
- The Quality Gate **skips by its own rule**: no staged file is source or gate-affecting config (`.sh`, `.bats`, `.yml`, `.md` only). Verified with the page's own grep against `main...HEAD`.

## Deliberately left alone

- **#1165** cites `:269` in the same hook file, one line above the fixed call, for an unrelated defect (the pre-flight mirrors only the missing-intersection arm, not the region-drift arm). It is scheduled as its own row and is a design choice, not a mechanical fix; batching the two would be exactly the mechanical-with-fork pairing the drain plan forbids.
- **#1229** is the reason neither of these call sites was caught by a check: `check-audit-base-derivation.sh`'s assertion 4 scans `.claude/agents/` only and can see no `.sh` or `.yml` site. Already filed, already routed, and its body already counts this pair as the class's fourth instance, so this PR does not advance that count.
- `.claude/hooks/pr-merge-audit-check.sh:485` and `:533` carry the same un-`-z` spelling. They were examined under #1213 and are not defects: a quoted path there is neither allowlisted nor classified in scope, so both fail closed.

## Round-1 audit dispositions

Both dispatched members (`code-audit-github-workflows`, `code-audit-maintainer-shell`) cleared with earned markers. Zero Critical, zero Important. Three Suggestions:

- **Fixed in place.** The paths-filter entry this PR added was a strict no-op, and its comment asserted the opposite as fact: the `code:` output already carries a `.github/workflows/**` catch-all and `dorny/paths-filter` ORs within an output. Verified by parsing the filter rather than by reading the finding. The entry is kept and its comment made honest, the same convention the `sandbox:` filter in the same file uses for its own redundant lines, so the test-to-source pairing stays documented if the catch-all is ever narrowed.
- **Fixed in place.** The new suite's `setup()` preconditions called `skip`, so a CI runner that lost `jq` would have reported `ok … # skip` five times and greened the job having run zero assertions against a required check's gate step. They now fail on CI and still skip locally, matching `require_repo_path` / `require_yaml_parser` in `retrigger-reachability.bats`. Both arms proved: skips off CI, reds on CI with its own diagnostic.
- **Accepted and noted.** The extracted `run:` body writes hardcoded `/tmp/pr-changed.txt` and `/tmp/missing.txt`. Running the shipped body unmodified is the suite's whole point, and the durable fix (`"${RUNNER_TEMP:-/tmp}"`) is a change to the gate step itself, which this PR is otherwise only encoding-fixing. The suite header discloses both filenames, and CI runs bats serially so the names cannot race there.

Filed as **#1238**: the same false load-bearing comment sits on two pre-existing entries in the same filter (`:95`, `:105`). A waive was licensed, since it is a non-security finding on a file this PR already changes, but by then the class had been observed on three entries including the one this PR added, so it is filed rather than recorded here.

## Round-2 audit dispositions

Both members re-dispatched (the round-1 repairs rotated both digests) and both cleared clean. Zero Critical, zero Important; the workflows member returned zero Suggestions. The shell member drove the new `require()` helper in isolation on bash 3.2.57 and bash 5.3.15 and end-to-end through a doctored `REPO_ROOT`, confirming both arms behave identically across versions.

Two Suggestions, both on prose the round-1 repair itself introduced, both **accepted and noted**:

- The new header comment says the runner "installs all three in the same job"; that job's apt install is `bats python3-yaml zsh`, so `jq` and `git` actually come from the ubuntu image. Confirmed against `audit-ci-tests.yml:419`. The guard's conclusion is unaffected (both are preinstalled, so an absent one still means something is wrong rather than the box being lean), but the attribution in the diagnostic is imprecise.
- `require()` has no self-test of its own two arms, where the sibling `retrigger-reachability.bats` self-tests its equivalents.

Neither is folded in. This PR is clean and both members are marked, so folding a comment clause and a helper self-test in buys a full two-member re-dispatch purely to re-earn the markers, and each round has now landed on the prose the previous round wrote rather than on the fix. Both are cheap for whoever next touches this suite.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.gMlLkArkIQ

PR #1240, opened 2026-08-06
debt(pricing): validate rate-overlay window contents, not just the array type (#1199)

### LINKED ISSUE #1199: The rate-overlay shape check validates only one level, so an empty or non-object window array is announced as an active overlay while pricing nothing
**Location:** `.gaia/scripts/token-pricing-lib.sh:137` (`gaia_apply_rate_overlay`'s overlay shape check)

## Failure mode

The check that validates the machine-local rate overlay reads:

```jq
(.models | type) == "object"
and ([.models[] | type] | all(. == "array"))
```

`all(. == "array")` accepts **any** array, so two operator-plausible overlays pass validation and then reproduce the outcome the check's own comment says is closed:

- `{"models":{"claude-sonnet-4-6":[]}}` merges. `token-tally.sh` then prints `(local rate overlay active for: claude-sonnet-4-6)` on the same readout as `(lower bound: unpriced model(s) claude-sonnet-4-6)`: the overlay is announced as working while pricing nothing.
- `{"models":{"claude-brand-new-1":[7,35]}}` merges, then `rate_window` raises `Cannot index number with "effective_through"`. `token-rollup.sh` reports `dollar pricing computation failed` followed by `Est. cost (USD): unavailable (records predate per-model attribution)` — a reason that is false and that never names the overlay the operator just hand-wrote.

Both were reproduced through the real consumers in a scratch sandbox. Neither is a regression: both shapes behaved identically before the overlay's shape check existed, and neither emits a confidently-wrong dollar figure, which is why this is a suggestion rather than a defect that blocked its pull request.

## Suggested fix

Validate the window array's contents, not just its type:

```jq
(.models | type) == "object"
and ([.models[]] | all((type == "array") and (length > 0) and (all(.[]; type == "object"))))
```

Verified against every shape the existing suite exercises: it refuses both leaks plus everything already refused, and still admits `{"models":{}}` and the multi-window fixture.

**Tripwire, do not skip this when fixing.** Do **not** tighten further to require numeric `input`/`output`. `token-tally.sh`'s `print_unpriced_remedy` deliberately emits those as **strings** (`"<usd-per-1M-input>"`) so an unedited paste degrades `dollars` to `null` instead of silently pricing the model at zero, and `.gaia/scripts/tests/token-rate-overlay.bats` depends on that placeholder merging and then failing at arithmetic time. A numeric-rate assertion would break that guarantee.

## Provenance

Raised by `code-audit-maintainer-shell` in PR #1198 round 2 and graded a Suggestion by the member itself. Accepted-and-noted rather than fixed, under a stopping rule written before that round ran: the finding lands on the repair round 1 had just made, which is the runaway shape where each widening of a validation invites the next. Recorded in PR #1198's body and filed here so it is tracked rather than only recorded.

Handler: prompt

stevensacks:
## What

`gaia_apply_rate_overlay`'s machine-local rate-overlay shape check validated only one level:

```jq
(.models | type) == "object" and ([.models[] | type] | all(. == "array"))
```

`all(. == "array")` accepts **any** array, so two operator-plausible overlays got past it, merged, and were then announced as active while contributing nothing usable:

- `{"models":{"m":[]}}` — a model key replaces that model's **whole** window array, so `[]` leaves `rate_window` nothing to select. The model prices at zero on a readout carrying `(local rate overlay active for: m)` and `(lower bound: unpriced model(s) m)` on the same block.
- `{"models":{"m":[7,35]}}` — the bare rate pair an operator reaches for. It merges cleanly and then raises `Cannot index number with "effective_through"` inside `rate_window`, so the consumer reports `dollar pricing computation failed` followed by a reason that is false and that never names the overlay just hand-written.

Neither is a regression: both shapes behaved identically before the shape check existed, and neither emits a confidently-wrong dollar figure.

## Fix

Validate the window array's **contents**, not just its type:

```jq
(.models | type) == "object"
and ([.models[]] | all((type == "array") and (length > 0) and (all(.[]; type == "object"))))
```

`jq`'s `and` short-circuits, so `.[]` is never applied to a non-array. The refusal message now names what is actually required (`an object whose every value is a non-empty array of window objects`) instead of the looser `an object of window arrays`, which a rejected operator would have read as describing their own file.

## The tripwire, honored

The check deliberately stops short of requiring numeric `input`/`output`. `token-tally.sh`'s `print_unpriced_remedy` emits those as **strings** (`"<usd-per-1M-input>"`) on purpose, so an unedited paste degrades `dollars` to `null` rather than silently pricing the model at zero. A new test (`a window array of hand-written placeholder strings still merges`) pins that entry merging, and the pre-existing end-to-end test that pastes the real remedy and asserts `dollars: null` would fail on any over-tightening.

## Verification

Two RED tests written first, both failing on the old check for exactly the two leak shapes, then green after the fix:

- `.gaia/scripts/tests/token-rate-overlay.bats` — 29/29 (3 new)
- every other suite consuming the pricing lib or the overlay (`token-price`, `token-rates-table`, `token-rollup`, `token-tally`, `token-tally-price`, `token-cost-e2e`, `token-command`, `token-review`, `cost-reprice`, `cost-unpriced-scan`, `spec032-e2e`, `token-rollup-merge`, `token-tally-git-op`) — 219/219, no failures

All bats runs under bash 5 via `.gaia/scripts/bats5.sh`, so local matches CI.

The Quality Gate does not apply: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.

Closes #1199

## Accepted residuals (audit Suggestions, not addressed)

`code-audit-maintainer-shell` passed clean with no Critical or Important findings and raised two Suggestions, both accepted-and-noted under the digest economics in `wiki/concepts/PR Merge Workflow.md`: the PR is clean and the member is already marked, so folding either in buys a full re-dispatch of that member solely to re-earn the marker. The member's own verdict was that neither needs addressing.

1. **The tightened check still admits a window object missing `input`/`output`** (`.gaia/scripts/token-pricing-lib.sh:152`). `[{"input":7}]` and `[{}]` satisfy all three conditions, merge, and raise `number (0) and null (null) cannot be multiplied` in `priced_row`. This is deliberate, not a gap: `token-tally.sh` swallows that raise and lands on `dollars: null`, the same accepted degrade the placeholder tripwire preserves, materially unlike the empty-array and `[7,35]` cases which produce a confident zero or a bare failure. The residual is comment accuracy only: bullet 3 justifies refusing `[7,35]` on the grounds that it raises at pricing time, and raising alone does not separate it from the shape the very next paragraph protects. The tripwire paragraph that follows is imperative and unambiguous on its own, so the imprecision misleads nobody who reads to the end of the comment block.

2. **SC2016 at `.gaia/scripts/tests/token-rate-overlay.bats:451`**, pre-existing and untouched by this diff, outside its changed line ranges. The single quotes are correct (`$7.00` is the literal needle) and `shell-lint.sh` runs bats at `severity=warning`, so the info-level hit does not gate. Recorded for traceability, not filed: it sits on a file this PR already changes.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.2frbUSQbKc

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.diA6QRjlYL

PR #1241, opened 2026-08-06
debt(cli): match a language-file value by the quote it opened with (#1197)

### LINKED ISSUE #1197: gaia init rename silently skips a language-file value wrapped in the quote its own pattern excludes
`.gaia/cli/src/init/rename.ts:266`, with the same content class at `:284` and `:299`.

## Failure mode

All three rewrite patterns spell the literal's body as `(?:[^'"\\]|\\.)*`, a class that excludes **both** quote characters. A perfectly valid literal wrapped in one quote and holding the other unescaped therefore never matches, and the rewriter silently does nothing:

- A project whose `app/languages/en/common.ts` has diverged to `siteName: "Steve's Template"` comes back **byte-unchanged** from `gaia init rename`, which exits 0 and records the step completed. Only the single-quoted keys in the same file are rewritten.
- The same class cannot re-match the command's own output when the title carries the non-wrapping quote. `--title 'Say "Hi" App'` into a single-quoted seed correctly writes `'Say "Hi" App'` (the escaper deliberately leaves the non-wrapping quote bare, since it needs no escape), but a **later** rename to a different title then leaves that key holding the old title, again at exit 0.

Both arms are the silent-no-op-at-exit-0 class this file has already been fixed for once, in #1176, where the missing `CLAUDE.md` heading made `renameClaudeMd` return as though it had succeeded.

Pre-existing rather than introduced: #1193's fix (PR #1196) emits the identical bytes here and does not touch the content class. It was left out of that PR deliberately, because it carries a design decision and #1196 is a mechanical escaping fix.

## Suggested fix

Two reasonable answers, and picking between them is the decision this issue carries:

1. **Build the content class from the quote actually captured**, so a literal wrapped in `'` may contain a bare `"` and vice versa. Closes both arms, and changes what the three patterns match, so it needs its own adversarial fixtures.
2. **Report a key the rewriter could not match**, so the step fails loudly instead of exiting 0 with the file untouched. This is the answer #1176 settled for this module's other precondition, and it generalizes: `rename` enforces preconditions rather than guessing around them.

The module's own precedent leans to 2, or to 2 plus 1.

Handler: prompt

stevensacks:
Closes #1197

## What was wrong

`gaia init rename` spells the body of every string literal it rewrites as a class excluding **both** quote characters. A perfectly valid literal wrapped in one quote and holding the other unescaped therefore never matches, and the rewriter silently does nothing at exit 0 with the step recorded as completed.

Two arms:

- A project whose `app/languages/en/common.ts` has diverged to `siteName: "Steve's Template"` comes back byte-unchanged.
- The class cannot re-match this command's own output. `--title 'Say "Hi" App'` into a single-quoted seed correctly writes `'Say "Hi" App'` (the escaper leaves the non-wrapping quote bare, since it needs no escape), but a later rename to a different title then leaves the key holding the old one.

## The fix

The issue named two answers and observed the module's precedent leans to the second, or to both. Both are here.

1. **The content class is built from the quote the match actually opened with** (`(?!\2)` rather than a class naming both quotes), so a single-quoted literal may hold a bare `"` and a double-quoted one a bare `'`. `\n` stays excluded, which bounds the body to one line now that it no longer stops at the other quote character.
2. **A key that is present but holds no rewritable literal fails the step by name** (`language_value_not_rewritable`, exit 1) before any file is written, the same precondition shape `#1176` settled for the missing `CLAUDE.md` heading. An absent key is still tolerated: the shipped `_index.ts` carries only `meta.title`.

The three rewrite patterns and the precondition's probe now read one table, so the check that a key *was* rewritten cannot drift from the rewrite meant to do it. `literalPattern` owns both capture-group indices, so a prefix can no longer shift the numbering the backreference and the replacer depend on.

## Verification

- Both guards mutation-tested. Restoring the both-quotes class fails exactly the two new quote tests; neutering the precondition fails exactly the three new refusal tests.
- `.gaia/tests/distribution/08-gaia-init-cli-sequence.sh` (the adopter `/gaia-init` sequence against the real staged tree) passes, confirming the new precondition does not refuse the shipped seed.
- The rebuilt bundle was smoke-tested end to end on all three behaviors.
- Full Quality Gate green: CLI suite (131 files), app typecheck / lint / unit / build / Playwright, `shell-lint`.

## Audit rounds

Three rounds of `code-audit-maintainer-node`. Round 1 and 2 each raised one Important finding, both about the same thing and both repaired:

1. The new precondition turns a shipped-seed value drifting out of quoted-literal shape from a silent no-op into an exit 1 for **every** adopter scaffold, and every test writes its own fixture, so nothing read the files that actually ship. Added an invariant over the real files.
2. That invariant asked the precondition whether the shipped files were fine, and the precondition reports only a key it can *see* but cannot rewrite, so it is silent on a key it cannot see at all: reindenting `_index.ts` moves `meta.title` out of the anchored pattern, and the assertion passed over a file nothing would rewrite. The invariant now copies the real files into a sandbox, runs the step, and reads the values back, which fails on indentation drift, on the file going missing, and on a non-rewritable value alike. All three mutation-proven.

Round 3 came back clean with two Suggestions, both accepted and noted here rather than folded in, per the digest economics in `wiki/concepts/PR Merge Workflow.md`: the PR is clean and the member is marked, so applying either would buy a full re-dispatch solely to re-earn the marker, and neither is consequential.

- The invariant reads back only the two keys today's seed carries (`siteName`, `meta.title`), not the never-seeded `heroTitle` / top-level `title` the rewrite table also owns. If a future seed adds one at an indentation the anchored prefix misses, that key would keep GAIA's value with the test green.
- The invariant's sandbox scaffolding (`package.json`, `CLAUDE.md`, the `app/languages/en/pages` mkdir) re-implements `setupSandbox` rather than calling it, so a future precondition file added to `setupSandbox` would not reach it.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.V1chQx7lSx

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.z3JZGBMcVO

Issue #1008, opened 2026-07-22
.husky/** has no Code Audit Team owner, so the hook gating every commit merges unaudited


stevensacks:
**Location:** `.gaia/audit-ci.yml` (the `auditors:` roster)

## Failure mode

No auditor's globs match `.husky/**`, including the default member's. A PR that changes `.husky/pre-commit` alone therefore dispatches no Code Audit Team member for that file, and the merge gate clears it via the out-of-scope bypass (`.husky/` is not among the in-scope prefixes `pr-merge-audit-check.sh` enforces).

Verified empirically on the #994 fix branch: `bash .gaia/scripts/resolve-audit-members.sh` named only `code-audit-github-workflows` (for the changed workflow YAML) and `code-audit-maintainer-shell` (for the changed `.bats` suite). The changed `.husky/pre-commit` itself pulled in no member. The hook's own change was reviewed only because the PR happened to touch two other owned surfaces.

## Blast radius

`.husky/pre-commit` is the Quality Gate floor for every developer and is manifest class `owned`, so it ships to every adopter. It is shell, it is security-relevant (it is what `block-no-verify.sh` exists to stop people bypassing), and a change to it can silently disable typecheck, lint, and tests for a whole class of commits. #994 is a worked example of that class reaching `main` unnoticed.

Note the shape: this is the same defect as #994 one level up. A change-detection allowlist silently omitted a directory, so work in that directory skipped its gate.

## Suggested fix

Add `.husky/**` to a member's globs in `.gaia/audit-ci.yml`. `code-audit-maintainer-shell` is the natural owner: the file is POSIX shell and that member already holds the shell lens and the shellcheck oracle.

Two things make this more than a one-line edit, hence the `plan` handler:

- Member remits are generated from the roster, so `bash .gaia/scripts/write-audit-remits.sh` must run and its output land in the same change, or the roster check fails.
- `.gaia/audit-ci.yml` is gate machinery, so editing it rotates every member's content digest and invalidates every in-flight marker. Expect a full-roster re-dispatch on the PR that does it.

Also consider whether `.husky/` belongs in the in-scope prefix list `pr-merge-audit-check.sh` uses, so the out-of-scope bypass cannot clear a husky-only diff even if the roster later drifts again.

Surfaced as a cross-remit finding by `code-audit-github-workflows` while auditing the #994 fix.

Handler: plan

Issue #806, opened 2026-07-16
tech-debt: workflow-YAML-only diffs route to code-audit-frontend, not maintainer-shell


stevensacks:
**Location:** `.gaia/audit-ci.yml:66` (the `code-audit-maintainer-shell` glob block, lines 61-67); mirrored at `.claude/hooks/lib/audit-scope.sh:129` (the builtin fallback block, lines 123-130).

**Failure mode.** `code-audit-maintainer-shell` claims `.github/**/*.sh` and `.github/**/*.bats` but no `.yml`, and `code-audit-maintainer-node` claims only `.gaia/cli/src/**`. So a PR that changes only a live workflow under `.github/workflows/*.yml` matches no specialized member and falls through to the default member (`code-audit-frontend`) via `audit_in_auditable_base` (`.claude/hooks/lib/audit-scope.sh:44`, the `.github/workflows/*` arm). CI and hook-contract YAML then gets reviewed by the React specialist, which is the wrong lens. The same content in its template form (`.gaia/cli/templates/workflows/code-review-audit.yml.tmpl`) routes correctly to maintainer-shell, so editing the rendered `.yml` versus the `.tmpl` splits auditors for byte-identical content.

Note: SPEC-044 folded `code-review-audit.yml` into the machinery digest set, so a workflow-YAML change now invalidates every member's marker. That is keying, not dispatch; the wrong-specialist routing still stands.

**Suggested fix (verify before implementing).** Add `.github/**/*.yml` to `code-audit-maintainer-shell`'s globs in both `.gaia/audit-ci.yml` (inside the maintainer-only marker block) and the mirrored builtin fallback in `.claude/hooks/lib/audit-scope.sh`, which must stay in lockstep (`audit-scope.sh:100-102`), plus a bats case asserting a workflow-YAML-only diff resolves to maintainer-shell. Before implementing, confirm the new glob does not collide with the existing `.github/**/*.sh` and `.github/**/*.bats` claims, and that adopter repos (whose `.github/workflows/` is legitimately frontend-owned) stay unaffected because the change lives inside the maintainer-only marker.

Dedup note: distinct from #730 (that issue is the `code-review-audit.yml` live-versus-source drift guard, a different concern).

Handler: plan

stevensacks:
[...]

PR #1243, opened 2026-08-06
debt(audit): give the CLI workspace policy file a Code Audit Team owner (#1207)

### LINKED ISSUE #1207: .gaia/cli/pnpm-workspace.yaml has no Code Audit Team owner, so a supply-chain policy change there merges unaudited
`.gaia/audit-ci.yml:78` lists `pnpm-workspace.yaml` as a root-relative literal, which never crosses a `/`, so it matches the repository-root file and nothing else. `.gaia/cli/pnpm-workspace.yaml` is matched by **no** Code Audit Team member: the maintainer-node claimant's list stops at `package.json`, `pnpm-lock.yaml`, and `tsconfig*.json`.

**Verified mechanically rather than read off the globs.** Sourcing `.claude/hooks/lib/audit-scope.sh` and calling `audit_owner_for_path`:

- `.gaia/cli/pnpm-workspace.yaml` → `[]`
- `pnpm-workspace.yaml` → `[code-audit-frontend]`

**Failure mode.** As of #1152, that unowned file carries the `.gaia/cli` workspace's entire supply-chain policy: `minimumReleaseAge`, `trustPolicy`, and both exclusion lists. A pull request whose only change is lowering `minimumReleaseAge` there, dropping `trustPolicy`, or adding an exclusion entry resolves an **empty dispatched member set**, so no member is dispatched, no marker is required, and it merges with no review. That is the fail-open direction: the gate reports nothing to do rather than reporting a gap.

It reached review at all only because the pull request that introduced the policy also touched two owned files. A follow-up that touches the workspace file alone would not.

**This is a recurrence, not a novel shape.** #1008 was the same defect for `.husky/**` (*"has no Code Audit Team owner, so the hook gating every commit merges unaudited"*), and #806 was a routing variant on the same file. Both are closed. The ownership table has now been wrong in three distinct places, which suggests the real gap is that nothing checks the table for uncovered security-relevant paths, not that three globs were forgotten.

**A partial mitigation now exists and is worth knowing about, because it is easy to overread.** `.gaia/cli/src/lint-pin-parity.test.ts` asserts a floor on both scalars and containment on both exclusion lists, so a weakening of that file fails `Vitest (.gaia/cli)`, a declared-required context, and `pnpm-workspace.yaml` is in that job's paths filter. So a weakening is not *unguarded*. But that is a test, not an audit: it catches the four values it names and nothing else about the file, and it cannot review an added key, a changed `allowBuilds` entry, or a comment that misstates the policy. The audit-ownership gap is still real.

**Suggested fix.** Add `.gaia/cli/pnpm-workspace.yaml` to the maintainer-node globs in `.gaia/audit-ci.yml` and to the `gaia:audit-remit` block in `.claude/agents/code-audit-maintainer-node.md`, following #1008's own repair. Note the cost, so it is not a surprise: `.gaia/audit-ci.yml` is gate machinery, so editing it rotates **every** member's content digest and forces a full re-dispatch of the whole roster on whatever pull request carries the change.

Worth considering beyond the one glob, given three recurrences: a check that every path a required job gates on, or that carries security-relevant configuration, resolves to at least one owner. That would catch the fourth instance instead of this issue's author catching it.

Handler: prompt

stevensacks:
`.gaia/audit-ci.yml` listed `pnpm-workspace.yaml` as a root-relative literal, which never crosses a `/`, so it matched the repository-root file and reached `.gaia/cli/pnpm-workspace.yaml` not at all. No other member's globs claimed it either: the `code-audit-maintainer-node` entry stopped at `package.json`, `pnpm-lock.yaml`, and `tsconfig*.json`.

That file carries the `.gaia/cli` workspace's entire supply-chain policy (`minimumReleaseAge`, `trustPolicy`, and both exclusion lists), so a PR whose only change lowered a scalar, dropped `trustPolicy`, or added an exclusion entry resolved an **empty dispatched member set**: no member ran, no marker was required, and it merged with nobody responsible for it. Verified before the fix with `audit_owner_for_path`, which returned `[]` for the CLI workspace file and `[code-audit-frontend]` for the root one.

## What changed

- **Both rosters** claim `.gaia/cli/pnpm-workspace.yaml` for `code-audit-maintainer-node`: the committed `.gaia/audit-ci.yml` and the builtin fallback in `.claude/hooks/lib/audit-scope.sh` that the degraded gate uses when the config is absent. A glob in one but not the other leaves the file ownerless in the fallback.
- **The member's remit region** is regenerated from the roster with `bash .gaia/scripts/write-audit-remits.sh`, not hand-edited.
- **Three pins**, at three different levels: the classifier-level ownership assertion (under both rosters, plus an assertion that the root file stays with the default member), the dispatch-level resolver test for a workspace-file-only diff, and the routing-parity suite's deliberate-move exception set.
- The wiki roster table and the CHANGELOG record it.

## Verification

Each new guard was proven to fail: with the glob removed from both rosters, the classifier pin, the dispatch test, and the routing-parity suite all go red (`ROUTING MISMATCH: .gaia/cli/pnpm-workspace.yaml before=- after=- expected=code-audit-maintainer-node`); restored, all three go green.

Green: the full `.gaia/tests/hooks/` suite, `resolve-audit-members.bats`, `verify-audit-roster.bats`, `write-audit-remits.bats`, `resolve-audit-spawn.bats`, `.gaia/tests/distribution/16-audit-remit-parity.sh`, and `bash .gaia/tests/shell-lint.sh`. The Quality Gate skips: no staged source or gate-affecting config.

## Cost, stated up front

`.gaia/audit-ci.yml` is gate machinery, so this diff rotates **every** member's content digest and forces a full roster re-dispatch on whatever PR carries it.

## Not in scope

The issue notes this is the third recurrence of the same shape (#1008 for `.husky/**`, #806 a routing variant), and suggests a deterministic check that every path a required job gates on, or that carries security-relevant configuration, resolves to at least one owner. That is a systemic guard rather than this issue's one glob, and it wants its own design; filing it separately keeps this PR the surgical fix the issue asked for.

Closes #1207

- `.gaia/audit-ci.yml:132` — three sibling `.gaia/cli` config files stay ownerless after this glob addition. The default member's `*.config.ts` / `*.config.mjs` globs are root-anchored and `code-audit-maintainer-node` claims only `.gaia/cli/src/**/*.ts`, so a PR whose only change is `.gaia/cli/vitest.config.ts`, `.gaia/cli/eslint.config.mjs`, or `.gaia/cli/prettier.config.mjs` resolves an empty dispatched set, falls to the merge gate's legacy branch, is out-of-scope-allowlisted by the `.gaia/*` arm, and merges with no member responsible for it. Pre-existing, independently reconfirmed by the orchestrator (all three classify to `[]`).
  `v1 class=holistic/unclassified path=.gaia/audit-ci.yml line=132`


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.z7o7KMYwOX

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.CeMOvUqIDJ

PR #1246, opened 2026-08-06
debt(deps): state release-age strict enforcement instead of inheriting it (#1208)

### LINKED ISSUE #1208: minimumReleaseAgeStrict defaults off, so the 7-day window admits an immature version and silently writes its own exemption
Neither workspace sets `minimumReleaseAgeStrict`, and its default is the **weak** value, so the 7-day release-age window is self-waiving: an install that wants an immature version gets it, and pnpm silently records the waiver in `pnpm-workspace.yaml`.

**Failure mode**, read out of pnpm 11.9.0's own pipeline rather than inferred from the docs:

```js
const strictMode = opts3.minimumReleaseAgeStrict === true;
handleResolutionPolicyViolations: async (violations) => {
  if (!strictMode) return;          // the install does not fail
  ...
},
pickManifestUpdates: (violations) => {        // NOT gated on strictMode
  const entries = pickImmatureEntries(violations, strictMode);
  return entries ? { addedMinimumReleaseAgeExcludes: entries } : void 0;
}
```

With strict mode off, a resolution that violates the window does not fail the install. Worse, the manifest-update path is **not** gated on strict mode at all: `updateWorkspaceManifest` merges the offending `name@version` into `minimumReleaseAgeExclude` and writes the file back, on an ordinary `install`, `add`, `update`, or `audit --fix` (guarded only by `opts.save !== false && !opts.dryRun`).

So the window degrades from "block an immature version" to "admit it and append a permanent exemption for it", with no prompt. `minimumReleaseAge: 10080` stays in the file and stays true, and every existing check stays green.

**Why this is not covered by what already exists.** `.gaia/cli/src/lint-pin-parity.test.ts` asserts a floor on both scalars, containment between the two exclusion lists, and that every exemption entry names an exact package rather than a pattern. None of those fire here: the auto-written entries are exact `name@version`, which is precisely the shape the exactness predicate exists to permit. The containment test catches the `.gaia/cli`-side case only, since an entry auto-appended there that root lacks breaks containment; the **root-side** case is invisible, and root is where the app's own dependencies resolve.

This is the same silence `#1152` was filed about, one layer in: the setting is present and correct, and the thing it is supposed to prevent happens anyway.

**Scope note, recorded so this is not re-derived.** This was deliberately not folded into PR #1206, on the audit member's own recommendation. It is a configuration decision plus a new assertion, not a repair to the guard that pull request builds, and the decision has a real cost to weigh: strict mode turns a silent admission into a hard install failure that a maintainer must then resolve by hand, which is the correct trade for a supply-chain control but is a behaviour change for anyone bumping a dependency.

It is also **not** a peer of the two settings deferred alongside it. `trustPolicyIgnoreAfter` and `ignoreMissingTimeField` are both absent-means-hardened, so leaving them unset is safe; this one is the only one whose default already sits in the weak position.

**Suggested fix.** Set `minimumReleaseAgeStrict: true` in both `pnpm-workspace.yaml` and `.gaia/cli/pnpm-workspace.yaml`, and extend the hardening-parity block in `.gaia/cli/src/lint-pin-parity.test.ts` to assert it on both sides, the same way `trustPolicy` is asserted. Before doing so, decide what a maintainer should do when an install is blocked, and write that into the comment beside the setting, since the whole point is that the escape hatch stops being automatic.

Worth checking while there: whether any entry currently in either `minimumReleaseAgeExclude` was auto-appended by this path rather than added deliberately.

Handler: prompt

stevensacks:
Closes #1208

## What

Both workspaces set `minimumReleaseAge: 10080`, but neither stated what a breach of that window should *do*, leaving the consequence to a pnpm default. This sets `minimumReleaseAgeStrict: true` in both `pnpm-workspace.yaml` files and extends the existing hardening-parity guard in `.gaia/cli/src/lint-pin-parity.test.ts` to assert it on each side.

## The filed premise does not hold, and that changes the risk

The issue reasoned that the default is the weak value, so the window is self-waiving. Measured against the pinned pnpm (11.9.0) rather than inferred:

- **`minimumReleaseAge` set, strict key absent** (what both workspaces do today): pnpm's config reader turns strict mode on by itself. A sandbox workspace at this floor refuses a same-day version with `ERR_PNPM_NO_MATURE_MATCHING_VERSION` and writes no exemption.
- **strict explicitly `false`**: the failure mode the issue describes reproduces exactly. The install succeeds and pnpm appends eight `name@version` entries to `minimumReleaseAgeExclude`, rewriting the workspace file.

That inference has shipped alongside the setting since pnpm 11.0.8 (10.x has neither), so **this change is behaviour-neutral today**. It is not the behaviour change the issue warned about: maintainers bumping a dependency already get a hard failure, and they get the same one after this lands.

What it buys is that the behaviour no longer rests on a default in pnpm's config reader that nothing here would notice going away, and that weakening it now takes a deliberate edit a reviewer sees.

## Adopter propagation

`pnpm-workspace.yaml` is manifest class `shared`, and `/update-gaia` merges it field-aware over `MANAGED_WHOLE_VALUE_KEYS` only. Without registering the key there, the policy would have held in this repository but never reached the product it ships. The key joins that list, the two committed CLI bundles are rebuilt, and the prose enumerations of the hardening settings that are now incomplete are refreshed (`wiki/decisions/pnpm.md`, `.claude/skills/update-gaia/SKILL.md`, `cli-tests.yml`, `audit-scope-lib.bats`).

## The issue's open question

> Worth checking while there: whether any entry currently in either `minimumReleaseAgeExclude` was auto-appended by this path rather than added deliberately.

No. Every entry in both lists traces to a deliberate human commit with a descriptive message (`bippy@0.5.42` to #442, `@gaia-react/lint` to #312 / #592 / #1206), and each carries its hand-written rationale comment. The auto-append path can only fire with strict explicitly false, which neither workspace has ever set.

## Verification

The guard was proven to fail before it was made to pass, both ways: **absent** key reds two tests, and **`false`** reds two tests. Full Quality Gate green (typecheck, lint, 1989 CLI tests, 156 root tests, 8 E2E, dev smoke HTTP 200, build), plus the touched bats suite (28/28) and `shell-lint`. Both workspaces re-install clean under `--frozen-lockfile` with lockfiles untouched.

## Deliberately not done

The parity guard enumerates each hardening setting across five sites, so adding one setting meant five edits. Collapsing that to a table driving all three tests is a real improvement but is a refactor of code this fix does not otherwise touch.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.1h4ncda7F5

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.QXGgRVVEQn

PR #1247, opened 2026-08-06
debt(hooks): scope the distribution pre-flight's mirror claim to the arm it mirrors (#1165)

### LINKED ISSUE #1165: Local distribution pre-flight mirrors only the missing-intersection arm, so region drift still surfaces as a red check after push

stevensacks:
## What

`.claude/hooks/distribution-preflight-check.sh` mirrors exactly one of `distribution-audit-pr.yml`'s two failure conditions: the `missing`-intersection arm. Two comments claimed it mirrored the workflow outright:

- `:269` "Matches distribution-audit-pr.yml exactly so the two gates never disagree."
- `:313` "This gate mirrors .github/workflows/distribution-audit-pr.yml deliberately."

Both read as completeness claims, so a maintainer debugging a red `Distribution Audit` that passed the local gate reads them and concludes CI misfired.

## Why direction 2 (scope the comments) and not direction 1 (add the region arm)

The issue framed the choice as a real one. Adding the arm loses:

- `missing` is intersected with the branch's **committed** changed set, and the file map behind it is a `git ls-files` key set, which an uncommitted edit cannot move. It answers from the state CI audits.
- `.regionDrift` is built by `readFileSync`-ing each shipped file's **content** off disk (`.gaia/cli/src/release/region-scan.ts`) and diffing marker-bearing paths against the committed declaration, and CI deliberately leaves it unscoped. Evaluated in the hook it would be the first arm that denies what CI would pass, on a working tree the PR never contains, and the deny's remedy ("regenerate the manifest") would be wrong advice.

That breaks the header's standing claim that the hook is only ever a cheaper way to find out what CI would have told you. Narrowing the condition to the changed set instead would enforce a different rule than CI's, which that workflow's own inline comment explicitly rules out.

So the gap is defensible; the comments were not. The change scopes both claims to the unanswered-file rule and documents the omitted condition and its reasoning in the header.

## Verification

- Comments only, no executable line touched.
- `.gaia/tests/hooks/distribution-preflight-check.bats` under bash 5: 38/38 pass.
- `bash .gaia/tests/shell-lint.sh`: passes.
- Quality Gate skipped per `wiki/decisions/Quality Gate.md` (no staged source or gate-affecting config; shell-only diff).
- No CHANGELOG entry: the hook is release-excluded, so no adopter-facing surface changes.

Closes #1165

## Audit rounds

Round 1 raised three Suggestions; two were real defects in the same header block and are fixed in `0427b629`:

- The `ADOPTER POSTURE` block claimed the registration lives only in the gitignored `.claude/settings.local.json` and that no scrub path could strip it from `.claude/settings.json`. Both stale: the registration is committed at `.claude/settings.json:326`, and `.gaia/release-scrub.yml`'s `json-strip-array-element` selector removes exactly that element before tar.
- My own added missing-arm justification claimed a `git ls-files` key set cannot be moved by an uncommitted edit. `ls-files` reads the index, so a staged `git add` of a new file does reach `missing`; the three-dot intersection is what confines the arm to committed state. Re-attributed, with an explicit warning not to drop the intersection.

The third (shellcheck `SC1091`) is pre-existing and excluded repo-wide by `.gaia/tests/shell-lint.sh`'s curated `TOOLING_EXCLUDE`; no fix owed.

Round 2 verified both repairs against the machinery and passed clean.

### Accepted residual

Round 2 left one cosmetic Suggestion, recorded here rather than folded in, per the digest economics in `wiki/concepts/PR Merge Workflow.md`: the PR is clean and the member is already marked, so a one-line rewrap would buy a full re-dispatch solely to re-earn the marker, and the finding is on prose the previous round's own repair wrote.

- `.claude/hooks/distribution-preflight-check.sh:82` wraps the identifier `json-strip-array-element` mid-token across a comment line break, so a `git grep` sweep for consumers of that scrub-rule type misses this file. No live failure mode.

### DIFF
--- a/.claude/hooks/distribution-preflight-check.sh
+++ b/.claude/hooks/distribution-preflight-check.sh
@@ -24,6 +24,33 @@
 # reads to a later audit as accidental drift. See wiki/decisions/Deliberate
 # Configuration Asymmetries.md for the sibling cases.
 #
+# WHAT THIS DOES NOT MIRROR: the CI gate has a second, independent failure
+# condition, region-declaration drift, and this hook does not evaluate it. That
+# omission is deliberate. The claim above is that this hook is only ever a
+# cheaper way to find out what CI would have told you, and that holds only
+# while every arm here answers from the same state CI audits:
+#
+#   The missing arm qualifies. It is intersected with this branch's COMMITTED
+#   changed set (the three-dot diff below), and the file map behind `missing`
+#   is a git ls-files key set, which an uncommitted edit cannot move. So it
+#   denies on committed state, the same state CI audits.
+#
+#   Region drift does not qualify. The checker reports it as `.regionDrift` in
+#   the same --json report parsed below, and builds it by reading each
+#   shipped file's CONTENT off disk and diffing the marker-bearing paths it
+#   finds against the committed manifest's declaration, so an uncommitted edit
+#   that adds or removes a marker pair moves it. CI also leaves it unscoped on
+#   purpose, so a stale declaration cannot slip through on a PR that touches no
+#   declared path. Evaluated here, it would be the one arm that denies what CI
+#   would pass, on a working tree the PR never contains, and its remedy text
+#   would tell the maintainer to regenerate a manifest that is not stale.
+#   Narrowing it to the changed set instead would enforce a different rule than
+#   CI's, which that workflow's own inline comment rules out.
+#
+# So region drift stays CI-only and does surface as a red check after the push.
+# That gap in local coverage is the accepted price of the guarantee that a deny
+# here always means a red check there.
+#
 # WHY `gh pr create` AND NOT `git push`: push-time would catch this one round
 # earlier, but it fires on every work-in-progress push to a branch that has no
 # PR and may never get one, where an unanswered manifest entry is not yet a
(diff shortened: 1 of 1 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.qTtmQpQO6z

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.y6Fu3PFWri

PR #1248, opened 2026-08-06
debt(ci): describe the two redundant workflow paths-filter entries honestly (#1238)

### LINKED ISSUE #1238: Two audit-ci-tests.yml paths-filter entries claim to be load-bearing, but the catch-all above them makes both no-ops
## Two paths-filter entries claim to be load-bearing, and the catch-all above them makes both no-ops

`.github/workflows/audit-ci-tests.yml:95` (and the identical case at `:105`)

```yaml
              - '.github/workflows/audit-ci-tests.yml'
              # ...
              - '.github/workflows/code-review-audit.yml'
```

### Failure mode

The `code:` filter output also contains a `.github/workflows/**` catch-all. `dorny/paths-filter` ORs every pattern within one output, so neither entry above changes how any input evaluates: remove either one and the filter answers identically on every possible changed-file set.

Both entries nonetheless carry a multi-line comment asserting the opposite as fact. The one at `:105` states that "without this line the filter would report `code=false`, skip every bats step, and green the job having run zero tests"; the one at `:95` reads the same way. That has been untrue since the catch-all landed.

The cost is not a wrong CI answer, the filter is correct today. It is that a maintainer auditing filter completeness reads a load-bearing justification where none exists, and reasons about the rest of the filter on that basis. This file's comments are unusually load-bearing precisely because the filter is long and hand-maintained: nine other entries carry the same style of justification and those ones *are* real, so a reader has no way to tell the two apart.

### Verification

Parse the filter and compare against the file's own claims:

```bash
python3 - <<'EOF'
import yaml
d = yaml.safe_load(open('.github/workflows/audit-ci-tests.yml'))
for s in d['jobs']['audit-ci-tests']['steps']:
    f = (s.get('with') or {}).get('filters')
    if f:
        for k, v in yaml.safe_load(f).items():
            print(k, v)
EOF
```

`.github/workflows/**` and both named workflow paths appear as siblings under `code:`.

### Suggested fix

Keep the entries and rewrite both comments to the honest form the `sandbox:` filter in the same file already uses for its own redundant lines: the entry is redundant with the catch-all today, listed deliberately so that narrowing the catch-all later cannot silently drop a guarded source, and it is documentation of the test-to-source pairing rather than the thing arming the suite. Deleting the two entries is the other defensible answer and costs that documentation.

PR #1237 applied exactly the first treatment to a third entry of this class it had just added, so the convention is already in the file; this issue is the two pre-existing entries it deliberately left alone as out of scope.

### Surface note

Filed `surface:maintainer`. `audit-ci-tests.yml` is release-excluded maintainer CI, and the defect (a comment that misdescribes its own line) is observable only by a maintainer reading this repository's workflow.

### Provenance

Found by the `code-audit-github-workflows` round-1 audit on PR #1237 (`#1228` + `#1230`), which added a third entry of the same class with the same false comment and corrected it in place. A waive was licensed here (a non-security finding on a file that pull request already changes), but the class had by then been observed on three entries including the new one, so it is filed rather than waived.

Handler: prompt

stevensacks:
## Problem

`audit-ci-tests.yml`'s `code:` paths filter lists `.github/workflows/audit-ci-tests.yml` and `.github/workflows/code-review-audit.yml` as siblings of a `.github/workflows/**` catch-all. `dorny/paths-filter` ORs every pattern within one output, so neither named entry changes how any input evaluates: remove either and the filter answers identically on every possible changed-file set.

The `code-review-audit.yml` entry nonetheless carried a multi-line comment asserting the opposite as fact ("without this line the filter would report `code=false`, skip every bats step, and green the job having run zero tests"), and the `audit-ci-tests.yml` entry carried no justification at all. Nine other entries in the same filter carry that style of justification and those ones *are* real, so a maintainer auditing filter completeness had no way to tell the two classes apart.

## Fix

Both entries now read the way the `distribution-audit-pr.yml` entry and the `sandbox:` filter already do: they state the test-to-source pairing they document, name the catch-all that makes them redundant today, and say they are listed deliberately so that narrowing that glob later cannot silently drop a guarded source. The `code-review-audit.yml` comment keeps its real content (ci-status-member-gate.bats executes `run:` bodies extracted from that workflow) and drops only the false causal clause.

Deleting the two entries was the other defensible answer; it costs that documentation, so this PR takes the treatment the issue recommends and the file already uses in a third place.

## Verification

- Parsed the filter before and after: semantics identical, 52 `code:` patterns and 10 `sandbox:` patterns either way. The change is comment-only.
- `.gaia/tests/lib/lint-yaml.bats` and `.gaia/scripts/tests/retrigger-reachability.bats` pass under bash 5 (28 tests).
- `bash .gaia/tests/shell-lint.sh` passes.
- No suite pins the replaced comment text.

Quality Gate skipped: no staged source or gate-affecting config (single `.yml` workflow file). No CHANGELOG entry: `audit-ci-tests.yml` is release-excluded maintainer CI and this is a comment-only change with no adopter-observable effect.

Closes #1238

### DIFF
--- a/.github/workflows/audit-ci-tests.yml
+++ b/.github/workflows/audit-ci-tests.yml
@@ -92,16 +92,28 @@ jobs:
           filters: |
             code:
               - '.github/audit/**'
+              # This job's own definition. The filter above, the step conditions
+              # below and the bats invocations they arm all live here, so an edit
+              # to this file changes what every suite in this job runs, and
+              # retrigger-reachability.bats reads it as one of the workflows it
+              # scans. Redundant today: the `.github/workflows/**` catch-all below
+              # already matches it. Listed deliberately anyway, the same
+              # convention the `sandbox:` filter uses for its own redundant
+              # entries, so that narrowing that glob later cannot silently drop
+              # this file. It is documentation of the self-coverage, not the thing
+              # arming it.
               - '.github/workflows/audit-ci-tests.yml'
               # ci-status-member-gate.bats EXECUTES `run:` bodies extracted from
               # code-review-audit.yml, so that workflow is a source this suite
               # guards, not just a consumer of it. Its structural lock pins the
               # set of steps that POST a `pending` GAIA-Audit status and fails on
-              # an unguarded new one -- but the edit that adds such a step touches
-              # only the workflow, which matches none of the other paths here, so
-              # without this line the filter would report `code=false`, skip every
-              # bats step, and green the job having run zero tests. Same test-dir/
-              # source-dir pairing as the three instruction files below.
+              # an unguarded new one. Redundant today: the `.github/workflows/**`
+              # catch-all below already matches it. Listed deliberately anyway,
+              # the same convention the `sandbox:` filter uses for its own
+              # redundant entries, so that narrowing that glob later cannot
+              # silently drop this source. It is documentation of the test-dir/
+              # source-dir pairing, the same pairing as the three instruction
+              # files below, not the thing arming it.
               - '.github/workflows/code-review-audit.yml'
               # distribution-audit-pr-gate.bats (.gaia/scripts/tests/) EXECUTES
               # the `run:` body extracted from distribution-audit-pr.yml, so that


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.oCAzJzr9L6

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.3zCwE08aTL

Issue #1245, opened 2026-08-06
Nothing checks the Code Audit Team roster for paths that dispatch no member, so an ownerless path merges unreviewed and is found only by hand


stevensacks:
`.gaia/audit-ci.yml:59`

## Failure mode

The Code Audit Team roster maps file-globs to auditor members, and the merge gate requires a per-member clearance for every member a diff dispatches. **Nothing checks the roster for the inverse**: which paths in the repository dispatch *no* member at all.

A path covered by no member's globs resolves an empty dispatched set, falls to the merge gate's legacy branch, is out-of-scope-allowlisted, and merges unreviewed. That is silent by construction: an empty member set is also the legitimate answer for a genuinely out-of-remit diff, so the gate cannot distinguish "nothing here needs an owner" from "this needs an owner and has none."

**The gap is discovered only by a human reading the roster, and every discovery so far has been incomplete.** The block covering the CLI build/config surface has been hand-extended twice, each time framed as covering that surface, and each time it under-enumerated:

- `34a66bca` (SPEC-034) created the roster.
- `9e236cb2` (#813), *"cover the CLI build/config surface"*, added `package.json`, `pnpm-lock.yaml`, `tsconfig*.json` — and missed four files.
- `dbff69b9` (#1243, closing #1207) added `pnpm-workspace.yaml` — one of those four — and missed the remaining three.

The second extension was authored with the first one's explanatory comment (`:122-129`) directly in front of it, and still under-enumerated. This file has also carried three separate closed tech-debt issues already (#806, #1008, #1207).

**So the recurring defect is not any particular missing glob.** Each hand-fix closes the instance in front of it and leaves the class intact, because the roster has no oracle for its own completeness.

## Suggested fix

Design-first, deliberately not prescribed here. A check would have to settle several things the current code does not:

1. **What the universe of paths to check is.** Every tracked file is the obvious candidate and is almost certainly too broad, since large parts of the tree legitimately need no auditor.
2. **What "should have an owner" means.** Candidate predicates: everything under a set of security-relevant prefixes; everything executable or executed during build/test/lint; everything shipped to adopters via the manifest; or an explicit opt-out list with the check failing closed on anything unlisted.
3. **Where it runs and how it fails.** A bats suite, a step in the audit workflow, or a pre-merge gate arm — and whether an uncovered path blocks a merge or only reports.
4. **How intentional non-coverage is expressed**, so the check does not become a rubber stamp that everyone appends to.

[...]

PR #1249, opened 2026-08-06
debt(audit): give the CLI's own tool configs a Code Audit Team owner (#1244)

### LINKED ISSUE #1244: Three .gaia/cli config files classify to an empty audit member set, so a PR changing only one of them merges with no member responsible
`.gaia/audit-ci.yml:132`

## Failure mode

Three sibling files in `.gaia/cli/` have no Code Audit Team owner and classify to an empty member set:

- `.gaia/cli/vitest.config.ts`
- `.gaia/cli/eslint.config.mjs`
- `.gaia/cli/prettier.config.mjs`

Neither claimant reaches them:

- `code-audit-frontend` (the default member) declares bare `*.config.ts` / `*.config.mjs` / `*.config.cjs` / `*.config.js` / `*.config.mts` at `:81-85`. These are **root-anchored**, so they match `vitest.config.ts` at the repository root and never `.gaia/cli/vitest.config.ts`.
- `code-audit-maintainer-node` declares `.gaia/cli/src/**/*.ts` at `:118`, which excludes anything beside `src/`, plus an explicit CLI build/config list at `:130-133` (`package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `tsconfig*.json`). The three files above are on neither list.

**Outcome:** a pull request whose only change is one of those three resolves an empty dispatched member set, falls to the merge gate's legacy branch, is out-of-scope-allowlisted by the `.gaia/*` arm, and merges with no member responsible for reviewing it. `.gaia/cli/vitest.config.ts` is the sharpest of the three, since `setupFiles` executes arbitrary code in every CLI test run.

Independently reconfirmed from the roster after PR #1243 merged: all three still classify to `[]`.

## Suggested fix

Extend `code-audit-maintainer-node`'s explicit CLI build/config list beside its existing `.gaia/cli/tsconfig*.json` entry, matching the convention that block already sets:

```yaml
- ".gaia/cli/*.config.ts"
- ".gaia/cli/*.config.mjs"
```

Two globs cover all three files. Ownership follows the adjacent entries: the CLI build/config surface belongs to `code-audit-maintainer-node`, not to the default member whose globs are adopter-root-scoped.

This is the mechanical half only. The reason this list keeps needing hand-extension is filed separately; see the companion issue on the roster having no completeness check. Per the drain's Invariant 8 the two are deliberately not batched.

Handler: prompt

stevensacks:
`.gaia/cli/` carries three tool configs beside `src/` that no Code Audit Team member owned: `vitest.config.ts`, `eslint.config.mjs`, and `prettier.config.mjs`.

Neither claimant reached them. The default member declares bare `*.config.ts` / `*.config.mjs`, which never cross a `/`, so they match the repository-root files and reach the CLI's copies not at all. The `code-audit-maintainer-node` member declares `.gaia/cli/src/**/*.ts` plus an explicit build/config list naming `package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, and `tsconfig*.json`, and none of the three.

A pull request whose only change was one of them resolved an empty dispatched member set, fell to the merge gate's legacy branch, was out-of-scope-allowlisted by the `.gaia/*` arm, and merged with no member responsible for reviewing it. `vitest.config.ts` is the sharpest of the three, since its `setupFiles` executes arbitrary code in every CLI test run.

## What changed

- Two directory-anchored globs (`.gaia/cli/*.config.ts`, `.gaia/cli/*.config.mjs`) claim all three for `code-audit-maintainer-node`, in both rosters: the committed `.gaia/audit-ci.yml` and the builtin fallback in `.claude/hooks/lib/audit-scope.sh` the degraded gate falls back to.
- The member's remit region is regenerated from the roster, in roster order, so the roster-verifier's remit-parity invariant holds.
- The member gains a tool-config review lens, so it can actually review what it now owns: a narrowed `include` or widened `exclude` that drops suites, a lowered coverage threshold, a disabled lint rule, and any `setupFiles` entry, which is read as code rather than config.
- Routing is pinned in three places: the classifier-level ownership assertion (both rosters), three dispatch-level resolver tests, and the routing-parity suite's deliberate-move exception set.

## Verification

- Both new guards were proven to fail without the fix: the classifier assertion goes red, and the routing-parity arm reports all three files as `expected=code-audit-maintainer-node after=-`.
- `bash .gaia/scripts/verify-audit-roster.sh` clean.
- `.gaia/tests/hooks/` 1300/1300 and `.gaia/scripts/tests/` 1452 tests, zero failures, under bash 5 via `.gaia/scripts/bats5.sh`.
- `bash .gaia/tests/shell-lint.sh` passed.
- Quality Gate skips per `wiki/decisions/Quality Gate.md`: no staged file matches its source or gate-affecting-config list.

Maintainer-only and release-excluded, so no adopter clone is affected.

Closes #1244

## Audit finding recorded, not fixed here

`code-audit-maintainer-shell` raised one Suggestion (no live failure mode): the two new globs cover `.ts` and `.mjs`, while the default member enumerates five config extensions. A future `.gaia/cli/vitest.config.mts`, `knip.config.js`, or any `*.config.cjs` would match no claimant glob and no default glob, landing in the same ownerless hole one extension over. No such file exists today.

Deliberately not fixed here. Widening the list would be the third hand-extension of the same enumeration (after `pnpm-workspace.yaml` in #1207 and these three configs), and hand-extension is the failure mode rather than the fix. The systemic answer, a completeness check that finds roster paths dispatching no member, is owned by companion issue #1245; this finding is another data point for it.

Recorded here rather than filed, per the PR Merge Workflow's cross-remit eligibility rule: a non-security finding on a path this pull request already changes is recorded in the PR body.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.hQwnt43hOb

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.KqwCM4GYud

Issue #1250, opened 2026-08-06
Accepted audit residuals accumulate in PR bodies under ad-hoc headings, so recorded findings are unfindable by grep, by gh, or by the disposition ledger


stevensacks:
`wiki/concepts/PR Merge Workflow.md:175` (`#### Applying the audit's own Suggestions: digest economics`)

## Failure mode

The digest economics rule is correct and is working as designed: when a PR is clean and its members are marked, folding in a one-line comment fix rotates the digest, invalidates the marker, and buys a full member re-dispatch purely to re-earn it. So sub-threshold audit findings are deferred and recorded in the PR body instead of fixed.

**What has no answer is where they go afterwards.** The workflow specifies one canonical destination — `## Out-of-scope machinery findings (recorded, not filed)`, one entry per finding, each carrying `file:line`, a one-line failure mode, and its dedup key. In practice that form is almost never used, and nothing checks that it is.

Measured across the twelve PRs of the current tech-debt drain:

| PR | Residual heading used |
|---|---|
| #1227 | `Accepted and noted, not fixed`; `Filed, not waived` |
| #1231 | `Accepted residuals (audit round 2…)`; `…(round 3…)` |
| #1233 | `Out-of-scope findings` |
| #1234 | `Known follow-up (out of scope here)`; `Accepted and noted, not fixed` |
| #1236 | `Audit dispositions` |
| #1237 | `Deliberately left alone`; `Round-1/Round-2 audit dispositions` |
| #1240 | `Accepted residuals (audit Suggestions, not addressed)` |
| #1241 | `Audit rounds` |
| #1243 | `Out-of-scope machinery findings (recorded, not filed)` ← the only canonical one |
| #1246 | `Deliberately not done` |
| #1247 | `Accepted residual` |
| #1248 | *(none — no residuals)* |

**Roughly 21 recorded findings sit under eleven different ad-hoc headings, and exactly one carries a dedup key** (#1243, and in bare `v1 class=…` form rather than the `<!-- gaia-debt-key: -->` wrapper).

> **Superseded 2026-08-15.** True of the twelve-PR sample above, and roughly a fifth of the population. A re-measurement across the most recent 400 of 956 merged PRs finds **105 PRs, 133 headings, 86 distinct spellings, ~225 entries, and 40 dedup keys**. See the correction at the foot of this issue.

[...]

PR #1251, opened 2026-08-06
debt(statusline): let a filed over-budget condition clear the audit nudge (#1150)

### LINKED ISSUE #1150: audit nudge re-fires for an over-budget condition the same run already filed, so /gaia-audit cannot clear its own indicator
`.gaia/scripts/check-updates.sh:289`

## Failure mode

The `auditNudge` segment has four signals in priority order: `resume draft` > `N new memories` > `N days since review` > `over budget`. The first three are debounces keyed to whether an audit ran, so completing an audit resets them by construction. The fourth, `project_drift`, is a live measurement of committed file sizes with no debounce and no suppression: it clears only when the files themselves shrink and the change is committed.

`/gaia-audit`'s post-flight (`.claude/skills/gaia/references/audit.md:530-542`) sets `auditNudge = false`, `auditNudgeReason = ""`, and `checkedAt = 0`. The first two clear the segment, and the third forces `check-updates.sh` past its TTL gate so the next statusline render recomputes every signal. If an over-budget file is still over budget, `project_drift` fires again immediately and the segment returns with a new reason. The post-flight comment names this case out loud ("including any over-budget condition a partial apply left unresolved") without treating it as a defect.

From the operator's seat this reads as the command doing nothing: run `/gaia-audit` to clear `Run /gaia-audit (10 new memories)`, watch it clear, then watch it come back as `Run /gaia-audit (over budget)` on the next render. The stated contract is that a single `/gaia-audit` resolves every firing signal, not just the top one.

There is a second, quieter cost. Because the segment renders exactly one reason, a latched `project_drift` cannot change text when a *new* over-budget file appears. The indicator already says `over budget`, so a third file crossing the threshold is invisible until the first two are fixed.

The condition is already tracked as debt at the time it re-fires (the audit files its out-of-scope findings before the post-flight cache bust runs), so the nudge and the debt count are both reporting the same unresolved work, and only one of them is actionable.

## Suggested fix

Suppress `project_drift` for files covered by an open `tech-debt` issue and let the debt indicator carry them, so a completed audit leaves the audit nudge clear and the work sits in the drain queue. Guard the obvious failure mode by recording the measured size in the issue at filing time and re-firing when the file grows past it, so suppression cannot hide unbounded growth.

Preserve one property of the current design while doing it: `project_drift` is repo-wide and shared across developers, while the other three signals are machine-local. A machine-local audit run must not silence a repo-wide condition for everyone. Keying suppression to an open issue holds that line, because the issue is shared state too, whereas keying it to the local report or the local cache would not.

Two design decisions the surrounding code does not settle, which is why this is graded hard: where the measured-size baseline lives (issue body vs. the shared `debt/` cache), and how `check-updates.sh` learns about open issues without adding a `gh` call to a path that currently does only local file IO. The existing `.gaia/scripts/debt-count-refresh.sh` and its `debt/count.json` shared cache are the obvious prior art for the second, but that cache stores a count rather than per-issue paths, so it does not answer the question as-is.

Depends on #1143 and the sibling remedy-collapse issue only for how often this path is reached, not for whether the suppression is correct: those two reduce the number of genuinely-unfixable over-budget conditions, they do not eliminate them.

Handler: plan

stevensacks:
Closes #1150

## Problem

`auditNudge` has four signals in priority order: `resume draft` > `N new memories` > `N days since review` > `over budget`. The first three are debounces keyed to whether an audit ran, so completing an audit resets them by construction. The fourth, `project_drift`, is a live measurement of committed auto-load file sizes with no debounce and no suppression.

`/gaia-audit`'s post-flight sets `auditNudge = false` and `checkedAt = 0`, which clears the segment and forces the next render past the TTL gate so every signal recomputes. An over-budget file that the audit **filed as tech-debt rather than trimmed** is still over budget, so `project_drift` fires again immediately and the segment returns with a new reason. From the operator's seat the command does nothing: run `/gaia-audit` to clear `Run /gaia-audit (10 new memories)`, watch it clear, watch it come back as `Run /gaia-audit (over budget)`.

**Reproduced live on `main` while scoping this**, both halves of the issue's failure mode:

- `CLAUDE.md` at 563 words against a 400-word budget, and `.claude/rules/playwright.md` at 211 lines against a 200-line budget;
- the segment renders one reason, so the second file is invisible behind the first.

Neither path has an open `tech-debt` issue, so this fix is correct and currently unexercised in this repository. It starts working the first time an audit files rather than trims. Fixing the two live over-budget files is not in this PR's scope.

## Fix

A path that already carries an open `tech-debt` issue is the debt indicator's to report, so `project_drift` skips it.

1. **`debt-count-refresh.sh` emits `coveredPaths`** — the repo-relative paths carrying an open `tech-debt` issue, parsed from each issue body's `gaia-debt-key` comment.
2. **`check-updates.sh` reads that cache** and suppresses `project_drift` for a covered path. Local file IO only, so the `auditNudge` block's stated no-network property is preserved.
3. **A growth baseline** records the size at which a path was first suppressed and re-fires when the file grows past it, so suppression cannot hide unbounded growth. The baseline never advances, or a growing file would nudge once and then go quiet.

### The two DECISION OWED items, and how they were settled

Both were put to the maintainer before any code was written.

**Where the baseline lives: the cache, not the issue body.** This overrides the issue's own stated reasoning ("keying it to the local report or the local cache would not [hold the line]"), because that reasoning ignores **direction of effect**. Suppression is keyed to shared state — an open issue — so a machine-local audit run can never silence a repo-wide condition for anyone else, which is the property the issue protects. The baseline only ever *un*-suppresses: a local value can add a nudge and can never remove one. It therefore costs no change to `file-tech-debt` or the filing routes, and it works for the issues already open, none of which records a size. It lives in `update-check.json`, which `check-updates.sh` already owns and writes, so there is no new state file and no second writer racing the refresher.

**How `check-updates.sh` learns about open issues: no new network call.** The issue frames this as avoiding "a `gh` call to a path that currently does only local file IO". That framing is wrong about the script: `check-updates.sh` already calls `gh release list` and shells out to the CLI; it is the 6h background refresher, not the statusline hot path, and only the `auditNudge` *block* is local-IO-only. Meanwhile `debt-count-refresh.sh` was already making precisely the needed call. Its existing single `gh issue list` now fetches bodies too and is filtered twice locally. Without local `jq` the original `gh --jq` count call runs unchanged and `coveredPaths` carries forward, a degradation that is one-directional by construction: fewer covered paths means more nudges, never fewer.

`coveredPaths` and `openCount` deliberately apply **different filters to the same fetch**. `openCount` keeps excluding `debt:in-progress` and `debt:spec-pending` so a claimed issue does not inflate the nudge for a peer session; `coveredPaths` excludes nothing, because a claimed issue still covers its path and suppression must hold while someone is working on it.

## Verification

- **New suite**, `.gaia/tests/statusline/audit-nudge-drift-suppression.bats`, 12 tests, written red-first: 8 failed before the change and the 4 that passed assert behaviour that was already correct.
- **`bash .gaia/tests/shell-lint.sh`** passes.
- **Regression**, 70 tests / 0 failures: `.gaia/tests/statusline/`, `debt-sentinel-touch`, `debt-session-reconcile`, `doc-debt-query`, `registry`.
- **Every new guard mutation-proved, 6 mutants, all killed**: `coveredPaths` wrongly applying the claim filter; the path regex stopping at the first space; suppression never applying; the baseline advancing instead of holding; the rule loop's early exit restored; stale baselines lingering instead of being pruned.

### One mutant survived first time, and the repair is worth reading

Dropping the rule loop's `break` was originally justified with two reasons, and **the first was false**. "A suppressed file must not consume the break" does not hold: the `break` sits inside the `if drifts …` branch and `drifts` returns non-zero for a suppressed file, so a suppressed file never reaches it. Restoring the `break` left all 12 tests green, meaning the assertion guarding it could not fail.

The load-bearing reason is the baseline, not the signal: an early exit leaves a covered file later in the glob unmeasured, so its growth baseline is taken only once the earlier file is fixed, at whatever size it had grown to by then — the growth guard reading a number it never observed. The comment now states only that, and the assertion was replaced with one that measures the later covered file's baseline. The mutant is killed.

Note that commit `a50ab900`'s message still carries the false reason; this PR squash-merges, so the landing message is this body.

### Round 1 — `code-audit-maintainer-shell` withheld its marker

Clean on shellcheck, on bash-3.2 quoting, on the `jq` round-trip, and on the new suite's assertion hygiene; it also confirmed both DECISION OWED calls are deliberate rather than oversights. Four findings were real and all were fixed:

- **Critical, and a pre-flight miss.** `.gaia/scripts/tests/debt-count-refresh.bats` was red at HEAD (7 failures) and green on `main`. The new `jq` arm asks `gh` for raw JSON with no `--jq`, while the suite's stubs emitted a bare count, so local `jq` could not iterate a number and the cache zero-seeded. Production fails *safe* here (the prior cache is preserved), so it was a test-double contract mismatch rather than a production defect, but it turned a CI-gated suite red. Fixed by replacing three bespoke stubs with one `write_gh_stub` that serves `issue list` the way real `gh` does: applies a `--jq` filter when given one, emits the raw `--json` array when not. One stub now covers both call shapes, so a test can no longer pass or fail on the *shape* of the call rather than the behaviour.
- **Important.** `check-updates.sh` claimed "no second writer" for `update-check.json`. False: `/update-deps` and `/update-gaia` rewrite it wholesale from an enumerated preserve list. Unlike `serenaLangDrift`, which recomputes from source, a baseline is a *historical observation*, so after a peer cache-bust it re-seeds at the current, grown size and the growth guard — the only safety valve on this whole feature — is silently defeated.
- **Important, partially fixed.** The count filter was duplicated verbatim at two call sites that could silently diverge; it is now one `COUNT_FILTER` variable read by both arms. The coverage half is recorded below as an accepted residual.
- **Two Suggestions.** The key extraction now anchors on the full `<!-- gaia-debt-key:` opener rather than the bare literal, with a new test pinning that body prose merely mentioning the key injects no path; and a guard that validated one `jq` read while assigning a second now reads once.

A wider run then turned up `debt-origin-contract.bats` test 6 red: it scrapes the count filter out of the script by regex, and the `COUNT_FILTER` hoist moved it out of the `--jq '…'` call site, firing the scraper's own drift guard. Repointed at the `^COUNT_FILTER='…'` declaration, which is the better anchor now that both arms read one variable, and mutation-proved.

### Round 2 — both members cleared, and one finding was folded in anyway

`code-audit-maintainer-shell` and `code-audit-maintainer-prose` (newly in remit from the two `SKILL.md` edits) both wrote earned markers, and both independently re-verified round 1's repairs rather than taking them on report. The shell member also confirmed `.gaia/tests/statusline/` is CI-wired via the `code` filter's `.gaia/scripts/**`, so the new suite is enforced rather than local-only.

**The prose member's Important finding was accepted after the clearance, so the extra round was paid deliberately.** Round 1's preserve-list repair fixed the instance rather than the class: the cache carries 14 fields, both cache-bust steps enumerated a handful, and five were still dropped — two of them reproducing this PR's own defect. `auditMemoryBaseline` re-anchors and disarms the new-memories nudge for another 10 memories, and the two `harden*` counts defeat their own offline-preserve guard. Fixed by the member's own proposed reduction: **stop enumerating what to keep, enumerate only what to change.** That is strictly shorter than the text it replaces, removes the mechanism rather than patching it, and matches the surgical-`jq` precedent that has never had this bug.

Two Suggestions were also accepted and fixed. The no-`jq` comment claimed `coveredPaths` "carries forward unchanged" — false, that write branch hardcodes `[]`, and the file contradicted itself two comments apart. And `write_gh_stub` spliced fixture JSON between single quotes, so one apostrophe in an issue body would emit a stub that is a syntax error, present as a `gh` failure, and red a test on a stale count; the fixture now goes to a sidecar file and an apostrophe fixture pins it.

### Round 3 — both members cleared, one Suggestion accepted rather than fixed

Dispatched in parallel against a freshly captured tree, with both findings sidecars pre-cleared so their presence was a fresh-write signal. The prose member returned **clean, zero findings**, judging round 2's preserve-list reduction a net improvement and verifying that the new lockstep pointer resolves and that the suite genuinely exercises the dedup-key format that contract governs. The shell member returned clean with one Suggestion, having re-derived the pre-change stub in a scratch tree to confirm the apostrophe fixture reds without the fix, and confirmed no coverage regression.

That Suggestion is **accepted as a residual and recorded below rather than fixed**, under a stopping rule written before the round ran: a finding on prose an earlier round of this step wrote is the degradation signal, not a defect. The PR is clean and both members are marked, so folding in a one-word deletion buys a full member re-dispatch solely to re-earn a marker.

**Four false comments were authored across this step**, each caught by a different instrument: the `break` rationale by the mutation pass, "no second writer" by round 1, "carries forward unchanged" by round 2, and "server-side" by round 3. Every one was a plausible claim about behaviour that had not been executed. This step's defects were overwhelmingly in prose *about* the code rather than in the code, and that is the most transferable thing it found.

## Verification, final

Re-run at the tree that was dispatched and merged, across the three CI roots rather than the one directory the diff sits in — **3404 tests, 0 failures**: `.github/audit/tests/` 165, `.gaia/scripts/tests/` 1455, `.gaia/tests/hooks/` 1300, `.gaia/tests/lib/` 447 + 1 skip, `.gaia/tests/statusline/` 36. `bash .gaia/tests/shell-lint.sh` clean across 196 `.sh` scripts, 192 `.bats` suites, the husky hook, and the hook-array guard. The Quality Gate correctly skips: every changed file is `.sh`, `.bats` or `.md`, so it has nothing to check.

## Accepted residuals

Recorded rather than fixed, one entry per finding, each carrying `file:line`, a one-line failure mode, and a dedup key so `git grep` finds it in one pass.

**`.gaia/scripts/debt-count-refresh.sh:198` — the `gh --jq` no-`jq` fallback arm has no test coverage.** Failure mode: a regression in that arm's filter expression or its numeric guard ships unobserved, because no suite constructs a machine without local `jq`. Bounded by the arm's own failure direction — it can only fail to recompute, leaving the prior cache intact and the sentinel set for the next tick.

Not fixed here because covering it means running the script with `jq` hidden from `PATH`, and the script resolves `date`, `stat`, `mktemp`, `mv`, `rm` and `mkdir` through `PATH` too. A `run_refresh_nojq` helper therefore needs a six-symlink `minbin` built and torn down per test, to exercise one four-line branch. **The reason first given for this residual was wrong and is withdrawn:** it claimed a stripped `PATH` would break the suite's own `jq` assertions, but `run_refresh`'s `PATH=` is a subshell command prefix, so the assertions already run under the unmodified `PATH`. An accepted residual is only as good as its stated reason; a wrong one stops the next reader from looking.

**On the heading.** The workflow's one canonical residual destination, `## Out-of-scope machinery findings (recorded, not filed)`, governs *waived cross-remit* findings, and the gates read a matching `machinery_waived` disposition-ledger entry for each. This residual is neither: it is in-domain, on a file this PR changes, and nothing was waived. Filing it under that heading would put a false entry in the census that #1250 exists to build. The dedup key is the findable half of #1250's proposed convention and is used here regardless.

**`.gaia/scripts/debt-count-refresh.sh:166` — the comment says `gh`'s own `--jq` computes the count "server-side". It does not.** `gh` applies `--jq` client-side with its embedded gojq, against the already-fetched response, and both arms issue the identical `gh issue list --limit 1000`. Failure mode: a reader reasons about the degradation budget from a false premise, believing the no-`jq` path transfers less data than the `jq` path when the two are identical on the wire.

Bounded, and the bound is why this is a residual rather than a fix. The comment's load-bearing claim — that the degradation is one-directional, fewer covered paths meaning more nudges and never fewer — is stated correctly and separately in the same block, and does not depend on where the filter runs. The repair is a two-word deletion, but it is a content edit on a file the shell member owns, so it would rotate that member's digest and buy a full re-dispatch purely to re-earn a marker on an already-clean PR.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.upwtF9xtPV

PR #1253, opened 2026-08-07
debt(ci): name the statusline suite's full guarded surface in its paths filter (#1252)

### LINKED ISSUE #1252: The statusline suite's paths-filter comment names a narrower guarded surface than the suite actually has
`.github/workflows/audit-ci-tests.yml:267`

## Failure mode

The paths-filter comment at line 267 names the statusline suite's guarded surface:

> The statusline suite (`.gaia/tests/statusline/`) exercises the shipped `.gaia/statusline/gaia-statusline.sh` worktree detection; trigger on both the test dir and the script it guards.

That is now false by omission. As of PR #1251 the statusline suite also guards `.gaia/scripts/check-updates.sh` and `.gaia/scripts/debt-count-refresh.sh`: the new `.gaia/tests/statusline/audit-nudge-drift-suppression.bats` exercises both, asserting the covered-path suppression and its growth baseline end to end.

**There is no live failure, and that is what makes this worth filing rather than shrugging at.** The trigger is already correct: the `code` filter's `.gaia/scripts/**` glob arms the suite for both scripts, so the suite runs when it should. The defect is that the comment states a narrower guarded surface than the suite actually has.

Trigger and bad outcome: an editor prunes or narrows the `code` filter's `.gaia/scripts/**` glob, reads line 267 to check what the statusline suite depends on, and concludes it guards only `gaia-statusline.sh`. They drop the coverage for `check-updates.sh` and `debt-count-refresh.sh` without noticing, and the audit-nudge suppression suite silently stops running on changes to the two scripts it exists to guard. The comment is the only place a reader is told what that suite covers, so a false statement there is load-bearing exactly when someone is deciding what to remove.

## Suggested fix

Widen the comment at line 267 to name the two additional guarded scripts, so the stated surface matches the suite's real one. Comment-only; no filter change, because the filter is already correct.

## Provenance note

Surfaced as a cross-remit finding by `code-audit-maintainer-shell` during PR #1251's audit gate. Its owner is `code-audit-github-workflows`, which that diff did not dispatch. Filed rather than waived into the PR body: the path is not in #1251's diff, so the machinery waive is not licensed and the routing rule says file. It is also the same honest-comment class as #1238 / PR #1248, i.e. a class already observed to spread, which independently forces a filing.

Handler: prompt

stevensacks:
## What

The `code` paths-filter comment in `.github/workflows/audit-ci-tests.yml` described the statusline suite as guarding only `.gaia/statusline/gaia-statusline.sh` worktree detection. The suite also runs real copies of `.gaia/scripts/check-updates.sh` (the `hardenUnclassifiedCount` segment and the audit nudge's `project_drift` suppression) and `.gaia/scripts/debt-count-refresh.sh` (the `coveredPaths` that suppression keys on).

The comment now names all three guarded sources and states why only the two statusline dirs are listed at that spot: the `.gaia/scripts/**` glob above already arms the suite for both refreshers.

## Why it matters

There is no live failure. The trigger is already correct. The comment is the only place a reader is told what that suite covers, so it is load-bearing exactly when someone is deciding whether the `.gaia/scripts/**` glob can be narrowed. A false statement there lets that pruning drop coverage for the two scripts the audit-nudge suppression suite exists to guard, silently.

## Scope

Comment only. No filter change, no behavior change. `.github/workflows/audit-ci-tests.yml` is maintainer-only and release-excluded, so nothing mirrors to an adopter template.

## Verification

- `.gaia/tests/lib/lint-yaml.bats` + `.gaia/scripts/tests/retrigger-reachability.bats` under bash 5: 28/28 pass
- Quality Gate skipped per its own skip logic (no staged source or gate-affecting config; a YAML comment is not inspectable by typecheck/lint/test/build)

Closes #1252

### DIFF
--- a/.github/workflows/audit-ci-tests.yml
+++ b/.github/workflows/audit-ci-tests.yml
@@ -265,8 +265,16 @@ jobs:
               - '.claude/skills/gaia/references/forensics*'
               - '.github/forensics/**'
               # The statusline suite (.gaia/tests/statusline/) exercises the
-              # shipped .gaia/statusline/gaia-statusline.sh worktree detection;
-              # trigger on both the test dir and the script it guards.
+              # shipped .gaia/statusline/gaia-statusline.sh worktree gate and
+              # segment rendering, and it runs real copies of the two
+              # refreshers that feed those segments:
+              # .gaia/scripts/check-updates.sh (the hardenUnclassifiedCount
+              # segment, and the audit nudge's project_drift suppression) and
+              # .gaia/scripts/debt-count-refresh.sh (the coveredPaths that
+              # suppression keys on). Both refreshers are already covered by
+              # the .gaia/scripts/** glob above, so only the two statusline
+              # dirs need listing here -- but narrowing that glob would stop
+              # arming this suite on two of the three sources it guards.
               - '.gaia/statusline/**'
               - '.gaia/tests/statusline/**'
               # retrigger-reachability.bats (.gaia/scripts/tests/) reads every


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.7uFUthQ7CX

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.TwYyLoJa5V

PR #1254, opened 2026-08-07
feat(wiki): catch the local base branch up after a wiki landing (SPEC-066)

stevensacks:
## What this changes

A GAIA wiki landing opens a pull request, queues auto-merge, and returns before the merge lands. The local base branch never catches up, so the wiki state file keeps naming a pre-sync commit and the next sync re-evaluates a range that already landed. This fixes it in two layers.

**Layer 1, the guarantee.** The session-start janitor's merged-wiki-branch sweep gains an existence gate, a bounded prune-fetch, a guarded reap, and a durable checkout-aware `--ff-only` fast-forward of the base branch. The catch-up obligation survives a session that cannot discharge it, carried by a bounded breadcrumb under `.gaia/local/cache/shared/`.

**Layer 2, the convenience.** A new self-discovering `gaia wiki sync await` CLI verb that the `/gaia-wiki` router invokes unconditionally after the landing stage on both call paths. Nothing in layer 2 is load-bearing.

Plus the CLI merge-budget collapse onto one exported symbol, `cleanupAfterMerge`'s export, the verb's registration surfaces, bats and vitest coverage for every acceptance arm, and every shipped prose surface this change falsifies.

## Verification

Per-phase gate: `pnpm typecheck`, `pnpm lint`, the whole `.gaia/tests/hooks/` bats directory via `.gaia/scripts/bats5.sh`, `bash .gaia/tests/shell-lint.sh`, and the `.gaia/cli` vitest run. A final phase mutates each new guard and confirms it fails, so no guard ships hollow.

## Audit findings recorded rather than fixed

Four pre-merge audit rounds ran. The suggestions below sit on paths this pull request already
changes, so they are recorded here rather than filed, and were declined because their regression
risk outweighs their value on a branch this heavily repaired.

- `finalizeMerge` calls `cleanupAfterMerge` as a bare statement and discards the boolean this
  change added, so it can print a terminal cleaned-up message on a refused `git branch -D`. The
  Node member could construct no reachable trigger: every caller's base either always exists or is
  guarded upstream.
- `rollbackLocalLanding` in `sync-land.ts` keeps bare `git checkout` / `git branch -D` calls while
  the same two shapes in `land.ts` gained `--end-of-options` / `--`. Both inputs are repository
  derived, so this is a cosmetic asymmetry rather than a live injection.
- Two prose reductions in the router's standalone-path paragraphs (three paragraphs re-establishing
  one condition, and the ceiling paragraph's budget arithmetic).
- Two shellcheck SC2314 sites in `local-janitor.bats` (bare non-final `!` negation). Pre-existing
  lines this change does not touch; they fail correctly today because each is its test's final
  command.

One cross-remit finding on `wiki/decisions/Wiki Management.md` was fixed rather than recorded,
because it contradicted the shipped skill outright.

Both sit on files this pull request changes and neither is security-class, so the waive rule in
`wiki/concepts/PR Merge Workflow.md` records them here rather than filing them.

- `.claude/hooks/local-janitor.sh:1329` — sweep #8's `git cherry` reap guard tests per-commit patch
  ids, so a squash-merged branch carrying the audit trailer's empty commit reads as having unpushed
  work and is never reaped. Every pull request that clears the audit gate carries that commit, so
  the orphaned-worktree reap is unreachable in practice and worktrees under `.claude/worktrees/`
  accumulate. Fails safe: nothing is deleted. The repair is a combined-diff patch id compared
  against `merge-base..origin/<base>`, plus a fixture whose branch is squash-merged with the trailer
  commit on top; the existing reap-success fixtures build branches zero commits ahead of
  `origin/main`, a shape no real candidate has, which is why the suite stays green.
  Key: `v1 class=reap-guard-over-broad path=.claude/hooks/local-janitor.sh line=1329`
- `.gaia/cli/src/wiki/sync-await.ts:160` — `discoverBranch` takes the most recent `wiki-sync/*` ref
  with no liveness filter. A landing whose pull request never merged leaves a branch that is not
  `[gone]`, so the janitor's reap skips it forever; once it becomes the only candidate, every
  `/gaia-wiki` run spends the full poll slice on it and relays a pending line for a landing that
  succeeded. Layer 2 is explicitly non-load-bearing, so the janitor still catches base up. The
  repair is a committerdate liveness window ahead of the recency pick.
  Key: `v1 class=stale-branch-discovery path=.gaia/cli/src/wiki/sync-await.ts line=160`


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.AZeF14F0G5

PR #1256, opened 2026-08-07
fix(janitor): prove the orphaned-worktree reap by combined-diff patch id (SPEC-067)

stevensacks:
## What

`.claude/hooks/local-janitor.sh` sweep #8 reaps abandoned GAIA worktrees under `.claude/worktrees/`. It proves a merge with `git cherry`, which compares per-commit patch ids and therefore matches nothing a squash merge produces, so it refuses every real candidate and worktrees accumulate.

This replaces that test with the branch's combined diff compared by whitespace-verbatim patch id against `<merge-base>..origin/<base>`, moves the cheap refusal gates ahead of the scan, repairs the clean-tree gate so a failed read keeps the worktree, and backs all of it with fixtures built in the shape a real candidate has plus a checked-in, re-runnable mutation-proof record.

## Why it matters

This is the janitor's one destructive path and it has already produced two data-loss findings. Every unanswerable question keeps the worktree: unresolvable base or merge base, a failed diff / rev-list / patch-id / log read, an empty patch id on a branch one or more commits ahead, and a bounded scan that ends without a match.

## Phases

1. Fixture infrastructure in the reap suite
2. The guard rewrite plus its behavioral tests
3. Argv-witness assertions for the scan bound, gate order, and silence
4. A structural guard over the reap suite's pre-existing test count
5. The checked-in mutation-proof record

## Verification

Per phase: `bash .gaia/scripts/bats5.sh .gaia/tests/hooks/ .gaia/scripts/tests/ .gaia/tests/lib/` and `bash .gaia/tests/shell-lint.sh`.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.LQfQVNbJtv

PR #1259, opened 2026-08-07
debt(janitor): memoize the worktree reap's declined merge-evidence scan (#1257 #1258)

### LINKED ISSUE #1257: janitor: the worktree reap re-walks the same upstream range for every unmatched candidate, every session
`.claude/hooks/local-janitor.sh:1451`

## Failure mode

Sweep #8's merge-evidence guard proves a squash merge by comparing the candidate branch's combined-diff patch id against the patch ids of `<merge-base>..origin/<base>`, walked by `git log -p ... -n 1000 | git patch-id --verbatim`. That walk runs once per candidate, inside the per-worktree loop, with nothing memoizing its result across iterations.

A `[gone]` worktree whose branch was abandoned rather than merged never produces a match, so it is kept, correctly, and then pays the identical full walk again at the next session start. Nothing about the repository has changed, and nothing ever will change to make that candidate match, so the cost recurs indefinitely for as long as the worktree sits there. Two candidates sharing the same base each pay their own separate walk in the same run: the gate-order test's positive control already records two distinct `log ... -n 1000` invocations for exactly that reason.

Measured by the pre-merge shell audit member: ~0.39s for a 193-commit range, so roughly 2s at the `-n 1000` bound, per unmatched candidate, per session.

This is a cost issue only, never a correctness one. Every path involved fails toward keeping the worktree, and the `-n 1000` bound already caps the worst case, so the ceiling is bounded and the outcome is always right. What is missing is that the bounded work repeats when its answer provably cannot have changed.

## Suggested fix

Memoize the negative answer keyed on the pair `(branch tip sha, origin/<base> tip sha)`. Both tips are already in hand at that point in the loop, and either one moving is exactly the condition that could turn a non-match into a match, so the key is precise rather than heuristic. Caching only the negative result keeps the change safe by construction: a stale entry can only ever cause a keep, which is the direction this guard already fails in.

Deliberately out of scope for the change that surfaced it. This reap is the janitor's one destructive path and has already produced two data-loss findings, so introducing new persistent state on it at merge time is not a change to bolt on. Whoever picks this up should decide where the cache lives against the existing `.gaia/local/` state conventions and the state registry's scope rules, rather than inventing a new location.

Confirmed for filing by the `code-audit-maintainer-shell` member across two review rounds, after it declined to treat the item as blocking.

Handler: plan

### LINKED ISSUE #1258: janitor: the file header's sweep #8 entry omits the merge-evidence guard from its own list of preconditions
`.claude/hooks/local-janitor.sh:164` (the sweep #8 entry in the file header), and its guard enumeration at `:186-189`.

## Failure mode

The file header is the summary a maintainer reads to understand what each sweep does before touching it. Its sweep #8 entry describes the reap's preconditions as: branch upstream is `[gone]`, working tree is clean, and no live RUNNING plan sentinel names the branch. Its closing enumeration repeats the same set:

```
#      run far more often than it would on its own. Its own guards are what
#      bound that: only worktrees under the main checkout's .claude/worktrees/,
#      never the current checkout, never a detached HEAD, a clean working tree
#      required, and never a branch a live RUNNING plan sentinel names.
```

Neither passage mentions the merge-evidence guard, which is the strongest and most important condition on this path: the reap also requires the branch's combined diff to match, by whitespace-verbatim patch id, a commit in `<merge-base>..origin/<base>`. A reader who takes the header at its word concludes that `[gone]` plus a clean tree plus no sentinel is sufficient to delete a worktree and hard-delete its branch.

That is precisely the inference the guard exists to prevent, and the header states it twice. `[gone]` does not prove a merge: a pull request closed without merging and then branch-deleted reads identically, and sweep #1's prune-fetch establishes `[gone]` routinely on every qualifying session, so the state arrives as a matter of course rather than incidentally.

The concrete bad outcome is a future edit made from the header's model rather than the code's. Someone simplifying, reordering, or extracting this sweep, or writing a new caller against the documented contract, has a written basis for treating the merge-evidence reads as incidental and removing or bypassing them. This reap is the janitor's one destructive path and has already produced two data-loss findings; the header is what a reader consults before deciding a guard is redundant.

The omission predates the change that surfaced it. It was found while replacing the sweep's merge-evidence test (SPEC-067) and deliberately left alone there, because that work froze its edit scope to the sweep's body and widening it at merge time was not a call to make unilaterally.

## Suggested fix

Add the merge-evidence condition to both passages, at the altitude the rest of the header uses: state that the reap additionally requires the branch's combined diff to match an upstream commit's patch id under a bounded scan, and that every read it cannot answer keeps the worktree. Keep it to a sentence or two in each place; the guard's own comment block at the reap already carries the full reasoning, and duplicating that here would create a second copy to drift.

While in the file, check the sibling sweep entries for the same class of omission rather than fixing only the one cited: the value here is that the header's enumerations match the code's actual refusals, not that this single entry is patched.

Handler: prompt

stevensacks:
Drains two tech-debt issues that both cite `.claude/hooks/local-janitor.sh`, batched onto one branch.

## #1257 — the reap re-walks the same upstream range every session

Sweep #8 proves a squash merge by comparing a candidate branch's combined-diff patch id against a bounded scan of `<merge-base>..origin/<base>`. That scan ran once per candidate inside the per-worktree loop with nothing memoizing it. A `[gone]` worktree whose branch was abandoned rather than merged never matches, so it is kept, correctly, and then pays the identical full scan again at the next session start. Nothing about the repository can change that answer, so the cost recurred indefinitely, roughly 2s at the `-n 1000` bound per unmatched candidate per session, each such worktree paying its own.

The declined answer is now memoized against the pair of tips whose movement is exactly what could turn a non-match into a match: the branch tip and the `origin/<base>` tip.

Safety properties, each with a test:

- **Only the negative answer is recorded.** A memo hit skips the scan and *keeps* the worktree, the same direction every unanswerable question in this sweep already fails toward. No memo can cause a reap.
- **Only a completed scan is recorded.** A failed upstream read proves nothing; recording it would suppress the real scan for as long as both tips held and the reap would go silently inert.
- **Both tips are in the key.** Landing the branch's work upstream moves `origin/<base>`, which brings the scan back on its own, so a memo cannot hold a worktree past the point its merge becomes provable.
- **Rewritten, not appended.** The file is rebuilt from each run's own misses, so an entry never outlives the worktree it describes and the file cannot grow without bound.

The memo file is a new registry entry (`worktree-reap-miss-memo`, scope `main-only`), anchored at main's cache for the same reason sweep #5b is. Without a registry entry, sweep #9 would report it as an unrecognized `cache/` child on stderr at every session start, forever.

## #1258 — the header's sweep #8 entry omits the merge-evidence guard

The file header described the reap's preconditions twice (the sweep #8 entry and its closing enumeration) and named the merge-evidence guard in neither, leaving `[gone]` + clean tree + no sentinel reading as sufficient to delete a worktree and hard-delete its branch. That is the inference the guard exists to prevent, on the janitor's one destructive path. Both passages now state it, at the altitude sweep #1 (c) already uses for the same distinction.

The issue also asked for a sweep across the sibling entries rather than a patch of the one cited. That found one more omission: sweep #4's entry did not say that its drop-zone exception list is read from the state registry, nor that a list it cannot read skips the sweep outright. Sweeps 1, 2, 3, 5, 6, 7 and 9 check out against their code.

## Verification

- `.gaia/tests/hooks/local-janitor-worktree-reap.bats`: 46 tests, 8 new, all green.
- Full `.gaia/tests/hooks/` suite: 1368 tests, zero failures.
- `bash .gaia/tests/shell-lint.sh`: passed.
- `check-registry-completeness.sh`, `check-registry-source-literals.sh` (no phantom entries), `check-registry-runtime.sh`, `state-registry-lib.bats` (37), and the CLI `link-worktree` test (19): all green.
- **The mutation-proof record was re-run in full against the modified source**, not just re-stamped. Mutations 1, 3, 4, 5, 7 and 9 each still red exactly their named test. Mutations 8 and 10 red additional memo tests, attributably: 8 rewrites the `-n 1000` argv token the memo tests count scans by, and 10 forces every scan to match, which reaps the unmerged fixtures the memo tests are built on. Two new rows (11, 12) prove the memo's own guard arms can fail. Mutation 11 retires a row that previously sat under "arms with no test that reds them": keep-versus-reap could not distinguish a failed scan from a scan that completed without a match, and the memo is what makes the two observable.

The Quality Gate does not apply: no `.ts/.tsx/.js/.jsx/.mjs/.cjs/.css` or gate-affecting config is staged.

## Accepted residuals

`code-audit-maintainer-shell` cleared this PR on its first round with no Critical, Important, or in-remit Suggestion findings. It raised one cross-remit Suggestion, accepted rather than fixed:

- **`.gaia/tests/hooks/local-janitor-worktree-reap-mutations.md`** — the memo adds one arm no fixture reds: the `else` branch that removes the memo file when a run records no misses at all. Replacing that `rm -f` with `true` leaves all 46 tests green, and the record's own "Arms with no test that reds them" section does not list it.
- **Why accepted.** The arm's only failure mode is a memo file outliving the last candidate that needed it. Entries are keyed on the exact `(branch, branch tip, origin/<base> tip)` triple, so a stale line can only ever re-hit when the answer it holds is still correct, and keep-versus-reap cannot observe the difference. There is no live failure mode behind it.
- **Why not folded in.** The assertion that would prove it belongs in the bats suite, which is inside the reviewing member's own remit, so adding it rotates that member's content digest and buys a full re-dispatch solely to re-earn a marker already held on a clean round. Per [[PR Merge Workflow]]'s digest economics, that is the accept-and-note case rather than the apply case.
- **Not filed as debt.** The path is one this PR already changes, and the finding is non-security, so the cross-remit eligibility rule records it here rather than opening an issue.

Closes #1257
Closes #1258


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.oY1gW5zhW9

PR #1260, opened 2026-08-08
test(janitor): prove sweep #8's memo-file removal arm

stevensacks:
## What

Sweep #8 of `.claude/hooks/local-janitor.sh` rewrites the negative merge-evidence memo on every run, and removes the memo file outright when a run records no misses at all. That `else` arm shipped in #1259 with no test that reds it: replacing its `rm -f` with `true` left all 46 tests green.

The arm is reachable, and cheaply. `reaps a memoized candidate once its work genuinely lands upstream` already drives its second janitor run to zero recorded misses, which is exactly the condition that takes the arm. It just never asserted the file was gone afterwards. This adds that one assertion.

## What changed

- `.gaia/tests/hooks/local-janitor-worktree-reap.bats`: one assertion added to an existing test. Test count stays 46.
- `.gaia/tests/hooks/local-janitor-worktree-reap-mutations.md`: a new `## Mutation 13` recording the observed red.

**No source change.** `.claude/hooks/local-janitor.sh` is untouched: the mutation was applied, observed, and reverted, so every checksum in the mutation record stays `190bc11f…`.

## Verification

- `bash .gaia/scripts/bats5.sh .gaia/tests/hooks/local-janitor-worktree-reap.bats` — 46/46 green, re-run after the last edit including the prose.
- Mutation applied (`rm -f "$wt_memo_file" 2>/dev/null || true` → `true`): exactly `reaps a memoized candidate once its work genuinely lands upstream` red, the other 45 green. Reverted; `git diff` on the janitor is empty and `shasum -a 256` matches.
- `bash .gaia/tests/shell-lint.sh` — passes.

## Notes

This resolves the cross-remit Suggestion accepted as a residual on #1259, which described the arm as unprovable rather than merely untested. It is provable, so it is proven rather than left as a permanent caveat on the janitor's one destructive path.

No CHANGELOG entry: test-only change to maintainer-only, release-excluded infrastructure with no adopter-visible behavior change.

### DIFF
--- a/.gaia/tests/hooks/local-janitor-worktree-reap-mutations.md
+++ b/.gaia/tests/hooks/local-janitor-worktree-reap-mutations.md
@@ -331,6 +331,36 @@ provable.
 Checksum before: `190bc11fbccfb6d28b76b05ab4e3dcc5e107b18f7abd689389efb3ea5228b5d3`.
 Checksum after revert: `190bc11fbccfb6d28b76b05ab4e3dcc5e107b18f7abd689389efb3ea5228b5d3`. Match.
 
+## Mutation 13: a run with no misses removes the memo file
+
+The memo is rewritten from scratch on every run rather than appended to, so a run that records no
+miss at all must leave no memo behind. Keeping the previous file in that case would hold entries for
+candidates the run no longer saw — including a worktree it just reaped — and those stale entries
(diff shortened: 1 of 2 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.ze2hrH1Leb

PR #1262, opened 2026-08-08
debt(chromatic): drop env for the deleted define pipeline and guard the built preview (#1211)

### LINKED ISSUE #1211: The Chromatic Storybook build injects env for a deleted define pipeline, and nothing asserts the published preview carries none
The Chromatic Storybook build still injects environment values for a substitution pipeline that no longer exists, and nothing asserts that the artifact it publishes carries none.

**Location:** `.github/workflows/chromatic.yml:127` (`API_URL`), with `COMMIT_SHA` at `:128`, `MSW_ENABLED` at `:129` and `npm_package_version` at `:130`, in the `env:` block at `:125` on the `Run Chromatic` step.

## Failure mode

Two halves, both on the same block.

**1. Four of the injected values are now dead configuration.** They reached the preview through `.storybook/main.ts`'s `loadEnv` call and its generated `define` map, which PR #1210 deletes along with `.storybook/env.ts`. With no `define`, nothing in the build reads them. They are not merely redundant: they read as a statement that the published preview carries deployment values, which is the opposite of what the build now does, so the next reader of this block infers a mechanism that is gone. This is the same defect class `#1020` named as its "second cost", one layer up from where `#1020` fixed it.

Note that `SESSION_SECRET` (`:134`) and `SITE_URL` (`:135`) are **not** part of this: `clientSchema` in `app/env.server.ts` omits both, so neither was ever in the `define` map. Their own comment says they exist to satisfy `app/env.server.ts`'s import-time `schema.parse(process.env)`. That premise looks stale too and is worth checking while here: a `pnpm build-storybook` run with neither variable set exits 0, which it could not do if that module were imported during the build.

**2. Nothing asserts the published preview carries no environment values.** PR #1210 adds `test/preview-env.test.ts`, which pins the Vitest **worker** path. The Storybook **build** path is a separate Vite pipeline Vitest never touches, so a `define` re-added to `.storybook/main.ts` or `.storybook/vite.config.ts` would bake literal values into `storybook-static/**/*.js` and no test and no CI step would fail. The artifact is uploaded to Chromatic and published, so the failure is silent and the blast radius is a public artifact.

## Suggested fix

Take the two halves in order; the first shrinks the second.

1. Remove the four dead entries. Verify rather than assume by checking that a build with them unset still produces a working preview, and settle the `SESSION_SECRET` / `SITE_URL` question separately rather than folding it in.
2. Add an assertion on the built artifact. The design is genuinely open and the surrounding code does not settle it, which is why this is graded `hard`:
   - grepping the output for the injected values risks false positives, since a version string or a `localhost` URL can appear in bundled dependency code for unrelated reasons;
   - grepping for the `SESSION_SECRET` placeholder literal is false-positive-free but would not catch the client-key class at all, since that key was never in the `define` map;
   - a **sentinel-value build** is the form with real teeth and no false positives: build with unique sentinel values in the build process's environment and assert none appear in the output. `loadEnv(mode, cwd, '')` includes `process.env` because an empty prefix matches every key, so a sentinel injected as an environment variable does reach any re-added `define`. The cost is that the Chromatic step builds internally, so the assertion needs either a second build or a restructuring of that step.
   - a source-conformance grep for the string `define` in the two Storybook configs is the cheapest and the weakest: it is defeated by moving the substitution into a plugin or a third config file.

Whichever is chosen, it belongs on the build path rather than in the unit suite, since that is the pipeline the artifact comes from.

Handler: plan

stevensacks:
## Problem

Two halves on one block, both in `.github/workflows/chromatic.yml`.

**1. Four injected values are dead configuration.** `API_URL`, `COMMIT_SHA`, `MSW_ENABLED` and `npm_package_version` reached the preview only through the `define` map deleted in `395b416e` along with `.storybook/env.ts`. Verified rather than assumed: that commit's parent defined exactly `import.meta.env.{API_URL, COMMIT_SHA, MSW_ENABLED, NODE_ENV, npm_package_version}`, so these four are precisely the dead set. Nothing in the build reads them now, which makes the block a statement that the published preview carries deployment values, the opposite of what the build does.

**2. Nothing asserted the artifact carries none.** `test/preview-env.test.ts` pins the Vitest **worker**. The Storybook **build** is a separate Vite pipeline that suite never touches, so a `define` re-added to `.storybook/main.ts`, `.storybook/vite.config.ts`, a third config or a plugin would bake literal values into `storybook-static/**/*.js` with no test and no CI step failing. The artifact is uploaded and published, so the blast radius is public.

## Fix

Remove the four entries, and with them the `get-npm-version` step, whose output had exactly one consumer.

Add an assertion on the build path, where the artifact comes from: build with unique sentinel values in the environment and fail if one survives into `storybook-static/`. It runs as a step in the existing `chromatic` job, before `Run Chromatic`, so a failure blocks the upload rather than reporting after it.

Two details are load-bearing and both are commented in place:

- **`pnpm exec storybook build --test`, not the `build-storybook` script.** `pnpm run` sets `npm_package_version` from `package.json`, which overwrites that sentinel and would silently drop one of the four keys from the guard's reach while still reporting green. Measured, not inferred: under the script only three of four sentinels reach the artifact, and the real version string appears instead.
- **The sentinel set is `clientSchema`'s browser-reachable keys**, which is what a substitution map would name, minus `NODE_ENV`, whose real value the build needs. A map reads them with `loadEnv(mode, cwd, '')`, and an empty prefix matches every key, so a sentinel set in the step's environment reaches one wherever it lives.

Placement is a step rather than a new job or workflow: it reuses the job's checkout, install and paths filter (`.storybook/**` and `vite.config.*` are already in the `code` allowlist, so every location a substitution could re-enter already arms it), and it adds no required check to register. Gating behind `has-chromatic-token` is correct rather than incidental, since the failure mode is a *published* artifact.

The publish path itself is untouched. Folding the assertion into `chromaui/action` via `storybookBuildDir` is strictly stronger and cheaper, and was declined on purpose: TurboSnap under `onlyChanged: true` cannot be verified locally, which is the constraint `#1211` names.

## Verification

**Quality Gate: skipped by its own rule.** The whole diff is one workflow file plus `CHANGELOG.md`; the quick check in `wiki/decisions/Quality Gate.md` (**When to skip the gate**) returns nothing.

**CI-gated suites, all six roots, run per `.claude/rules/bats-assertions.md`:** `.github/audit/tests/` 165/165, `.gaia/scripts/tests/` 1458/1458, `.gaia/tests/hooks/` 1368/1368, `.gaia/tests/lib/` 448/448, `.gaia/tests/statusline/` 36/36, `shell-lint.sh` clean. Zero failures.

**The guard was proved red before being relied on, twice**, running the step's `run:` block verbatim from a scratch file with the step's `env:` applied:

| Case | Tree | Result |
|---|---|---|
| Green | as merged | exit 0, no sentinel in the artifact |
| Red 1 | the `viteFinal` + `loadEnv` + `define` map from `395b416e^` restored to `.storybook/main.ts`, plus a reader in `.storybook/preview.ts` | exit 1, error names `storybook-static/assets/iframe-*.js`, **all four** sentinels present |
| Red 2 | `define` in `.storybook/vite.config.ts` instead, `main.ts` untouched | exit 1, all four sentinels |

Red 2 exists because the step's comment claims a re-added map is caught *wherever it lives*. The claim is executed rather than asserted.

**One thing the first fixture got wrong, recorded because it shapes what the guard means.** Restoring the deleted pipeline exactly, `env.ts` re-created and imported from `preview.ts` as `import './env'`, produces **no** sentinel in the artifact. `package.json` sets `"sideEffects": false`, so Rollup drops a bare side-effect import from a production build; a probe confirmed `preview.ts` reaches the bundle and `env.ts` does not. This is consistent with the record rather than new (the `#1210` CHANGELOG entry states the shim never survived bundling), and it is why the faithful fixture puts the read in a retained module. It also means the guard asserts what it should: a substitution that publishes nothing is not a failure, and one that publishes values is.

Every mutation was backed up and restored with `cp` per the runbook's restore-verb rule, and the tree was verified identical to `HEAD` afterwards.

## Audit rounds

**Round 1, `code-audit-github-workflows`: refused, on one Important finding, and it was right.** The step's comment claimed a sentinel "reach[es] any such map wherever it lives", which reads as *a re-added map is caught*. `define` substitutes only at a literal reference site, so a map whose keys nothing in the story graph reads inlines nothing and the guard correctly stays green. The member proved it by execution (a `define` in a fixture config, proved loaded by throwing from it, left the grep green), and I verified its supporting claim independently: `app/utils/environment.ts`, the only reader of `npm_package_version` and `COMMIT_SHA`, has zero importers repo-wide.

That is the same defect class as #1263, in this PR's own new prose: a claim about behaviour the author never executed. My own red fixtures added a reader, so I proved the reader-bearing case and then described the reader-less one. `def56c4f` states the assertion the step actually makes and both cases are now executed.

The guard itself was kept rather than replaced. The member judges it against *detect the config regression*; #1211 asks for *nothing asserts the published preview carries none*, and a substitution with no reader publishes nothing, so the artifact is the right thing to assert. Two arms were declined: giving each guarded key a reference site would re-add the shim #1210 deleted purely to make a guard fire, and a static config check is the source-conformance arm declined when the guard's form was settled.

Both of that round's Suggestions were folded in, because the round was rotating the member's digest anyway: the `SITE_URL` sentinel and the missing-output check.

**Round 2: clean, marker written.** The member independently re-derived the red case and confirmed the guard is not fail-open.

## Accepted residuals

- **The guard's coverage is publish-only, and the step does not say so.** The step sits in the `chromatic` job, which is gated on `has-chromatic-token`, so a checkout with no `CHROMATIC_PROJECT_TOKEN` gets the workflow and never runs the guard. Recorded rather than fixed: the gating is correct rather than incidental, because the failure mode is a *published* artifact and a checkout that cannot publish cannot suffer it, so the member's first arm (move the sentinel build to `tests.yml`) would guard a build nobody uploads. What survives is its second arm, that the comment is silent on the boundary. Silence is not a false claim, so this is wording rather than the class above, and folding one comment line into an already-marked PR buys a full member re-dispatch for it.

## Out-of-scope findings

- **Filed as #1263, not waived: the rest of that `env:` block.** `SB: 1` (`:126`) has no reader anywhere in the tree and predates the `define` pipeline entirely, and the comment above `SESSION_SECRET` (`:131-133`) asserts the value exists to satisfy an import-time `schema.parse` in `app/env.server.ts` during the build. That claim is false, and this PR's own runs falsify it: a build with neither `SESSION_SECRET` nor `SITE_URL` set exits 0, which it could not do if that module were imported. `#1211` names this and asks for it to be settled separately rather than folded in, so it is not fixed here. It is filed rather than waived even though this PR changes the file: it is a false claim about behaviour the author never executed, which is the class already filed twice (`#1238`, `#1252`), so the observed-spread override applies.

Closes #1211


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.aLB3K5xwEV

PR #1264, opened 2026-08-08
debt(hooks): ask rather than deny on eslint.config edits (#1220)

### LINKED ISSUE #1220: The eslint.config guard cannot admit the sanctioned preset-spread migration without also admitting rule-silencing, so it stays a blanket deny
`.claude/hooks/block-eslint-config-edit.sh:39`

## What the guard does today

The path gate at `:39` is the entire decision: every `Edit`/`Write`/`MultiEdit` naming an `eslint.config.{js,cjs,mjs,ts}` is denied on the filename alone. That is correct and honest as it stands, and the message at `:41` says so out loud. It is also broader than the rule it serves ("fix the lint error in the source file where it occurs"), and it denies the `...lint.reactRouter` migration `CHANGELOG.md`'s **Action required** tells adopters to confirm. An agent asked to make that edit is refused and told to hand it to a human.

## The failure mode

An allowlist over *what an edit changes* is the obvious repair and it does not work. It was built and taken through six audit rounds, and the round-6 finding is what retires it: **the shape the allowlist admits turns rules off.**

Adding a bare `...lint.<group>` preset spread is exactly the sanctioned migration, and it is also a rule-silencing edit, because ESLint applies later config objects over earlier ones and the groups in `@gaia-react/lint` overlap heavily rather than partitioning the rule space. Measured against the installed package, not inferred:

| permitted edit | what it turns off |
|---|---|
| `+...lint.reactRouter` after `...lint.base` | `no-empty-pattern` (`error` → `off`) |
| `+...lint.storybook` after `...lint.react` | `react-hooks/rules-of-hooks` (`error` → `off`) |
| `+...lint.styleHygiene` after `...lint.guardrails` | `sonarjs/no-unused-vars`, `sonarjs/unused-import` |
| `+...lint.playwright` after `...lint.base` | `no-empty-pattern` |
| `+...lint.base` after `...lint.guardrails` | `import-x/no-restricted-paths` |
| `+...lint.prettier` after `...lint.base` | ~70 rules (which is what a prettier preset is for) |

13 group pairs share rule ids at all; `base` × `prettier` alone share 155, most at differing severities. So adding a preset is categorically the same act as reordering or deleting one, which the allowlist denied. **There is no repair to the comparison that separates them**: the separating property is whether the adopter wants that group's overrides, and that is not present in the text of the edit.

The first row of that table is the sharp one. The single edit the guard most needs to admit is also a rule-silencing edit, so "admit the migration" and "refuse silencing" are in direct conflict under any text-level test.

## Suggested fix

Settle the design question before writing code, because the question is what makes this hard rather than the implementation: **what evidence available to a synchronous `PreToolUse` hook distinguishes a config edit the adopter wants from one that silences a rule they wanted?**

Three candidates, each declined for the allowlist for reasons about the allowlist rather than about the candidate:

1. **Compare resolved configs.** Load the before and after config through ESLint's own `calculateConfigForFile` and deny when any rule's effective severity drops for any file. This is the only candidate that reads the property that actually matters, and it answers the table above directly. It was declined as too heavy for a synchronous hook: Node startup, a real ESLint load, and an "after" file that does not exist on disk yet. Whether that cost is real at hook latency is measurable and nobody has measured it.
2. **Ask rather than judge.** Deny with an approval path instead of a verdict, so the adopter confirms the intent the hook cannot read. Moves the decision to the only party that holds it.
3. **Narrow the admitted set to a named migration.** Admit only the specific spread the current release's **Action required** names, keyed to the CHANGELOG rather than to a shape. Small, honest, and stale by design: it needs an edit per release.

A fourth option is to keep the blanket deny permanently and close this as `wontfix`. That is a legitimate outcome, and if it is chosen the reasoning belongs here rather than being rediscovered.

Do not re-attempt a text-level allowlist without first answering the table above. That is the specific path already measured and closed.

## The specification is already written

The retired allowlist has a 49-test bats suite, and it is the artifact worth keeping: it is a written specification of the intended contract, RED-proven against the filename match and mutation-tested (5 mutants killed, 1 recorded as surviving fail-closed).

It lives on PR #1218 at commit `aec46083`, path `.gaia/tests/hooks/block-eslint-config-edit.bats`. The branch squash-merges, so reach it through the pull-request ref, which survives branch deletion:

```
git fetch origin refs/pull/1218/head
git show aec46083:.gaia/tests/hooks/block-eslint-config-edit.bats
```

What it pins, so this issue stands alone if that ref is ever unreachable:

- **The two allowed shapes**: a comment-only change, and an added bare `...<name>.<group>` preset spread, including the reactRouter migration, including on a config carrying a glob that contains `/*`.
- **The denied shapes**: an added `rules`/`ignores` entry, a removed spread, a commented-out spread, a changed called-preset option, a flipped severity at unchanged line count, a spread carrying an inline object literal.
- **Comment-prefix smuggling**: a line whose comment prefix *closes* the block and continues with executable code, in all three prefix spellings.
- **Wrapping**: neutralizing a spread by wrapping it in a block comment rather than touching it, including a text-carrying opener, an appended lone opener, and a `MultiEdit` splitting the wrap across two pairs.
- **Order as precedence**: swapping two adjacent spreads, swapping two with others between them, and hoisting one past a non-spread line.
- **Whole-file judgment**: every content test runs against a real file, because an `Edit`'s `old_string`/`new_string` are a fragment whose boundaries the caller picks.
- **Uncertainty denies**: an unreadable tool shape, missing strings, an empty `edits[]`, replaced text absent from the file, a target that exists and cannot be read.

Two known defects in that implementation, which any successor inherits as requirements rather than as code to copy:

- The subsequence comparison admits an unmatched spread line unconditionally, so **appending a duplicate of a spread already present is allowed**, and a duplicate at the end has the same effect as the move the guard denies. Repair: seed a set of the before stream's spreads and refuse a skip that duplicates one.
- The `Edit` arm's `|| deny` after `before=$(cat …; printf x)` is unreachable, since the substitution's status is `printf`'s. Not live (the `-f`/`-r` check above covers it), but knowingly dead.

Handler: spec

stevensacks:
Closes #1220

## Problem

`.claude/hooks/block-eslint-config-edit.sh` denied every edit to `eslint.config.{js,cjs,mjs,ts,mts,cts}` on the filename alone. That is broader than the rule it serves ("fix the lint error in the source file where it occurs"), and it denies the `...lint.reactRouter` preset-spread migration GAIA's own CHANGELOG **Action required** tells adopters to make: an agent asked to perform it is refused and has to hand it back to a human.

No test over the text of the edit separates the two cases, because they are the same shape. ESLint applies later config objects over earlier ones and the `@gaia-react/lint` groups overlap rather than partitioning the rule space, so adding a bare `...lint.<group>` spread is simultaneously the sanctioned migration and a way to turn rules off. `#1220` carries the measurement of that and retires the text-level allowlist that was taken through six audit rounds.

## The decision, and why it is not candidate 1

`#1220` names four arms and leaves the design question open. The maintainer settled it before any code was authored: **ask rather than judge.** The separating property is adopter *intent*, which no hook reads, so the call goes to the party that holds it rather than being guessed or refused forever.

**Candidate 1 (compare the resolved configs before and after) was measured for the first time and is dead on correctness, not only on cost.** The issue asserts it "answers the table directly". It does not:

1. **It does not separate the two cases.** Applying the sanctioned `...lint.reactRouter` migration to this repo's own config drops `no-empty-pattern` from `error` to `off` on `app/routes/_public+/_index.tsx`. A rule that denies any effective severity drop therefore denies the migration. The conflict reproduces one layer down.
2. **The answer is sample-dependent.** Same migration, same config: **0** drops for `app/root.tsx`, **1** for a route file, because the groups are glob-scoped. "Any rule for any file" needs a representative file set that cannot be enumerated for an arbitrary adopter tree.
3. **It has to execute the config it is judging**, which is arbitrary JavaScript, inside the guard deciding whether to allow it. A partial config also throws outright (`Could not find plugin "react"` for a tree carrying `reactRouter` without `react`).
4. **Its slow path fails open.** A `PreToolUse` hook past its deadline is cancelled with no `permissionDecision` at all and the write proceeds uninspected. Measured baseline is ~2.6 s wall clock (node start, eslint import, two resolves), and that is the fast case: an `eslint.config.ts` loads through jiti and a monorepo samples more files. This hook's own comment already treats "a guard whose failure mode is *allows what it matched*" as disqualifying.

Arm 3 (a named-migration allowlist keyed to the CHANGELOG) was declined as stale by design and a text-level allowlist in a narrower coat. Arm 4 (keep the blanket deny, close `wontfix`) was declined although the issue licenses it.

## Fix

The path gate is unchanged: same regex, same herestring reasoning, same silent `exit 0` for a payload naming no config. What changes is the verdict. The hook now emits `permissionDecision: "ask"` as `jq -n` JSON with `exit 0`, the structured form every sibling hook already uses, with a reason that admits the prompt is on the filename alone, names the sanctioned migration, points the common case back at the source file, tells the operator the answer is theirs, and refuses the workaround of disabling the hook.

The ask is **uniform**: every edit to a config prompts, and the hook does not pre-classify silencing from migration and ask only on the uncertain ones. That would reintroduce exactly the comparison the retired allowlist proved cannot separate them.

The hook's comment block recorded `#1220` as an open design question. It now records the answer and why the resolved-config arm is not it, so the next reader does not re-derive the measurement above.

## Verification

- `.gaia/tests/hooks/block-eslint-config-edit.bats`: 24 → 26 tests, all green. Every deny case becomes an ask case; the three message tests become reason tests.
- **Two new tests, neither decoration.** One pins that the hook never emits a deny and never blocks by exit code, because asserting the ask positively cannot see a regression to the old contract. One pins that the output parses as JSON, because the reason is machine-read and a payload that broke out of it would report no decision at all, which is the fail-open direction.
- `helpers/run-hook.sh` gains **mechanism 3**, `assert_asked_by_json`, rather than a third verdict under mechanism 2. `assert_allowed_by_json` passes on an `ask`, so a suite reaching for it to mean "not blocked" would green on a hook that had stopped asking altogether.
- **Every guard proved by mutation**, each mutant restored with `cp` from a `mktemp -d` copy rather than `git checkout --`, and the file confirmed byte-identical to `HEAD` after each:

  | Mutant | Fires |
  |---|---|
  | `"ask"` → `"deny"` | 19 tests, **including the new never-denies guard (16)** |
  | path gate drops `.mjs` | 18 tests |
  | reason loses the do-not-disable clause | test 20 alone |
  | `jq -n` → hand-built truncated JSON, ask substring intact | tests 17-21 |
  | valid reason plus one stray byte | **test 17 alone**, so the well-formedness guard is load-bearing independently of the reason assertions |

- Full CI-gated set green locally under bash 5: `bats .github/audit/tests/` 165, `bats .gaia/scripts/tests/` 1458, `bats .gaia/tests/hooks/` 1370, `bats .gaia/tests/lib/` 448, `bats .gaia/tests/statusline/` 36, `bash .gaia/tests/shell-lint.sh` clean. All 0 failures.
- Quality Gate **skipped per its own written skip logic**: no changed file matches its source or gate-affecting-config globs (`.sh`, `.bats`, `.md` only).
- `cli-tests.yml`'s distribution job does not arm: its filter is `.gaia/cli/**`, `.gaia/release-exclude`, `.gaia/release-scrub.yml`, `.gaia/tests/distribution/**`, `.github/workflows/cli-tests.yml`, none of which this diff touches.

## Docs

`wiki/concepts/Claude Hooks.md` and `wiki/modules/Claude Integration.md` both described a blanket refusal, and the hook-contract list documented only `allow`/`deny`. Both now carry the `ask` form and when to reach for it. `CHANGELOG.md` gets a `### Changed` entry that states the loosening plainly rather than selling it: what was an absolute refusal is now a prompt the adopter can approve.

One process note worth recording, because it is the failure class this repo keeps filing. The first attempt at the last mutant appended the stray byte **after** the hook's own `exit 0`, so it never executed, and the suite came back clean. Read as written, that is evidence the test is redundant; it is actually evidence of an inert mutant. It was caught by checking the mutant's tail before believing its result, and the rule it argues for is general: **a mutation that fires nothing has not proved the guard is weak until you have proved the mutation ran.**

## Audit rounds

**Round 1 (`code-audit-maintainer-shell`, the only dispatched member): clean pass.** No Critical, no Important. Two Suggestions, both in-domain on files this PR changes. **Both fixed in `3e1d3553` rather than accepted**, and the reasoning is worth stating because the pre-committed stopping rule licensed the other answer.

The PR was clean and marked with nothing else rotating the digest, so digest-economics **arm 2** (accept-and-note) applied. Suggestion 2 was taken anyway under its "consequential enough" exception: the helper's header still said "two incompatible deny contracts" and "one assertion pair per deny mechanism", **both of which this PR itself made false** by adding a third mechanism that is a single function. Its named failure mode is that the header is what a suite author reads to pick an assertion, so a stale count routes them past mechanism 3 to `assert_allowed_by_json`, which passes on an `ask`. That is a false claim about behaviour a later reader acts on, not wording polish. Once that rotated the digest, **arm 1** applied to Suggestion 1 and it rode the re-review already being paid for.

Suggestion 1 was independently found: it is the same open property this PR had already recorded as a residual before dispatch. The member supplied the evidence the PR lacked, a live in-repo call site, and on verification it is **five** wiki-sync smoke scenarios rather than the one cited.

**Round 2 (on `3e1d3553`): clean pass.** No Critical, no Important. It verified both round-1 repairs mechanically rather than by reading: stripping comments and blank lines from both files at `8eac2dca` and at `3e1d3553` yields byte-identical output, so both edits are provably comment-only. It also independently re-verified the widened citation at 6 call sites across 5 files.

Three new Suggestions, all on the comment prose round 1's repair wrote. Two fixed in `f9ea11b2`, one declined:

- **`block-eslint-config-edit.sh:58`, "settled" overstated a strict comparison.** The comment said an unattended session is where the guard is "weaker than a flat refusal" and called that settled. If the permission engine resolved an unattended ask to *deny*, the guard would be exactly as strong, not weaker. Only the non-strict bound is settled. **Fixed:** "an ask is at best no stronger than a refusal."
- **`block-eslint-config-edit.sh:53`, a dangling adopter-facing reference.** The hook is `manifest: owned` so it ships verbatim, `.gaia/tests/` is release-excluded, and the maintainer-only marker scrub strips markdown only, never `.sh`. The comment said "this repo's own wiki-sync smoke scenarios", which on an adopter clone names a directory that does not exist and reads as a claim about the reader's repo. **Fixed**; `git grep -F '.gaia/tests' -- .claude/hooks/` now returns nothing.
- **`helpers/run-hook.sh:90`, mechanism 3's banner does not name which allow form to borrow. DECLINED.** It documents practice the sole consumer already follows, so it enriches the artifact rather than correcting it, and round 1's repair adding prose is what produced these three findings in the first place. Single instance, no live failure, no spread, so recorded here rather than filed.

**The stopping rule was overridden at round 2, and the warrant is recorded rather than assumed.** The rule's second branch covers findings on prose this change authored and says accept-and-note then merge, which would have merged all three. The adopter-facing one is above that threshold: leaving it means shipping a known-false line in a file that ships verbatim and then filing it as debt, which converts a one-word fix into backlog. The precision one is the same class in miniature and rode the same digest rotation. The genuinely sub-threshold one was declined rather than swept along, so the override stays bounded to the correctness half.

## Known open property, recorded rather than claimed

**How `permissionDecision: "ask"` behaves under `bypassPermissions` / `--dangerously-skip-permissions` is not verified here, and this PR deliberately claims nothing about it in either direction.** The published hooks reference confirms `ask` is a valid decision but does not document its interaction with permission modes, and neither this row nor the audit executed the permission engine. The direction of the risk is nameable even though the behaviour is not: a refusal needs no prompt, while an `ask` does, so an unattended session is where this guard is weaker than a flat refusal, and it is also where nobody is reading the reason string. That is now recorded in the hook's own comment block, phrased as an open property.

Not filed as debt: there is no reproduction behind it, and `file-tech-debt` rejects a finding with no demonstrated failure mode. A bats suite can prove the script *emits* `ask`, which it does; whether the harness *honours* it in that mode is a Claude Code fact no suite in this repo can reach.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.2KohAAVA1u

Issue #1235, opened 2026-08-06
The git-sandbox maintenance suppression #1216 added lives in one suite, while seventeen others build the same unsuppressed sandbox


stevensacks:
`.gaia/cli/src/wiki/chain.test.ts:32` (the `git init` inside `setupSandbox`)

Surfaced by the altitude review on PR #1234, which fixed the same class in `commit-classify.test.ts` for #1216. Filed rather than folded in: the repair touches seventeen files and three shared fixture modules, which is its own change.

## Failure mode

`#1216` established that a git sandbox built by `mkdtempSync` + `git init` + a run of `git commit`s is exposed to git's own background auto-maintenance. Every `git commit` spawns `git maintenance run --auto --quiet --detach` unless the repository suppresses it, and a detached run has two ways to hurt a suite: it repacks the object store while the next `git commit` is still running, which is what corrupted HEAD in #1216, and it outlives the test into an `afterEach` that `rmSync`s the repository out from under it.

That suppression now exists in exactly one file. `grep -n 'gc\.auto\|maintenance\.auto' .gaia/cli/src` returns `commit-classify.test.ts` and nothing else. `chain.test.ts:32` is the plainest case: `execFileSync('git', ['init', '-q'], {cwd: root})` with no config at all, in a suite with twelve commit sites.

The other suites building the same unsuppressed sandbox:

- `.gaia/cli/src/wiki/sync-land.test.ts`
- `.gaia/cli/src/update/regen-regions.test.ts`
- `.gaia/cli/src/wiki/state.test.ts`, `state-init.test.ts`, `diff-size.test.ts`
- `.gaia/cli/src/ci/__tests__/revert.test.ts`
- `.gaia/cli/src/setup/__tests__/setup-worktree.test.ts`, `link-worktree.test.ts`
- `.gaia/cli/src/release/commit-and-tag.test.ts`, `manifest.test.ts`, `manifest-cli.test.ts`, `manifest-cli-check.test.ts`, `manifest-cli-args.test.ts`

And the three shared fixture modules that already own this shape, each running its own copy of `init` plus the identity config: `.gaia/cli/src/util/main-checkout-fixture.ts`, `.gaia/cli/src/automation/__tests__/sandbox.ts`, `.gaia/cli/src/setup-ci/__tests__/sandbox.ts`.

[...]

PR #1265, opened 2026-08-08
debt(cli): refuse gaia init in a tree that is not the scaffold (#1242)

### LINKED ISSUE #1242: gaia init rename rewrites whatever directory it is run from, because every precondition asks whether the rename is doable and none asks where it is running
`.gaia/cli/src/init/rename.ts:469`

## Failure mode

`gaia init rename` resolves the tree it rewrites from ambient state:

```ts
const cwd = options.cwd ?? process.cwd();
```

Every refusal the command has asks whether the rename is **doable**, never **where it is running**:

- `claudeMdPreconditionMet` (`:209`) refuses when `CLAUDE.md` carries no H1.
- `unrewritableKey` (`:414`) refuses when a language-file key appears somewhere the rewrite would look but cannot be rewritten.
- the quote-class refusal in `parseFlags` refuses a title naming both quote characters.

In any tree that already has a rewritable `package.json` and an H1-bearing `CLAUDE.md`, all of those preconditions **pass**, so the command proceeds and rewrites that tree. The GAIA repo itself is exactly such a tree. No guard anywhere in `.gaia/cli/src/init/` refuses on "this is not a fresh scaffold" or "this is the GAIA template source".

**Concrete trigger and outcome:** invoke `gaia init rename --title 'X' --kebab x` from any directory that is not the intended scaffold — the GAIA repo root, a parent project, a sibling checkout. `package.json`'s name and `CLAUDE.md`'s H1 (plus any matching language-file values) are rewritten in place, with a success exit and no confirmation, no dry-run, and no refusal.

## Observed, not hypothesised

This fired **three times in a single session** during the `#1197` drain row: once in the executing session and twice in the audit agent's, each time because a CLI probe ran with the repo root as cwd instead of a sandbox. Each was caught and reverted, and the resulting tree was verified clean (working tree clean, `main` level with `origin/main`, and PR #1241's merge commit contains the six expected files with zero renames under `--find-renames`). Nothing leaked.

The relevant point is not the damage, which was none, but that three probes run by operators holding full context all made the same mistake against a command that offers no refusal.

## Reach

`gaia init rename --title <T> --kebab <K>` is a shipped standalone subcommand present in the adopter `.gaia/cli/gaia` bundle, not maintainer-only. An adopter invoking it from the wrong directory hits the same behavior.

## Suggested fix

The correct behavior is genuinely open and this issue deliberately does not prescribe one. The arms, all reasonable:

1. **Refuse on a dirty git tree.** Cheap and catches the observed case, but refuses a legitimate rename in a scaffold with uncommitted work.
2. **Refuse when init state says this step already ran.** Reuses the existing `markStepCompleted` / step-state mechanism (`:524`), so it fits the module's conventions, but does nothing in a tree that never ran `gaia init` at all — which is the actual misfire case.
3. **Require an explicit `--cwd`** rather than defaulting to `process.cwd()`. Removes the ambient dependency outright; a breaking change to a shipped subcommand's interface.
4. **Prompt for confirmation** naming the resolved path. Safe interactively, wrong for the non-interactive scripted path `gaia init` drives.
5. **Change nothing.** It is a scaffolding command; pointing it at the wrong tree is user error, and every tree it can fire in is under git, so the damage is a `git checkout` away.

Arms 1, 2 and 4 are also not mutually exclusive, and there is a prior question none of them settles: whether the guard belongs on each `init` step or once at the `gaia init` entry point.

Handler: plan

stevensacks:
Closes #1242

## Problem

`gaia init rename` resolves the tree it rewrites from ambient state, and every refusal it has asks whether the rename is *doable*, never *where it is running*. In any tree with a rewritable `package.json` and an H1-bearing `CLAUDE.md`, all preconditions pass and the tree is rewritten in place, at exit 0, with no confirmation and no dry run. It fired three times in a single session during the `#1197` row, each time because a CLI probe ran with the repo root as cwd.

`rename` is not special. All eight `init` steps resolve the target identically as `options.cwd ?? process.cwd()`, and several besides `rename` are destructive (`strip-branding`, `configure-i18n`, `finalize`).

## Decision

The issue deliberately prescribed nothing: five arms plus a prior question none of them answers. Both were settled by the maintainer before authoring, in one question.

**Guard arm: refuse unless the target is a GAIA scaffold that is not GAIA itself.** `.gaia/manifest.json` must be present, `.gaia/cli/src/` must be absent. Verified rather than assumed: `.gaia/cli/src` is 0 of 639 manifest entries, present in this repo, absent from a real adopter tree, and `.claude/rules/gaia-folder.md` already uses that exact discriminator in prose.

- Arm 1 (dirty git tree) declined: it guards the wrong variable. It would have missed any probe run from a clean repo root, and it false-refuses a legitimate rename in a scaffold with uncommitted work.
- Arm 2 (init state) declined by the issue itself: it does nothing in a tree that never ran `gaia init`, which is the actual misfire case.
- Arm 3 (`--cwd`) declined: `--cwd` is not a flag today at all, only a programmatic test seam, so it would add a required flag to a shipped subcommand plus a matching change to the shipped `/gaia-init` skill.
- Arm 4 (confirmation) declined: wrong for the non-interactive path `/gaia-init` drives.
- Arm 5 (change nothing) not taken.

**Guard site: once at the `gaia init` router**, not in `rename.ts`. A `rename`-only fix leaves seven siblings carrying the identical defect, which is the one-of-N shape already filed three times on this backlog (#1229, #1232, #1235). The issue's own prior question names the entry point as a candidate site, so this is inside its scope rather than beyond it.

## Fix

`init/index.ts` refuses before dispatch, with a distinct code and message per case, each naming the path it resolved:

- `not_a_gaia_project` when `.gaia/manifest.json` is absent (a parent project, an unrelated sibling checkout).
- `gaia_template_source` when `.gaia/cli/src` is present (GAIA itself rather than a project made from it).

Two deliberate exemptions, both pinned by tests: a per-step `--help` request stays readable from anywhere because it writes nothing, and an unknown subcommand still reports `unknown_subcommand` rather than a target refusal.

The router also hands the step the cwd it guarded. Without that the guard validates one path while the step independently re-resolves another from `process.cwd()`, so the guard's subject and the rewrite's subject would agree only incidentally. `resume.ts:277` already passes `{cwd}` this way, so this follows the module's existing convention.

`.gaia/cli/gaia` and `.gaia/cli/gaia-maintainer` are rebuilt: the adopter binary is a committed artifact, so the guard reaches an adopter only once the bundle carries it.

## Verification

The load-bearing question is whether the guard can brick a fresh scaffold. It cannot, and this was checked rather than reasoned: `.gaia/manifest.json` is an explicit ships-in-the-tarball sentinel (`release/manifest.ts` `ADOPTER_OWNED_SENTINELS`), and `create-gaia` extracts that tarball into the target directory, so a fresh scaffold carries it before `gaia init` ever runs.

- `bash .gaia/tests/distribution/run-all.sh`: **17/17 PASS, 0 FAIL**. Scenarios 07 and 08 drive `gaia init` against a staged adopter tree, so they are the empirical confirmation above. Both were re-run **after** the bundle rebuild; the first run tested the old bundle and proved nothing.
- Quality Gate: root `typecheck`, `lint`, `test --run` (30 files / 156 passed), `pw` (8 passed), dev smoke (HTTP 200), `build` all green. CLI `typecheck`, `lint`, and full vitest (134 files / 2027 passed) green.
- CI bats roots under `bats5`: `.github/audit/tests/`, `.gaia/scripts/tests/`, `.gaia/tests/hooks/`, `.gaia/tests/lib/`, `.gaia/tests/statusline/` all with zero `not ok`. `bash .gaia/tests/shell-lint.sh` passed; `region-marker-conformance.bats` clean.
- **Guard proofs: six mutations, every one of the nine new tests provably killable.** Drop the not-a-GAIA-project refusal → 3 red. Drop the template-source refusal → 1 red. Stop handing the step the guarded cwd → 2 red. Guard the help path → 1 red. Route an unknown subcommand to a real step → 1 red. Drop the router help early-return → 1 red. Each mutation was restored by copying the file back rather than with `git checkout --`, so no restore reached beyond its own mutation.

Worth recording: the third mutation reproduced this issue's own defect inside the harness. With the pass-through removed, `bootstrap-env` wrote its state file into the CLI package directory instead of the sandbox.

## Scope

`util/repo-root.ts` was considered for reuse and deliberately not used: it shells out to git, and it would *redirect* the rewrite to the repo root rather than refuse it, which is a worse answer than refusing. A subdirectory invocation now gets a clear refusal naming the path, where previously it produced an undefined rewrite.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.D80FIoVPsv

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Q2jqw3cXYo

PR #813, opened 2026-07-18
fix(audit): cover the CLI build/config surface + drain orphaned ulid

stevensacks:
## What

Two related tech-debt fixes, both anchored at `.gaia/cli/package.json`, resolved together on one branch.

### #812 - CLI build/config surface in no audit member's globs (severity: important)

`code-audit-maintainer-node` claimed only `.gaia/cli/src/**`, so an edit to the bundle build scripts and runtime deps (`.gaia/cli/package.json`), the resolved dependency tree (`.gaia/cli/pnpm-lock.yaml`), or the build config (`.gaia/cli/tsconfig*.json`) reached the required merge gate with **no** code-audit review. That is exactly how the bundle-shebang shell-portability defect shipped to `main` during the SPEC-045 work.

- Extend `code-audit-maintainer-node`'s globs to cover those three surfaces, in **both** `.gaia/audit-ci.yml` and the built-in fallback roster in `.claude/hooks/lib/audit-scope.sh` (kept in lockstep by `verify-audit-roster.sh`'s fallback invariant).
- The three new globs are disjoint from every other claimant, so `verify-audit-roster.sh` stays clean.
- Add three resolver regression tests (built-in-roster path, one file each) locking the new ownership.

### #764 - Orphaned `ulid` dependency (severity: suggestion)

`ulid` was left standing when SPEC-038 deleted its only consumer. Nothing imports it, so `pnpm -C .gaia/cli remove ulid` drops it from `package.json` + `pnpm-lock.yaml` and the rebundled binaries are **byte-identical** (proving it was unreachable). Landing it here means the `package.json` edit itself is now audited under the coverage #812 adds.

## Verification

- `resolve-audit-members.bats`: 27/27 green (3 new).
- `verify-audit-roster.sh` + `verify-audit-roster.bats`: clean (fallback-lockstep, disjointness, digest).
- `audit-write-clearance` / `resolve-audit-spawn` / `audit-guard-structural` / `audit-machinery-complete` bats: green (digest rotation).
- CLI `typecheck` / `lint` / `test` (1447 passed): green. Binaries byte-identical.

Closes #812
Closes #764

PR #1267, opened 2026-08-08
debt(cli): guard every shared rule-bearing package for lockfile parity (#1205)

### LINKED ISSUE #1205: The lint rule-set parity guard covers one package, while eight more drift the same way and two already have

stevensacks:
Closes #1205

## Problem

`.gaia/cli` is its own pnpm workspace root, so it inherits nothing from the repository root and every rule provider exists twice with nothing holding the copies together. The lockfile-resolution guard pinned one package name, `typescript-eslint`, but the thing that lets a package drift is **the arrival path, not the package**: anything reaching both workspaces through `eslint-config-airbnb-extended`'s caret is frozen independently by each lockfile, and can therefore sit on two versions with every existing assertion green.

Measured rather than estimated: **30 rule providers resolve in both lockfiles**, the guard watched **one**, and **two had already drifted**.

| package | root | `.gaia/cli` |
|---|---|---|
| `eslint-plugin-import-x` | 4.16.2 | 4.17.1 |
| `@next/eslint-plugin-next` | 16.2.6 | 16.2.12 |

## Fix

**Which packages earn parity is now a pattern, not a list.** A guarded package is one present in **both** lockfiles whose name matches the npm convention for a rule provider (`@scope/eslint-plugin*`, `@scope/eslint-config*`, `eslint-plugin-*`, `eslint-config-*`, plus the bare `typescript-eslint` meta-package). **Unlisted means guarded**, so a plugin arriving later through the same caret is covered with no edit.

A hand-written inclusion list was the alternative and was declined, because it makes the default silence. That is the failure mode this repository has already paid for twice on the audit roster: #813 claimed to cover the CLI config surface and missed four files, then #1243 added one of those and missed three.

The escape hatch is `PARITY_EXEMPT`, whose entries carry their reason as data. A hygiene test reds when an exemption stops naming a package both workspaces install, so it cannot outlive its subject.

**The two live divergences got opposite dispositions, and both were decided by execution rather than by reading.** `eslint --print-config` was run in each workspace:

- **`eslint-plugin-import-x` is converged.** It is loaded with **23 enabled rules in both** workspaces, so a difference between 4.16.2 and 4.17.1 is a real difference in what each workspace considers an error. Root was the lagging side and re-resolves to 4.17.1, which is the repair the guard's own docblock prescribes (re-resolve the lagging workspace; do not reach for an exact pin).
- **`@next/eslint-plugin-next` is exempt.** It appears in **neither** workspace's resolved config: it is installed transitively and never loaded, so its versions can differ without changing a single rule. Converging it would also not hold, since the next re-resolve floats it apart again and reds a required check over rules nobody runs.

`eslint --print-config` was used as evidence here rather than wired in as a runtime check, and that is a deliberate call. The job that runs this guard installs `.gaia/cli` alone (`cli-tests.yml`), so root `node_modules` never exists in it; computing the root workspace's config would mean adding a full root install to a required context to compare version strings. Independently, six loaded plugins expose no `meta.name`, and `eslint-plugin-react-hooks` reports a **stale** `meta.version` (`7.0.0` against a resolved `7.1.1`), so plugin metadata is not a usable version source either.

### Lockfile scope, disclosed rather than trimmed

`pnpm update eslint-plugin-import-x --depth Infinity` re-resolves siblings, so the lockfile diff carries five bumps beyond its subject: `@napi-rs/wasm-runtime` 1.1.6 → 1.2.2, `baseline-browser-mapping` 2.11.5 → 2.11.9, `brace-expansion` 1.1.16 → 1.1.18 and 5.0.8 → 5.0.9, `electron-to-chromium` 1.5.397 → 1.5.399, and `@package-json/types` dropped (import-x 4.17 no longer depends on it).

**They are pnpm's resolution, not pre-existing staleness:** a plain `pnpm install` against the reverted lockfile leaves it byte-identical, so `update` produced them. They are not hand-trimmed, because a lockfile edited by hand is one no tool would generate. Note also that `pnpm update eslint-plugin-import-x@4.17.1 --depth Infinity` rewrote 55 lines and **did not move import-x at all**; the unpinned form is the one that works.

`pnpm-workspace.yaml` is untouched by the re-resolve, which is the check that matters under `minimumReleaseAgeStrict: true`: a breach of the 7-day window writes exemptions back into that file rather than failing, and none appeared.

### Prose the change falsifies

Three sites described the guard as covering `typescript-eslint` alone: the `pnpm-lock.yaml` paths-filter comment in `cli-tests.yml`, `wiki/decisions/pnpm.md`, and `wiki/dependencies/gaia-lint.md`. Each is what a reader consults when deciding whether that filter entry or the guard can be pruned, so a description narrower than the guard is the defect rather than wording polish. The guard's own docblock, which named #1205 as its tracked follow-up, is rewritten rather than left standing.

## Verification

**The guard was proved RED on the live divergence before anything was converged**, naming `eslint-plugin-import-x` with both versions across a 29-package comparison, with `@next` correctly absent from it.

**Nine mutants, nine deaths.** Each mutation was restored with `cp` from a `mktemp -d` backup, never `git checkout --`, and both mutated files verified byte-identical to their commit afterwards.

| Mutation | Died as |
|---|---|
| regex: scoped arm dropped | pattern-arm test (and the exemption-hygiene test, since `@next` leaves the shared set) |
| regex: `eslint-plugin-*` arm dropped | pattern-arm test |
| regex: `eslint-config-*` arm dropped | pattern-arm test |
| regex: `typescript-eslint` arm dropped | `typescript-eslint` presence test |
| regex: matches nothing | floor, all three arms, and the presence test |
| `PARITY_EXEMPT` emptied | parity test, naming `@next` (proves the exemption is load-bearing **and** that `@next` is genuinely diverged) |
| `PARITY_EXEMPT` given a stale entry | exemption-hygiene test |
| root lockfile: one guarded version changed | parity test, naming the package and both versions |
| root lockfile: a second copy of one package | single-copy test |

**Quality Gate**, all green: `pnpm typecheck`, `pnpm lint`, `pnpm lint:cli`, `pnpm -C .gaia/cli typecheck`, root `pnpm test --run` (30 files, 156 passed, 1 skipped), CLI `pnpm test --run` (134 files, 2029 passed, 1 skipped), `pnpm build`, `pnpm pw` (8 passed, 1 skipped), dev smoke HTTP 200. **`pnpm lint` was re-run with `node_modules/.cache/eslint` deleted**, because a cached lint can report green without re-linting a single file under a new plugin minor; that uncached run is what says import-x 4.17.1 brings no rule fallout. The gate caught five lint errors in this change's own new code, all fixed before the first commit.

**The gate this diff actually arms**, checked against the workflows rather than inferred: `audit-ci-tests.yml`'s `.github/workflows/**` catch-all arms all five bats roots, and `cli-tests.yml`'s filter matches `.gaia/cli/**` and `pnpm-lock.yaml` for both its jobs.

- `bats .github/audit/tests/` 165/165
- `bats .gaia/scripts/tests/` 1458/1458
- `bats .gaia/tests/hooks/` 1370/1370, `bats .gaia/tests/lib/` 448/448, `bats .gaia/tests/statusline/` 36/36
- `bash .gaia/tests/shell-lint.sh` exit 0
- `bash .gaia/tests/distribution/run-all.sh` **17 PASS / 0 FAIL**, read from the per-scenario summary rather than the exit code alone
- the committed-binary and template freshness check: `gaia`, `gaia-maintainer` and `templates/` all identical after a rebuild

All bats runs went through `.gaia/scripts/bats5.sh` so local matches CI's bash 5.

The root suite prints a jsdom `HTMLCanvasElement.getContext()` notice. It cannot originate here, since this diff changes ESLint plugin resolution and one CLI test file and the root suite loads neither, so it is reported as pre-existing rather than claimed verified against `main`.

## Audit rounds

Three rounds. **Zero Critical and zero Important in every round.** Round 1 dispatched the whole resolved set; rounds 2 and 3 re-dispatched `code-audit-maintainer-node` alone, because the repairs touched only its glob plus `CHANGELOG.md` and `wiki/**`, which rotate no digest. That was confirmed with `resolve-audit-spawn.sh` each round rather than assumed.

**Every finding across all three rounds was a false claim in this change's own prose, while the code audited clean throughout.** All were fixed rather than accepted, so this PR carries no accepted residuals.

**Round 1** (`code-audit-frontend`, `code-audit-github-workflows`, `code-audit-maintainer-node`), three Suggestions:

1. The exemption map's rationale was described as data "so a stale entry can explain itself in the failure message", and nothing read it. The hygiene test now asserts over entries, and a mutation confirms the reason reaches the failure output.
2. The intersection was justified by claiming root resolves providers `.gaia/cli` never installs. **Both lockfiles resolve the identical 30**, which this PR's own earlier measurement had already shown. The comment now states the measured truth, and the gap behind it is closed: a provider *leaving* one lockfile dropped out of the intersection, taking its assertion with it rather than failing, and 30 → 29 clears a floor of 20. A new assertion compares the populations as sets. Mutation: renaming one provider in the root lockfile reds it (30 against 29) **while the version parity test stays green**, which demonstrates the silent narrowing rather than arguing it.
3. The CHANGELOG and both wiki pages said the guard covers every rule provider the workspaces share; it deliberately excludes the tools each workspace declares for itself. All three now name that boundary.

**Round 2** (`code-audit-maintainer-node`), two Suggestions, both on prose round 1's own repair had written:

1. The docblock said the guard covers every package arriving through the caret. It covers those that also announce themselves as rule providers *by name*, and the import resolvers are the counterexample: same arrival path, they decide whether `import/no-unresolved` can resolve at all, and they are diverged today. Verified against both lockfiles rather than taken on report. Filed as **#1269**; the docblock now names the boundary and points there. **Widening the pattern to a fourth family was deliberately not done in-round**, because the selection criterion is the decision this PR exists to settle and its arms are not equivalent.
2. The exemption map said a named package "is not compared". The population assertion added in round 1 compares raw name sets with no exemption filter, so only the *version* is waived. Corrected, and proved: removing the exempt package from one lockfile reds both the population test and the hygiene test.

**Round 3** (`code-audit-maintainer-node`): clean, zero findings at every severity.

### Debt filed

- **#1268** — the two workspaces declare different exact pins for `eslint` (9.39.5 / 9.39.4) and `prettier` (3.9.6 / 3.9.1, with root resolving two copies), and neither guard covers a declared tool version. `prettier` reaches lint output through `eslint-plugin-prettier`. Re-verified by hand against both manifests and both lockfiles. It also carries the one claim this PR deliberately left standing: the docblock at `:22-25` says the tool versions "move without changing what either workspace considers an error", which is doubtful for `prettier`. **No formatting comparison between 3.9.1 and 3.9.6 was executed**, so correcting that sentence would replace one unverified claim with another; it is recorded there instead.
- **#1269** — the naming convention excludes the import resolvers, which arrive by the same caret and are diverged (`eslint-import-resolver-typescript` 4.4.4 / 4.4.5; `eslint-module-utils` carrying a second copy at root).

Both are `surface:maintainer`: an adopter has no `.gaia/cli` workspace, so neither divergence is observable from an adopter clone.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.1IEmYLjQEz

Issue #1271, opened 2026-08-08
The CI-template contract guard cannot see a quoted invocation, the same assumption #1037 fixed in the sibling reachability guard


stevensacks:
**Location:** `.gaia/cli/src/automation/__tests__/gaia-ci-template-refs.test.ts:163`, the `template.includes(\`gaia ${command}\`)` call inside `collectMissingReferences`.

## Failure mode

The CI-template contract guard decides whether a declared CLI reference still appears in its `gaia-ci-*.yml.tmpl` template by testing for the literal substring `gaia <command>`. That requires the byte immediately after the binary name to be a space, so a template that invokes a declared command through a **quoted or path-prefixed** binary satisfies the contract in fact and fails it in the check.

Demonstrated rather than asserted, against the live `wiki` contract entry (`cli: ['wiki sync land']`):

```js
const command = 'wiki sync land';
'      - run: gaia wiki sync land --branch-aware'.includes(`gaia ${command}`);
// true
'      - run: "$LATEST_DIR/.gaia/cli/gaia" wiki sync land --branch-aware'.includes(`gaia ${command}`);
// false  <-- reported as a missing reference
```

The bad outcome is a false red that misdirects: `collectMissingReferences` pushes `wiki: gaia wiki sync land` onto `missing`, the guard fails, and it tells the author a contract reference has been dropped from a template that in fact still invokes it. The natural repair is to edit the contract or the template to satisfy the checker, i.e. to change working code to appease a guard that is wrong.

**This is latent today, and the issue is filed on that basis.** Both declared `cli` references (`update-deps run`, `wiki sync land`) render bare in the current templates, because those workflows invoke the `$PATH`-resolved binary rather than a release-resolved path. No template carries a quoted invocation, so nothing is red now.

## Why it is worth a record anyway

`#1037` fixed the identical assumption in the sibling guard `.gaia/cli/src/command-reachability.test.ts`, where it was **not** latent: `/update-gaia` invokes two subcommands through a quoted `"$LATEST_DIR/.gaia/cli/gaia"` path, and the guard read them as dead. The invocation form that broke that guard is a real, frozen contract in this repo, so a future `gaia-ci-*` template needing a release-resolved invocation is the same event arriving at the second guard. Reconciling once rather than rediscovering per feature is the point.

## Suggested fix, and the argument against it

[...]

PR #1272, opened 2026-08-08
debt(cli): let the reachability guard read a quoted invocation (#1037)

### LINKED ISSUE #1037: The CLI reachability guard cannot see a quoted release-resolved invocation, so every future subcommand invoked only that way rediscovers the constraint
**Location:** `.gaia/cli/src/command-reachability.test.ts:229` (the `new RegExp` construction inside `isReachable`)

> **Note on the citation.** This finding has no single natural `path:line` — it is a tension between a guard's assumption and an invocation contract, spread across the guard, its docstring, and the wiki. The line above is the **most representative** site: it is where the assumption is actually encoded.

## Failure mode

The reachability guard requires every map-dispatched leaf command to appear somewhere in an invocation-shaped string: the binary name, then **whitespace**, then the command path. The pattern at :230 is:

```
(?<![\w-])gaia(?:-maintainer)?\s+<tokens>(?![\w-])
```

A quoted, release-resolved invocation such as `"$LATEST_DIR/.gaia/cli/gaia" update merge-region` can never satisfy it, because a closing double quote sits exactly where the regex requires whitespace.

The region-aware update work froze precisely that quoted form as its invocation contract, and the guard's own docstring rules out the `INTERNAL_COMMANDS` allowlist as the remedy for a command that is genuinely live. The branch worked around it by documenting each new subcommand's bare invocation form in the wiki — a genuine invoker surface, so the workaround is legitimate rather than a cheat — but the tension recurs for any future subcommand that is only ever invoked through a quoted release-resolved path.

Two independent guards on that branch shared the same unquoted-invocation assumption, which is what makes this worth reconciling once rather than rediscovering per feature.

## Suggested fix

Teach the guard to recognise a quoted invocation prefix (a binary path ending in `gaia` or `gaia-maintainer`, optionally closed by a quote, before the required separator), so a release-resolved quoted invocation counts as reachability evidence on its own.

The design decision this carries: what exactly counts as an invoker surface. Widening the regex admits any string that merely *looks* like an invocation, which weakens the guard; keeping the wiki-documentation workaround keeps the guard tight but obliges every future feature to remember it. The surrounding code does not settle which the guard is meant to be.

Handler: plan

stevensacks:
Closes #1037

## Problem

`command-reachability.test.ts` proves every map-dispatched CLI leaf command has an external invoker by scanning seven repo surfaces for an invocation-shaped string. Its pattern required whitespace immediately after the binary name:

```
(?<![\w-])gaia(?:-maintainer)?\s+<tokens>(?![\w-])
```

A **release-resolved** invocation puts a closing quote exactly there:

```
"$LATEST_DIR/.gaia/cli/gaia" update merge-region
```

That form is a frozen invocation contract, not a stylistic choice: `/update-gaia` resolves these subcommands from `$LATEST_DIR` so an adopter whose installed binary predates the subcommand can reach it at all, and `.gaia/tests/distribution/17-gaia-update-merge-region.sh` pins it. The guard could not see it, so it read a live command as dead.

**The defect is narrower than the issue frames it.** An *unquoted* path prefix already matched, because `/` is neither `\w` nor `-` and the lookbehind admits it. The closing quote is the whole of the blind spot. Verified before cutting the branch (the issue carries `debt:pre-provenance`, so DRAIN's premise-reproduction flag was up):

| invoker text | matched |
|---|---|
| `gaia update merge-region` | yes |
| `$LATEST_DIR/.gaia/cli/gaia update merge-region` | yes |
| `"$LATEST_DIR/.gaia/cli/gaia" update merge-region` | **no** |

`update merge-region` and `update regen-regions` are green on `main` only because their bare forms happen to appear in `wiki/concepts/Update Workflow.md` and `wiki/concepts/Generated Regions.md`. Those two lines are load-bearing and unmarked: rewriting either page to show the actual quoted contract would red CI, and the guard's own comment sends the next person to `INVOKER_SURFACES`, which is the wrong place.

## Fix

Admit an optional quote **beside** the separator, never instead of it, and extract the matcher so the pattern is testable in isolation.

```diff
-    String.raw`(?<![\w-])gaia(?:-maintainer)?\s+${tokens}(?![\w-])`
+    String.raw`(?<![\w-])gaia(?:-maintainer)?["']?\s+${tokens}(?![\w-])`
```

**This is not a loosening.** The file's own docstring already declares the oracle's floor: *"an invocation-shaped string in operator-facing prose … counts as reachable; that is the intended floor, the target is the command referenced by nothing at all."* An invocation that is actually **executed** is stronger evidence than one merely written down, so admitting it is what the guard already says it means.

`isReachable` closed over the module-level `invokerText`, which cannot distinguish *"this form is unmatchable"* from *"no such invocation exists here"*. It is now a one-line delegation to `matchesInvocation(commandPath, text)`.

### The decision this carried

`#1037`'s stated fork is *what counts as an invoker surface*, and both arms were settled by the maintainer before a line was authored.

1. **A quoted release-resolved invocation counts.** Rejected: keeping the guard tight and continuing to document bare forms (the workaround is legitimate, but its two wiki lines are load-bearing and unmarked); and widening to any binary path prefix (admits grep hits, stack traces and log lines, the real weakening the issue warns about, with no observed instance).
2. **Fix this guard only; file the sibling.**

The quote **class** (`["']?` rather than the precedent's `"?`) is an authoring call, not part of the settled decision. The precedent only ever sees a double quote because `$LATEST_DIR` cannot expand inside shell single quotes; this guard also scans markdown, YAML and prose across seven surfaces, where a single-quoted literal path is ordinary. It costs one character and closes the family rather than one member.

### Prior art, and it is not a coincidence

`.gaia/tests/distribution/17-gaia-update-merge-region.sh:307-312` already matches `\.gaia/cli/gaia"? update …`, and its comment states this exact rationale: *"Match the invocation whether or not a quote closes the binary path … a quoted working-tree invocation puts `"` where the space is expected and goes unseen."* That was written by the same branch that froze the quoted contract, so the repo had already solved this once, in a different tree, without the guard learning it.

## Verification

**TDD.** Both new tests observed red (`expected false to be true`) against the unwidened pattern before the fix, green after.

**Quality Gate**, run from `wiki/decisions/Quality Gate.md`: typecheck, lint, lint:cli (5 `prettier/prettier` errors auto-fixed, re-run clean), app tests 156 passed / 1 skipped, **CLI tests 2033 passed / 1 skipped across 134 files** (2031 before, so exactly the two new tests), build, Playwright 8 passed / 1 skipped, dev smoke HTTP 200. `simplify` ran as four parallel angle agents; reuse, efficiency and altitude returned clean, simplification found two duplicated rationale paragraphs and both were applied.

**Mutation, three mutants, all killed**, restored with `cp` from a `mktemp -d` and the tree verified byte-clean after each:

| mutant | result |
|---|---|
| drop `["']?` | both new tests red |
| `["']?\s+` → `["']?\s*` (quote *instead of* the separator) | fixture test red |
| drop the trailing `(?![\w-])` boundary | fixture test red |

The last two matter: they prove the **negative** assertions are load-bearing, so the widening cannot be over-general and still green.

**The first mutation attempt was a lying instrument and is worth recording.** Three `perl -0pi` substitutions silently failed to match, so all three "mutants" were no-ops and each reported `9 passed`. Caught only because the script also printed an occurrence count. Redone in `python3` with a hard assert that exactly one site matched, plus `git diff --stat` proving the file changed on every mutant.

**Prose sweep by identifier, not by sentence.** Grepped `command-reachability`, `INVOKER_SURFACES`, `INTERNAL_COMMANDS` and `isReachable` rather than the wording of the rule. Two other sites state something about invoker reachability (`gaia-ci-template-refs.test.ts:41`, `wiki/decisions/Folding Shell Scripts into the CLI Binary.md:61`); both remain true under this change and are untouched. The one this change owns, the guard's own docstring, is updated.

## Scope

One file plus the CHANGELOG. No production code: `command-reachability.test.ts` is a maintainer guard and `.gaia/cli/src` is release-excluded, so no adopter clone is affected and the committed bundles are unchanged (a `.test.ts` is not in either binary's entry graph, so no rebuild is owed).

## Filed, not fixed here

**#1271**, `severity:suggestion` / `difficulty:hard` / `surface:maintainer`. `gaia-ci-template-refs.test.ts:163` does `template.includes(\`gaia ${command}\`)` and shares the identical assumption. It sits on a path this PR does not change, so the touched-file waive is unavailable and it files.

It is **dormant**: both `TEMPLATE_CONTRACT` cli refs render bare today, because those templates invoke the `$PATH`-resolved binary. Demonstrated rather than asserted, against the live `wiki` entry:

```js
'      - run: gaia wiki sync land --branch-aware'.includes('gaia wiki sync land');                       // true
'      - run: "$LATEST_DIR/.gaia/cli/gaia" wiki sync land --branch-aware'.includes('gaia wiki sync land'); // false
```

Graded `hard` because that file's docstring at `:32-39` argues **against** the obvious repair: it keeps the CLI half a literal substring check on purpose, since extracting invocations from template prose yields false positives. Whether that argument reaches the *verification of a declared string* is the crux, and one of the three arms is *change nothing*.

Fixing it here was considered and declined: there is no live red, the test would need a synthetic fixture, and the deeper "both guards share one matcher" form spans two unrelated test areas with no production consumer.

## Audit rounds

**Two rounds, both clean passes.** Zero Critical and zero Important in each; one Suggestion each. Every finding landed on this PR's own new test scaffolding, never on the fix.

- **Round 1** found that the skills-surface test's stated isolation was not self-enforcing: its comment claims it isolates the *quoted* call site, but it asserts through `matchesInvocation`, which accepts bare and quoted alike, so the isolation held only because no skill happens to carry a bare form. **Fixed in `64476bdd`**, overriding this row's own pre-committed stopping rule and paying the re-dispatch, because it is a correction rather than an enrichment: the comment already claimed the property, and my own mutation proof had killed that test only by corpus accident. The control is derived from a shared `invocationPattern(commandPath, quoteClass)` builder rather than hand-copied, so it cannot drift from the matcher it controls for, and it is mutation-proven (giving the control the quote class turns it red).
- **Round 2** re-verified that repair independently, confirming `''` reproduces the pre-widening regex exactly, and cleared.

## Accepted residuals

**`.gaia/cli/src/command-reachability.test.ts:592`, round 2's Suggestion: the negative control's haystack is wider than the property it protects.** `skillsText` is the whole `.claude/skills` tree, while both quoted call sites live in one file (`update-gaia/SKILL.md:370,573`). An unrelated skill that later documents a bare `gaia update merge-region` in prose would red this test without the protected property having changed, and the red would land on an author who never touched the CLI. The remedy is to build `skillsText` from `.claude/skills/update-gaia` instead; the member confirmed all four assertions stay green under that narrowing.

**Accepted rather than fixed**, under `PR Merge Workflow.md`'s digest economics and this row's pre-committed stopping rule. It is round 2's finding on round 1's own repair, which is the enriching-rather-than-correcting shape that rule exists to stop; the failure direction is **closed** (a noisy red, never a silent green), so nothing ships unguarded; there is no live failure mode; and the in-file comment already prescribes this exact remedy for whoever hits the first red.

One correction to my round-1 reasoning, recorded so a later reader does not re-litigate it. I rejected narrowing then, on the ground that it relocates the contingency rather than removing it, since a bare form could appear in `update-gaia/` too. That still holds, but it misses the real gain: narrowing shrinks the false-alarm **blast radius** from any of ~40 skills to the one skill that actually contains the call sites, so the red lands on someone who can act on it. Take the remedy on the first red.

## Observation, not a finding

`pnpm test --run` emits a jsdom notice (`Not implemented: HTMLCanvasElement's getContext()`). It is pre-existing on `main` and cannot be from this change, which touches zero app files; it is not an i18n, HydrateFallback, or React warning. Recorded because two gate runs classified it differently. No failure mode, so nothing to file.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.jwTQdNEnoM

stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.GeRZgEpUqQ

Issue #1273, opened 2026-08-09
The react-perf StrictMode canary asserts a 1.1 wall-clock ratio against an observed 1.2, so it reds CI on timing noise


stevensacks:
**Location:** `.playwright/e2e/react-perf-smoke.spec.ts:149`, the closing assertion of
`noStrict bypass disables StrictMode (render-time inflation collapses)` (test begins at `:111`).

## Failure mode

The test asserts a ratio between two wall-clock render-time aggregates measured in a live browser:

```ts
const strictAvg = (strictA.total + strictB.total) / 2;
const relaxedAvg = (relaxedA.total + relaxedB.total) / 2;
expect(strictAvg).toBeGreaterThan(relaxedAvg * 1.1);
```

`total` is the sum of every captured record's `totalTime` for one page load. When the machine is
busy, the StrictMode double-invoke signal is swamped by scheduling noise, `strictAvg` lands at or
below `relaxedAvg * 1.1`, and the test fails. Playwright reports it as
`1 failed` and the whole `pnpm pw` job reds.

**The margin is the defect.** The test's own comment records the expected effect size:
`(observed ratio ~1.2)`. The assertion demands `> 1.1`. So the entire tolerance is ~0.1 of a ratio
derived from wall-clock timings, and the test already concedes the measurement is noisy: it
interleaves two loads per mode specifically to "damp single-run jitter", which is a mitigation for
this exact variance rather than a cure.

**Measured, not asserted:** 2 failures in 11 full `pnpm pw` runs on one branch, against 0 in 19
runs on `main` from the same session (10 of those immediately after a fresh `pnpm build`). Both
failures were this one assertion. The branch/main split is sampling noise rather than signal: the
branch under test changed only comments under `.gaia/cli/src` plus one new vitest file, none of
which this spec imports or executes.

**This reds CI, not just local runs.** `.github/workflows/tests.yml` runs Playwright, and a CI
runner is a noisier timing environment than a developer laptop, so the observed local rate is a
floor rather than a ceiling.

## Suggested fix

The prior question is whether a wall-clock ratio is the right oracle for "the StrictMode
double-invoke is gone" at all, and the code does not settle it. Three directions, none prescribed:

1. **Raise the sample count** and compare medians rather than means of two, which is what the
   existing two-loads-per-mode interleave is already reaching for. Reduces the rate; does not make
   the assertion deterministic, and a flake's fix established statistically is expensive to verify.
[...]

Issue #1034, opened 2026-07-22
region-registry.ts module docblock sits below the first import, so it reads as JSDoc for the node:fs import


stevensacks:
**Location:** `.gaia/cli/src/release/region-registry.ts:1`

## Failure mode

The `import {load as parseYaml} from 'js-yaml';` statement sits at line 1, **above** the module docblock that begins at line 2. Because a JSDoc block attaches to whatever follows it, the file's module-level documentation now reads as the doc comment for the `node:fs` import beneath it rather than for the module.

Almost certainly an import-ordering autofix artifact. Every sibling module in `release/` and `update/` leads with its docblock before any import.

Cosmetic — nothing executes differently — but it misattributes the file's primary explanation, and any tooling or reader that resolves JSDoc by adjacency gets the wrong answer.

## Suggested fix

Move the docblock above all imports, matching the sibling modules.

Handler: prompt

PR #1274, opened 2026-08-09
debt(cli): guard that a module docblock precedes the first import (#1163)

### LINKED ISSUE #1163: 80 CLI module docblocks sit below the first import, so each reads as JSDoc for an import; the import sort recreates the class

stevensacks:
Closes #1163

## Problem

A module docblock placed below a file's first `import` no longer documents the module. JSDoc binds
by adjacency, so it attaches to whatever follows it, usually the next `import`, and the file's
primary explanation is silently misattributed to a dependency.

**The class recreates itself.** Imports sort by specifier, so a docblock written above what was
then the first import is stranded the moment an import that sorts higher is added. Nobody chooses
this and nothing reported it: `#1034` fixed one instance and asserted it was unique; **74 files**
carried it when this branch measured.

## Fix

A vitest structural guard plus the sweep that greens it, in that order.

- `.gaia/cli/src/module-docblock-placement.test.ts` asserts that no module docblock sits inside a
  file's leading import block, across every `.ts` file under `.gaia/cli/src`. It sits beside the
  two structural guards already at that level, `command-reachability.test.ts` and
  `lint-pin-parity.test.ts`.
- 74 files move their docblock to line 1.

**The guard is the sweep's oracle**, which is why they land together: the sweep is provably
complete exactly when the guard goes green. Landing them separately would leave the sweep with no
oracle at all and the guard trivially green.

## Decisions

`#1163` is a DECISION OWED row. Three arms, none left to emerge from the implementation.

- **The convention is docblock-first**, settled from the code. The issue frames it as open at
  "122 to 81, a majority but not an overwhelming one", and that counts the wrong thing: the
  non-conforming files are artifacts of the import sort, not positions anyone took. Counting
  deliberate choices it is 126 to roughly zero, and `#1034` had already accepted the correctness
  argument.
- **A vitest guard, not a custom ESLint rule**, settled by the maintainer. It is the repo's
  existing idiom for a structural convention over `src/**`, needs no rule authoring, and changes
  no ruleset, so no later CLI change is gated under a rule this PR invented.
- **One PR, guard and sweep together**, settled by the maintainer, for the oracle reason above.

## Scope boundary, stated because it is a floor rather than a clean bill of health

A docblock sitting immediately above the first declaration, with no blank line between, is **not**
reported. At that position a module docblock and ordinary JSDoc for that declaration are textually
identical and only a reader can tell them apart: `schemas/zod-error.ts` documents its one export
from there and is correct, `release/exclude-parser-parity.test.ts` describes the whole file from
there and arguably is not. Reporting the position would make the guard demand that every
documented first export lose its JSDoc. The boundary is pinned by its own fixture so widening it
later is a deliberate act.

## Verification

**The diff is provably zero-behavior.** Every swept file's non-blank line multiset is unchanged, so
the sweep is a pure reordering of lines; `git diff --check` is clean; and the only non-docblock
comment that moves is a file-level `eslint-disable` in `sandbox/__tests__/marker.test.ts` that
still precedes all code. No file carries a `#!` shebang or a `@vitest-environment` pragma that the
move could have displaced, and all 74 now begin with `/**`.

**Quality Gate**, per `wiki/decisions/Quality Gate.md`: `typecheck` clean, `lint` clean,
`lint:cli` clean, app suite **156 passed / 1 skipped** across 30 files, CLI suite **2048 passed /
1 skipped across 135 files**, `build` clean, Playwright **8 passed / 1 skipped**, dev smoke
**HTTP 200**.

`lint:cli` initially failed on nine real errors in the new file (`no-continue` x3, cognitive
complexity 22 against a limit of 15, a union sort, two sort rules, two prettier). Fixing them
decomposed the header scan into `blockCommentEnd` / `importEnd` / `detachesDocblock`, so the lint
failure produced a better design than the one it rejected.

**CI-gated bats roots**, all through `.gaia/scripts/bats5.sh`, counted by `not ok` rather than by a
tail: `.github/audit/tests/` 165/165, `.gaia/scripts/tests/` 1458/1458, `.gaia/tests/hooks/` 1370/1370,
`.gaia/tests/lib/` 448/448 (one skip, `flock` absent on this machine, the `mkdir` fallback is the
load-bearing path), `.gaia/tests/statusline/` 36/36, `region-marker-conformance` 6/6,
`shell-lint.sh` exit 0, distribution harness **17 PASS / 0 FAIL**. Read from the per-scenario
summary rather than a tail, so a second failure could not hide behind the first.

**Mutation proof, five mutants, all killed.** Each `cp`-backed from a `mktemp -d` and restored the
same way, never `git checkout --`, and each carries a hard assert that the mutation applied.

| Mutant | Killed by |
|---|---|
| M1 strand a real docblock behind a new first import (`release/manifest.ts`) | corpus assertion |
| M2 strand it with a `// Node builtins` banner between docblock and import | corpus assertion, reported `manifest.ts:2` |
| M3 stop `nextStatement` skipping comments | the banner fixture |
| M4 drop the blank-line arm of `detachesDocblock` | the below-the-block fixture |
| M5 break the source walk's extension filter | the floor assertion |

**M2 is the one that matters.** It proves the widening is load-bearing on real source rather than
only against a fixture string. My first attempt at M2 was a copy-paste of M1's mutation and would
have certified nothing; the diff stat is what exposed it.

**The guard was authored red and watched go green**, rather than written after the sweep: it
reported all 74 offenders before a single docblock moved.

## Audit gate: four rounds, and every finding landed on the guard rather than the sweep

`code-audit-maintainer-node` was the only member `resolve-audit-spawn.sh` named across all four
rounds, which is plausible for a diff that is entirely `.gaia/cli/src/**`. It is advisory-only, so
it repaired nothing and returned a byte-identical tree each time.

| Round | Result | Findings |
|---|---|---|
| 1 | clean pass | 1 Important, 2 Suggestions |
| 2 | clean pass | 0 Critical, 0 Important, 3 Suggestions |
| 3 | clean pass | 0 Critical, 0 Important, 2 Suggestions |
| 4 | clean pass | **zero at every severity**, empty findings sidecar |

**Not one finding in four rounds landed on the 74-file sweep.** Every one was a blind spot in the
guard's own line-based scanner, and all but two are fixed here.

Round 1's Important is the one worth reading: `importEnd` scanned for a bare `;`, so an import
carrying a trailing comment did not terminate it, the scan ran past the docblock below to the next
import's terminator, and **the guard reported the file clean on exactly the defect it exists to
catch**. One `// eslint-disable-line` would have disabled it for a whole file, silently. The member
proved it with a runnable probe rather than asserting it.

Rounds 2 and 3 fixed `import.meta` and a dynamic `import(` being read as import declarations, and
corrected two false claims in this PR's own prose: a `STATEMENT_END` docblock that implied coverage
it does not have, and a residual note citing the wrong issue number.

**A stopping rule was pre-committed before round 2 and capped at round 4**, on the ground that the
direction of the repairs was the real finding: the same move (narrow a lookahead, widen a
terminator) kept being correct, which is a detector at a local optimum rather than one unravelling.
Round 4 came back clean, and the member independently endorsed the cap with a sharper argument than
the one it was written on: the guard fully covers **its own threat model**, the import sort, and the
two accepted residuals need shapes the sort does not produce and `eslint --fix` does not survive. A
guard that fully covers its generating mechanism and partially covers shapes outside it has a
documented boundary, not a hole.

## What review found in this PR's own work

The `simplify` pass found a hole in the guard I had just written. `detachesDocblock` peeked exactly
one line past a docblock's close, while the header scan beside it treats `//` notes and further
block comments as transparent. So a docblock followed by an import-group banner evaded the guard
entirely: the detection check's notion of "what follows" was narrower than the parser's own notion
of "header", and nothing said so. It now binds to the next **statement** through a shared
`nextStatement` skip.

**The corpus gained no new offender**, so the sweep stands. The hole was reachable by ordinary
editing rather than already live, which is the honest reading of it.

A second self-inflicted risk was closed in the same pass: a scan that reaches nothing reports
nothing, so a broken walk or extension filter would have read as a clean corpus. A floor assertion
now turns that into a red (M5).

## CHANGELOG

**No entry owed**, verified rather than reasoned. `.gaia/release-exclude:121` lists
`.gaia/cli/src`, and `.gaia/manifest.json` carries **0** entries under it against 35 under
`.gaia/cli/`. Nothing here reaches an adopter, so there is no behavior change and nothing at Keep a
Changelog altitude.

**No bundle rebuild owed**, also checked: `findStrandedDocblock` and `nextStatement` return 0
occurrences in both committed bundles, since a `.test.ts` is in neither entry graph, and this
branch modifies neither binary.

## Filed, not fixed here

**#1273**, `severity:suggestion` / `difficulty:hard` / `surface:maintainer`. The Quality Gate's
Playwright run surfaced a flake in `.playwright/e2e/react-perf-smoke.spec.ts:149`: the StrictMode
canary asserts `strictAvg > relaxedAvg * 1.1` on wall-clock render aggregates while its own comment
records the observed effect as ~1.2, so the whole tolerance is ~0.1 of a noisy measurement.
Measured at 2 failures in 11 full `pnpm pw` runs, against 0 in 19 on `main`; the branch/main split
is sampling noise, since this spec imports nothing under `.gaia/cli/src` and this PR's only
executable change is a vitest file Playwright never runs. **Filed rather than waived** because
`.playwright/` is not a path this PR changes, so the touched-file waive does not reach it.

## Accepted residuals

Two audit Suggestions accepted rather than fixed, and **filed as #1275** rather than only recorded:
a waive records but does not track, and DRAIN's override says a finding already observed to spread
gets filed even where a waive is licensed. Five detector blind spots across two rounds is that
observation. Both are named explicitly in `STATEMENT_END`'s docblock so the boundary is stated
rather than implied.

- **A trailing block comment left open to a later line**, and **a `;` inside a comment on a
  continuation line of a multi-line import.** Both fail silent-green. They are one finding, not
  two: closing either means stripping comment content before the terminator test, and doing that
  correctly needs string awareness as well (`import x from 'https://cdn/x.js';`), at which point
  the guard contains a tokenizer. `#1275` proposes reading the header from the TypeScript AST
  instead, which is already a direct `.gaia/cli` dependency.

One further observation the round-4 member deliberately did not file, recorded here for the same
reason: `import ('./x');` with whitespace before the paren is still read as a declaration. Its
failure direction is a false positive (loud red), no formatter here emits it, and `#1275`'s
direction closes it for free.

## Declined, recorded so a reviewer need not re-derive them

Both are `simplify` suggestions on this PR's own new file, declined deliberately.

- **Folding `blockCommentEnd` and `importEnd` into one higher-order `scanUntil`.** They share a
  loop shape and nothing else; they take different predicates and never need to agree. The
  extraction trades two six-line functions for a callback plus two wrappers, which is not simpler,
  and the two were split out of one loop in the first place only because `no-continue` and
  cognitive complexity required it.

- **Deleting the no-imports fixture as redundant branch coverage.** It is redundant on branches, as
  reported. It is kept because it pins a real input class: the guard applies only to files that
  have imports at all, and that contract is otherwise stated nowhere.

### DIFF
--- a/.gaia/cli/src/a11y-structural/check-a11y-triviality.test.ts
+++ b/.gaia/cli/src/a11y-structural/check-a11y-triviality.test.ts
@@ -1,4 +1,3 @@
-import {afterEach, describe, expect, test} from 'vitest';
 /**
  * Tests for the structural a11y-triviality floor AST helper
  * (`.gaia/scripts/a11y-structural/check-a11y-triviality.mjs`).
@@ -22,6 +21,7 @@ import {afterEach, describe, expect, test} from 'vitest';
  * bytes); a stories fixture is written to a temp file and passed via
  * `--stories <path>`.
  */
+import {afterEach, describe, expect, test} from 'vitest';
 import {execFileSync} from 'node:child_process';
 import {mkdtempSync, rmSync, writeFileSync} from 'node:fs';
 import {tmpdir} from 'node:os';
--- a/.gaia/cli/src/audit-ledger/append-worthiness.test.ts
+++ b/.gaia/cli/src/audit-ledger/append-worthiness.test.ts
@@ -1,4 +1,3 @@
-import {afterEach, beforeEach, describe, expect, test} from 'vitest';
 /**
  * Tests for the worthiness-audit ledger writer
  * (`.gaia/scripts/audit-ledger/append-worthiness.mjs`).
@@ -17,6 +16,7 @@ import {afterEach, beforeEach, describe, expect, test} from 'vitest';
  * from `node_modules`; this `.gaia/cli` workspace carries its own `typescript`
  * devDependency, so the test runner can exec it.
  */
+import {afterEach, beforeEach, describe, expect, test} from 'vitest';
 import {execFileSync} from 'node:child_process';
 import {mkdtempSync, readFileSync, rmSync, writeFileSync} from 'node:fs';
 import {tmpdir} from 'node:os';
--- a/.gaia/cli/src/automation/__tests__/audit-template-dogfood.test.ts
+++ b/.gaia/cli/src/automation/__tests__/audit-template-dogfood.test.ts
@@ -1,4 +1,3 @@
-import {describe, expect, test} from 'vitest';
 /**
  * Maintainer drift-guard: asserts that the in-tree
  * `.github/workflows/code-review-audit.yml` is byte-identical to the
@@ -9,6 +8,7 @@ import {describe, expect, test} from 'vitest';
  * must match; any divergence means either the live gate or the install
  * source was edited without updating the other.
  */
+import {describe, expect, test} from 'vitest';
 import {existsSync, readdirSync, readFileSync} from 'node:fs';
 import path from 'node:path';
 import {resolveRepoRootFromImportMeta} from '../../util/repo-root-fixture.js';
--- a/.gaia/cli/src/automation/__tests__/gaia-ci-template-refs.test.ts
+++ b/.gaia/cli/src/automation/__tests__/gaia-ci-template-refs.test.ts
@@ -1,9 +1,3 @@
-import {describe, expect, test} from 'vitest';
-import {existsSync, readdirSync, readFileSync} from 'node:fs';
-import path from 'node:path';
-import type {ToolId} from '../../schemas/automation-config.js';
-import {resolveRepoRootFromImportMeta} from '../../util/repo-root-fixture.js';
-import {workflowTemplatePath} from '../paths.js';
 /**
  * Maintainer drift-guard for the outbound references in the four
  * `gaia-ci-*` workflow templates.
@@ -49,6 +43,12 @@ import {workflowTemplatePath} from '../paths.js';
  * checkout where the templates or routers are missing, mirroring the
  * sibling guards.
  */
+import {describe, expect, test} from 'vitest';
+import {existsSync, readdirSync, readFileSync} from 'node:fs';
+import path from 'node:path';
+import type {ToolId} from '../../schemas/automation-config.js';
+import {resolveRepoRootFromImportMeta} from '../../util/repo-root-fixture.js';
+import {workflowTemplatePath} from '../paths.js';
 
 type TemplateContract = {
   readonly cli: readonly string[];
--- a/.gaia/cli/src/classifier/classify-determinism.test.ts
+++ b/.gaia/cli/src/classifier/classify-determinism.test.ts
@@ -1,4 +1,3 @@
-import {describe, expect, test} from 'vitest';
 /**
  * Tests for the determinism classifier AST helper
  * (`.gaia/scripts/classifier/classify-determinism.mjs`).
@@ -18,6 +17,7 @@ import {describe, expect, test} from 'vitest';
  * supplies the bytes). The three named real fixtures are classified from disk by
  * repo-relative path.
  */
+import {describe, expect, test} from 'vitest';
 import {execFileSync} from 'node:child_process';
 import path from 'node:path';
 import {resolveRepoRootFromImportMeta} from '../util/repo-root-fixture.js';
--- a/.gaia/cli/src/command-reachability.test.ts
+++ b/.gaia/cli/src/command-reachability.test.ts
@@ -1,4 +1,3 @@
-import {describe, expect, test} from 'vitest';
 /**
  * Maintainer guards for the CLI subcommand surface. Two of them, sharing one
  * router scan: a reachability guard (is each command invoked from anywhere?)
@@ -66,6 +65,7 @@ import {describe, expect, test} from 'vitest';
  * for those lines this guard checks only that the command is documented and
  * dispatched, never that the verbs after it are complete.
  */
+import {describe, expect, test} from 'vitest';
 import {existsSync, readdirSync, readFileSync, statSync} from 'node:fs';
 import path from 'node:path';
 import {resolveRepoRootFromImportMeta} from './util/repo-root-fixture.js';
--- a/.gaia/cli/src/harden/__tests__/marker.test.ts
+++ b/.gaia/cli/src/harden/__tests__/marker.test.ts
@@ -1,4 +1,3 @@
-import {describe, expect, test} from 'vitest';
 /**
  * Guard for the frozen provenance marker (`marker.ts`).
  *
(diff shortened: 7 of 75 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.3K6Bmecnpq

Issue #1275, opened 2026-08-09
The module-docblock guard reads its header from raw lines, so two comment shapes make it report a stranded docblock as clean


stevensacks:
**Location:** `.gaia/cli/src/module-docblock-placement.test.ts:91` (`STATEMENT_END`), and the
line-based header scan it serves (`blockCommentEnd`, `importEnd`, `nextStatement`,
`findStrandedDocblock`).

## Failure mode

The module-docblock guard decides where a file's leading import block ends by classifying raw
lines. That scan has an open-ended blind-spot surface, and both known holes fail **silent green**:
the guard reports a file clean while a stranded docblock sits in it, which is the one outcome a
guard exists to prevent.

Two demonstrated instances, both verified by driving the detector directly rather than reasoned
about:

1. **A trailing block comment left open to a later line.** `STATEMENT_END` requires the comment to
   close on the same line, so

   ```ts
   import {z} from 'zod'; /* a
      b */
   /**
    * What this module is.
    */
   import fs from 'node:fs';
   ```

   yields `null` where the answer is `3`. The scan never sees a terminator on either line, runs
   past the docblock to the next import's terminator, and steps over it.

2. **A `;` inside a comment on a continuation line of a multi-line import.** `  a, // see foo;`
   satisfies the terminator, so `} from 'x';` is then read as the first non-import statement and
   the header closes early. **Pre-existing rather than a regression**: the previous
   `endsWith(';')` form behaves identically.

**The reason this is filed as one issue about the instrument, and not as two shape fixes.** Both
close with the same repair, strip comment content before testing for the terminator, and doing
that correctly requires **string awareness as well**: `import x from 'https://cdn/x.js';` would
otherwise have its specifier eaten by a naive `//` strip. At that point the guard contains a
tokenizer, and each shape taught to it invites the next. Two audit rounds on the guard produced
**five** distinct detector blind spots, every one of them in the scanner:

| Round | Blind spot | Direction |
|---|---|---|
| pre-audit | one-line peek let a comment banner hide a stranded docblock | silent green |
| pre-audit | a broken source walk would report an empty, clean corpus | silent green |
| 1 | `endsWith(';')` missed a commented import line | silent green |
| 1 | raw-line classification: indented `/*` closed the header early | silent green |
| 1 | raw-line classification: `importantThing();` held the header open | noisy red |

Three were repaired in round 1 and two more arrived in round 2. The direction of travel is what
makes this an instrument question rather than a backlog of shapes.

[...]

PR #1276, opened 2026-08-09
debt(cli): guard and fix the unquoted diff --name-only class (#1229)

### LINKED ISSUE #1229: No deterministic check reaches a shell call site of diff --name-only, and the quoting class has now spread four times

stevensacks:
## Problem

Under git's default `core.quotePath`, `git diff --name-only` C-quotes any path carrying a
non-ASCII byte, a control character, a double quote or a backslash: it prints
`"caf\303\251.txt"`, not `café.txt`. Every consumer that then matches the output against a
path pattern silently stops matching.

The class has been found and fixed **four times in four files**, every time by a human or an
audit member reading the code and **never once by a check** (#1213, #1224, #1225, #1228).
`.gaia/scripts/check-audit-base-derivation.sh`'s assertion 4 exists to catch it, but its
candidate net is scoped to `.claude/agents/code-audit-*.md`, so it has never been able to see
a `.sh` or `.yml` call site. It prints `0` and passes whether or not any shell consumer quotes.

**Turning a check on found five more live instances, and three fail open:**

| Site | Output consumed as | Consequence |
|---|---|---|
| `.github/workflows/tests.yml:128` | ERE deciding whether the suite runs | **the entire test suite skips, required check green** |
| `.github/workflows/code-review-audit.yml:170` | ERE → `has_source` | **the audit skips, required check green** |
| `.github/workflows/code-review-audit.yml:619-620` | ERE vs `AUDIT_SELFHEAL_REFUSE_ERE` | **a self-heal touching a refused surface evades refusal** |
| `.github/workflows/code-review-audit.yml:356` | recorded changed-file list | the recorded path text is a quoted rendering |
| `.gaia/cli/health/comprehensive/gauge.sh:52` | glob → lens classification | the file drops out of the lens counts |
| `.github/workflows/code-review-audit.yml:196` | ERE vs a fixed ASCII path | benign; git never quotes that path |
| `.claude/hooks/pr-merge-audit-check.sh:485`, `:533` | allowlist / three-way classifier | fail-closed, not defects (#1213's reading, re-verified) |

Each of the three fail-open cases reports its required check green while doing none of the
work, which is **indistinguishable from a genuinely clean run**.

## Decision

Both arms of the DECISION OWED were settled by the maintainer before authoring and neither
was reopened.

**Instrument: a new standalone linter, not a widening of assertion 4.** That check's remit is
the audit-base derivation, and its `consumes` predicate keys on `resolve-audit-base.sh` /
`BASE_REF` / `BASE_SHA` / `FULL_BASE` — none of which any shell call site names, so
redefining it for shell collapses to "every call", which is a different check in the wrong
file. `.gaia/scripts/lint-diff-name-only-quoting.sh` mirrors the existing
`lint-hook-array-guard.sh` shape: a bespoke static scan, a sibling bats suite, folded into
`.gaia/tests/shell-lint.sh`, CI-gated by `Audit CI Tests`.

**Scan surface includes `.github/workflows/*.yml`.** The obvious reading of the issue is to
fold into `shell-lint.sh`, which walks `*.sh`, `*.bats` and `.husky/*` — and **three of the
four real fail-open defects are in workflow `run:` blocks**, which no `*.sh` glob sees. A
shell-only guard would print `0` while the worst instances stood, which is the failure the
issue's own Suggested fix warns against.

**No exemption mechanism.** The issue assumes one is needed because
`pr-merge-audit-check.sh:485`/`:533` are not defects. They take `-z` anyway: it is correct,
it is defensive, and it costs nothing. That removes the blocker rather than building the
mechanism the issue itself warns "quietly becomes a place to put new violations". **Zero
exemptions exist in this change.**

## Fix

- **The guard**, authored red and watched go green. It flags an *executed*
  `git diff --name-only` that does not carry `-z` immediately after the call. Three
  discriminations, each earning its place on a real line in this repository rather than on
  symmetry: an invocation must be preceded by `git` (else
  `check-audit-base-derivation.sh:278`, where the call is a *variable's value*, is a hit); a
  call inside a markdown code span is prose (else `code-review-audit.yml:486`, an agent
  prompt, is a hit); and `-z` must be **anchored**, since unanchored a pathspec carrying the
  token vouches for a call that still quotes — the same discrimination assertion 4 makes.
- **The sweep**, nine call sites. **Each site keeps its existing failure posture**: a
  `set -o pipefail` subshell where a git failure was detected before, a plain pipe where it
  was already tolerated. This is load-bearing rather than tidy — `$(...)` discards NUL bytes,
  so `tr '\0' '\n'` must live *inside* the substitution, and without the pipefail subshell the
  `set -eu` workflow sites would stop aborting on a git error and fail open in a *new* way.
- **The wiring**: folded into `shell-lint.sh`; withheld from the adopter bundle on exactly
  the reasoning `shell-lint.sh` gives for itself; and `.github/workflows/**` added to the
  gate's paths filter, without which a workflow-only PR resolves `shell=false`, skips the
  gate and greens the check having linted the files it just changed.

## Verification

- **The guard was authored red.** The suite ran **17/17 failing** before the linter existed,
  then 16/17 with only the real-tree regression gate red, then **17/17** after the sweep.
- **Mutation proofs against the real scan surface**, not just fixtures: dropping `-z` from
  `.gaia/scripts/resolve-audit-members.sh`, `.claude/hooks/lib/audit-dispositions.sh`,
  `.claude/hooks/distribution-preflight-check.sh` and
  `.github/workflows/distribution-audit-pr.yml` each reds the guard **naming that file**, and
  each restores clean. Every mutation asserts it actually applied before the guard runs.
- **A check on the check.** Neutering the guard to `exit 0` reds **9** of the suite's tests,
  so the suite is a real oracle rather than one passing by corpus accident.
- **The premise is empirical, not reasoned.** In a throwaway repo `git diff --name-only`
  prints `"caf\303\251.txt"` and `grep -cE '^caf'` returns `0`; with `-z | tr` it prints
  `café.txt`.
- `bash .gaia/tests/shell-lint.sh` passes. `bats .github/audit/tests/` 165/165.
- Quality Gate **correctly skips**: no `.ts|tsx|js|jsx|mjs|cjs|css` and no gate-affecting
  config in the diff, verified with the page's own applicability grep.

## Notes

- **The guard flagged its own diagnostic message**, which contained the literal
  `git diff --name-only`. Fixed by rewording the message rather than exempting the line, per
  the zero-exemption decision. It is also the cheapest available proof that the "invoked"
  discrimination is live.
- **`*.bats` is deliberately outside the scan surface**, and this narrows the surface I
  described when the decision was put to the maintainer. The reason is measured, not
  stylistic: `check-audit-base-derivation.bats` carries five *intentionally unquoted*
  agent-prose fixtures that assertion 4 exists to catch, and five sibling suites run an
  unquoted `diff --name-only` under `core.quotePath=true` as the **positive control proving
  git really quotes**. A scanner reading raw lines cannot tell a fixture string from an
  executed call, so including them would demand "fixes" that delete the evidence the class
  exists. It changes nothing about which arm won — `*.bats` was identical in both offered
  arms, the discriminator was YAML coverage — and all nine live sites are reached without it.
- **`code-review-audit.yml` has three byte-identical copies**, two of them templates under
  `.gaia/cli/`, asserted by `self-heal-scope-gate.bats`. Editing the workflow alone reds that
  suite; all three are resynced.
- **This PR self-modifies `code-review-audit.yml`**, so CI's own audit skips cleanly by design
  and no `GAIA-Audit` stamp lands from CI. Expected, not a failure: the local audit handshake
  is the authority.

Closes #1229

### Round 1 — `code-audit-github-workflows`, `code-audit-maintainer-node`, `code-audit-maintainer-shell`

**All three cleared. Zero Critical, zero Important in every member.** 8 Suggestions and 1
cross-remit finding.

**Fixed in-round** (reproduced defects in logic this PR authored; rotates only the shell
member's digest, so only that member is re-dispatched):

- **A backtick command substitution was read as prose and silently skipped — fail-open.**
  The member built `` changed=`git diff --name-only "$B...HEAD"` `` and the guard exited 0.
  An odd backtick count cannot distinguish a markdown code span from legacy command
  substitution; a backtick opening in command position (after `=` or `(`) now counts as
  substitution. The class is not fully closed and the docblock now says so.
- **Any git global option other than `-c`/`-C` hid the call outright.**
  `git --no-pager diff --name-only …` produced no hit. The `invoked` prefix now admits any
  option token.
- **The blind-spot block called all three boundaries false positives that fail closed**,
  while its own third bullet was a false negative. That was a false claim about this PR's own
  guard. The block now separates the fail-closed misses from the fail-open ones and names a
  tokenizer as what would close the latter.
- `.gaia/tests/shell-lint.sh`'s header still described one folded guard, now two.
- `shell-lint.bats` gained the fold-assertion for the new pass that its sibling already has.

Four regression tests cover the two reproduced gaps, so neither can return silently. Suite is
now 21/21.

**Filed, not fixed** — both are pre-existing and neither is logic this PR authored, so under
the workflow's digest economics folding them in would buy a full re-dispatch of the workflows
and node members purely to re-earn markers they already hold:

- **#1277** — `code-review-audit.yml:636`, the self-heal refusal gate fails open when its own
  path enumeration fails: the substitution returns `sort -u`'s status, so a failed `git diff`
  yields empty at status 0 and the refusal grep matches nothing. Not fixable in-round: the
  obvious `pipefail` repair would strand the PR with no status, because the terminal status
  writers carry no `always()`.
- **#1278** — the guard's reach: executable instances of its own class it still cannot see
  (an executed self-heal instruction inside a markdown code span at `:500`; shipped skill
  prose at `sync.md:231`; the `gaia-ci-*.yml.tmpl` adopter templates, **verified to carry no
  instance today**, so filed on class membership as `#1271` was). Filed as one issue about
  the instrument, per `#1275`'s precedent.

**Declined:** `.github/actions/*/action.yml` is outside the scan surface and contains no
`diff --name-only` call; the member itself calls it "a theoretical gap, not a live one".
Folded into #1278's scope question rather than tracked separately.

### Round 2 — `code-audit-maintainer-shell` (the only member whose digest the round-1 fixes rotated)

**Cleared. Zero Critical, zero Important, 3 Suggestions.** All three fixed in-round.

- **`:66` — an empty scan set reported `clean` and exited 0.** The discovery loop reads from
  a process substitution, whose failure `set -o pipefail` cannot see, so a `git ls-files` that
  errors left the guard printing `clean` having scanned nothing. **A misresolved scan surface
  was indistinguishable from a clean tree** — the lie-green failure this whole PR exists to
  stop, in its own new guard. Now a hard error, mirroring what `.gaia/tests/shell-lint.sh`
  already does for its own discovery, and mutation-proved: neutering it reds the new test.
- **`:147` — the `(` half of the round-1 fix flags parenthetical prose** (fail-closed,
  unreachable in this tree). The finding that mattered is that the round-1 blind-spot block
  claimed to enumerate the false positives and omitted the one that same commit created.
- **`:106` — the residual fail-open bullet's example under-determined it.**

**The fix for the last two was not to extend the list.** One asked to tighten the character
class and the other to widen it; taken as written they pull in opposite directions, which is
this page's own tell that a round has moved from the parser into the design. Every widening
of a prose list invites the next one, so the block now states the boundary as a **closed
rule** with both error directions following from it, and two tests pin those directions. That
is `#1245`'s recorded design input applied in place: replace the enumeration with the pattern.

Two pre-existing tests were hardened in the same commit: each tracked no scannable file, so
after the new guard they would have passed via its error path rather than for their stated
reason.

**Declined:** applying the same empty-set guard to the sibling `lint-hook-array-guard.sh`.
That file is untouched by this PR and the gap there is pre-existing; widening the diff to a
second guard buys nothing this PR owes.

### Round 3 — `code-audit-maintainer-shell`

**Cleared. Zero Critical, zero Important, one Suggestion, accepted.** Suite 24/24.

## Accepted residuals

**`.gaia/scripts/lint-diff-name-only-quoting.sh:115` — the docblock's closed rule slightly
overstates the predicate the code implements.** The block says the test recognizes a backtick
opening "immediately after `=` or `(`, and NOTHING ELSE"; the awk tests `prefix ~ /[=(]`/`
over the whole prefix, so an earlier parenthesized code span anywhere before the call also
clears `inspan`. Reproduced: ``echo "See (`foo`) then run `git diff --name-only $B` please."``
is flagged.

Accepted rather than fixed, on four grounds. It is **fail-closed only** — `inspan` is only
ever cleared, never set — and the reporting member grades it "documentation drift with no
runtime failure mode and no missed defect". The **direction converges**: 5 findings, then 3,
then 1, with severity falling fail-open → fail-closed → documentation, so each repair left
the artifact more closed rather than larger. It is the **third consecutive finding on the
same paragraph**, and the stopping rule pre-committed before round 2 was already overridden
once to fix a kind-2 comment defect; overriding a second time is how a guard round becomes a
prose treadmill. And under this page's digest economics, folding a docblock correction into
an otherwise-clean PR buys a full member re-dispatch solely to re-earn a marker, which a
docblock overstatement in a maintainer-only, release-excluded linter does not justify.

## Corrections to this body

- **Verification above says the suite went 17/17.** That was true when written; after rounds
  1 and 2 it is **24/24**, with seven tests added across the two rounds.
- **The Fix section describes the in-span discrimination as "a call inside a markdown code
  span is prose".** Round 1 qualified that: a backtick opening in command position (after `=`
  or `(`) counts as a command substitution, not prose. The accepted residual above states the
  remaining gap between that rule and its implementation.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.Qii8Fwom6n

PR #1280, opened 2026-08-09
debt(tests): give the concurrency meter one hook-delivery idiom (#1232)

### LINKED ISSUE #1232: The concurrency suite hand-rolls the hook-invocation idiom 14 times and cannot reach the shared helper

stevensacks:
The INV-7 concurrency meter hand-rolled the hook-invocation idiom at every site that drives a hook, with no shared primitive to reach for. The next scenario added to it copies a neighbour or invents a variant, which is exactly how the sibling `.gaia/tests/hooks/` directory acquired the drift `#1219` documents: 29 hand-copies, two incompatible deny contracts, three different matchers.

Closes #1232

## The inventory was understated, and it is corrected on the issue

The issue reports 14 delivery sites, all of them "already the quote-safe positional form". Measured with `grep -n "printf.*|.*bash" .gaia/tests/concurrency/concurrency.bats`, there are **19**, and two of them are not quote-safe:

| Shape | Sites | Count |
|---|---|---|
| bare `run` | 107 | 1 |
| `run run_in DIR --` | 141, 151, 160, 170, 1293, 1301, 1309, 1315 | 8 |
| `run env HOME=` | 920, 929 | 2 |
| `$(run_in DIR -- …)`, no bats `run` at all | 963, 968, 979 | 3 |
| plain subshell, output discarded, one with a `PATH=` prefix | 1204, 1259, 1383 | 3 |
| **interpolated, not quote-safe** | 675, 687 | 2 |

Sites 675 and 687 read `run bash -c "cd '$tree' && … printf '%s' '$payload' | bash '$tree/…'"`, splicing the payload into a quoted string. That is the shape `run-hook.sh`'s "WHY THE PAYLOAD IS POSITIONAL" block warns against. It is **latent rather than live**: both payloads are `jq -nc` output over fixed strings and mktemp paths, so no single quote reaches them today. The issue's framing survives, but the file already contained two instances of the shape rather than zero. Recorded as a comment on `#1232` so the record carries the contradiction rather than only the conclusion.

## The decision this issue owed

The issue offered three options. This takes the third, and the code settles it rather than taste:

1. `invoke_hook` in `.gaia/tests/hooks/helpers/run-hook.sh` calls bats' `run` itself, so its contract fits **1 of the 19** sites here. The other 18 need `run_in`, an environment prefix, a `$( )` capture, or no capture at all: four capture modes.
2. The suite already has a primitive home, `lib/concurrency-harness.sh`, sourced from `setup()` and already owning `run_in`. Delivery belongs beside the runner it has to compose with.
3. Option 1 as written ("have both directories use it") costs either rewriting `run-hook.sh` against 33 suites and 98 `invoke_hook` / `invoke_hook_in` call sites, or sourcing the hooks suite's private helper cross-directory. This repo's convention for a cross-directory bats helper is `.gaia/tests/helpers/` (`files.sh`, sourced by 9 suites across 3 directories via `$REPO_ROOT`), not reaching into `hooks/helpers/`. The genuinely shared surface is one line of bash and pays for neither.
4. Option 2 is the one the issue flags as most likely to grow the artifact for no behavioural gain, and the four capture modes are why.
5. "Change nothing" rested on the quote-safety claim above, which is false.

Both files now name the other, so two implementations of one idiom is a recorded decision rather than drift.

## What landed

- `gaia_deliver_hook PAYLOAD HOOK` in `lib/concurrency-harness.sh`: the invocation only, no `run` and no `cd`, so the caller supplies the runner. Carries the positional-payload rationale.
- `run_with VAR=VALUE... -- CMD...`: the environment analogue of the existing `run_in`, same subshell rationale, `--` required rather than optional. It exists because `env` is an external binary and cannot invoke a shell function, and three sites carry an `env HOME=` / `PATH=` prefix. Not speculative: those three sites are its only justification.
- All 19 sites routed through them, including the two interpolated ones.
- The README `## Files` entry and the harness header updated to name the primitive.

The form was chosen against the linter rather than by preference. At the `.sh` `style` floor the function pair lints clean with **zero** suppressions, while a string constant needs three (SC2089/SC2090/SC2016) and an array two (SC2034/SC2016).

Fifteen other `bash -c` sites in the same file source a lib or run a command rather than delivering a payload to a hook. Different idiom, deliberately untouched.

## Verification

- **The meter, which is the oracle this issue names**: `bash .gaia/tests/concurrency/meter-gate.sh` reads **23 / 23**, exit 0, matching `expected-status.txt` in both directions. Baselined at 23/23 **before** the first edit, so the number is a comparison rather than an assertion. No `## Published assertion changes` entry is owed: delivery mechanics changed, no scenario's assertion moved.
- **Vacuity proof, because a green suite is not the property at risk here.** The risk of this change is a scenario that still passes while delivering nothing, so `gaia_deliver_hook` was mutated to deliver an **empty** payload and the meter re-run. It read **18 / 23**, exit 1, with five scenarios red: `C3-02`, `C3-03`, `C4-07`, `C5-02`, `C7-01`. Those five are proved load-bearing on delivery.

  The other four converted scenarios (`C5-01`, `C6-02`, `C6-03`, `C7-02`) stayed green under that mutation, so they were checked separately rather than assumed: their hooks resolve the acting tree from **cwd**, which `run_in` supplies, so an empty payload does not change what they assert. This is a pre-existing property of those scenarios, not something this change introduced, and it is the reason the mutation alone could not speak for them. Delivery at those five call sites was proved directly instead, by instrumenting the primitive to record what it is handed and re-running exactly those scenarios: five payloads recorded for five call sites, every one non-empty and well-formed (three `EnterWorktree` payloads of ~300 bytes, two statusline `workspace.current_dir` payloads). Combined with a standalone probe showing the primitive delivers a payload carrying `'`, `"`, `$( )` and backticks byte-exact, that closes the path from call site to hook stdin.

  Both the mutation and the instrumentation were restored with `cp` rather than `git checkout --`, and the restore was verified with `git diff --quiet` returning byte-identity against the commit, then the meter re-run clean.
- `bash .gaia/tests/shell-lint.sh`: passed, 197 `.sh`, 194 `.bats`, husky, and both repo-authored guards clean.
- The five bats roots CI runs (`.github/audit/tests/`, `.gaia/scripts/tests/`, `.gaia/tests/hooks/`, `.gaia/tests/lib/`, `.gaia/tests/statusline/`), run locally through `bats5.sh` so local matches CI's bash 5. **All five pass, zero failures**, and CI runs the same set on this push.
- Quality Gate skipped, correctly: its documented grep over the staged set returns empty, no `.ts|tsx|js|jsx|mjs|cjs|css` and no gate-affecting config.

## CHANGELOG

No entry. Every file here is under `.gaia/tests/`, which is release-excluded (`.gaia/release-exclude:104`) and carries zero manifest entries, and no gate's behaviour changes. This is the CHANGELOG gate's "test-only changes that alter no shipped behavior" case.

## Audit round 1, and what it changed

`code-audit-maintainer-shell` (the whole resolved member set) **refused**, on one Important finding in `run_with`, the helper this PR introduces. It is worth stating plainly because it is the same defect class the PR exists to remove, reintroduced by the fix for it:

**`run_with` documented `--` as required and did not enforce it, and the unenforced path failed open at status 0.** Under bats' `run` errexit is off, so dropping the separator made the loop eat the command as environment assignments; the hook never ran, `status` was 0, and `output` held only export noise. A scenario asserting `[ "$status" -eq 0 ]` plus an absence check then passes against a hook that never ran. `meter-gate.sh` cannot see this, because it reports green. The member reproduced it under bats 1.13.0 on both bash 3.2.57 and 5.3.15.

Beside it, `${1?}` was inert: it fires only on an *unset* parameter, which the loop condition already precludes, so an empty or malformed assignment reached `export`, failed non-fatally under `run`, and the command then ran **without** its override.

Both are fixed. Misuse now exits 64, and the separator is validated by an exact scan **before** anything is exported. That ordering was itself a second iteration: checking inside the export loop made the diagnostic report whichever later word first failed to parse, and exported the valid-identifier words on the way to refusing. My own new test caught that, not review.

**A regression suite now pins both primitives**, at `.gaia/scripts/tests/concurrency-harness-helpers.bats`. It cannot live in the concurrency meter, whose scenario list and `target` denominator are frozen and move only by the maintainer's word, so it sits beside `bats-files-helper.bats`, which pins the other cross-suite bats primitive for the same reason. Eleven tests: delivery byte-exactness (including a payload carrying `'`, `"`, `$( )` and backticks), status propagation, `run_with`'s happy path, non-leakage, `run_in`+`run_with` nesting, and four misuse arms that must fail loudly.

Every new assertion is proved by mutation rather than asserted to work: dropping the separator check reds test 8 alone, dropping the after-separator check reds test 9 alone, and reverting the assignment guard to `${1?}` reds tests 10 and 11 alone. Verified on bash 3.2.57 and 5.3.15.

The member also confirmed all four points I asked it to check independently, found no cross-remit and no out-of-scope findings, and noted that the conversion **tightened** C4-07: the old `cd '$tree' && … ; printf …` still ran the delivery from the wrong directory when the `cd` failed, where `run_in`'s `( cd "$dir" && "$@" )` does not.

Re-verified after the fix: meter **23 / 23** exit 0, the full `.gaia/scripts/tests/` root exit 0 with zero failures, `shell-lint.sh` passing over 195 tracked suites (the new file included).

## Audit rounds 3 and 4

**Round 3** cleared on `15f1d8ea` with two Suggestions, both about the new suite asserting less than the helper documents: the five misuse arms checked only a non-zero status while the header promises 64 (mutating every `exit 64` to `exit 1` left all twelve green), and the empty-assignment test named no refusal message, so deleting the assignment check reddened only the bare-name test. Both were applied rather than deferred, because the alternative was not free: under this drain's residue trigger a filing here becomes a whole extra row, which costs more than the one round folding them in cost. Each is now mutation-proved, and the member independently re-derived all nine mutants against a `git archive` extraction rather than trusting the report.

**Round 4 cleared with the marker written**, one Suggestion, recorded below.

Rounds 1 and 2 are described above. The arc is worth stating plainly, since it is this PR's own subject turned on itself: round 1 found the new helper failing open at status 0, round 2 found the test written to catch that could not fail, and rounds 3 and 4 found the assertions weaker than the prose around them. Every one is the same class the PR exists to remove.

## Accepted residuals

Recorded rather than fixed, under the digest economics in `wiki/concepts/PR Merge Workflow.md`: the PR is clean and the member is marked, so folding a comment reword in buys a full re-dispatch to re-earn a marker already earned. Round 4's finding also sits on prose an earlier round of mine wrote, which is the enrichment loop that page names, so the pre-committed stopping rule takes the accept-and-note branch.

- `.gaia/scripts/tests/concurrency-harness-helpers.bats:114`, the block comment above the misuse arms reads "Each checks the exact status AND that the command never ran". The first clause is true for all five arms; the second is true for four. Test 9 (`no command after --`) carries no absence check, correctly, because nothing follows the separator so no command could have run. A maintainer auditing the suite against its own comment could read coverage into test 9 that is not there. The omission in the test is right and should stay; only the comment overreaches. Suggestion tier, no live failure mode.

### DIFF
--- a/.gaia/scripts/tests/concurrency-harness-helpers.bats
+++ b/.gaia/scripts/tests/concurrency-harness-helpers.bats
@@ -20,6 +20,14 @@
 # Assertion style (`.claude/rules/bats-assertions.md`): POSIX `[ ]` and explicit
 # `return 1`, never a bare `[[ ]]`, so these fail correctly under macOS's bash
 # 3.2 as well as CI's bash 5.
+#
+# The misuse arms pin the exact status 64 rather than merely a non-zero one, and
+# each names the refusal it expects. A `-ne 0` check greens on 127 or 1 as
+# readily as on 64, so the documented status would drift unnoticed; and a test
+# that names no message can be satisfied by a different arm firing, which is how
+# a guard disappears with its own test still green. Test 11 is the exception: its
+# message is bash's own `not a valid identifier` wording, which is not safe to
+# pin across bash versions.
 
 setup() {
   THIS_DIR="$( cd "$( dirname "$BATS_TEST_FILENAME" )" && pwd )"
@@ -103,12 +111,12 @@ SH
 # --- run_with, the misuse arms. Each must fail LOUDLY, never green ----------
 #
 # These are the assertions that would have caught the silent-green defect. Each
-# checks a non-zero status AND that the command never ran, because a status
-# check alone would pass on any failure including the wrong one.
+# checks the exact status AND that the command never ran, because a status check
+# alone would pass on any failure including the wrong one.
 
 @test "run_with refuses a missing -- separator rather than eating the command" {
   run run_with STUBVAR=x gaia_deliver_hook '{"c":3}' "$HOOK"
-  [ "$status" -ne 0 ]
+  [ "$status" -eq 64 ]
   grep -qF -- 'missing -- separator' <<<"$output"
   # The hook must not have run: its output would carry PAYLOAD=.
   grep -qF -- 'PAYLOAD=' <<<"$output" && return 1
@@ -117,20 +125,24 @@ SH
 
 @test "run_with refuses when nothing follows the separator" {
   run run_with STUBVAR=x --
-  [ "$status" -ne 0 ]
+  [ "$status" -eq 64 ]
   grep -qF -- 'no command after --' <<<"$output"
 }
 
 @test "run_with refuses an empty assignment rather than running without it" {
   run run_with "" -- gaia_deliver_hook 'x' "$HOOK"
-  [ "$status" -ne 0 ]
+  [ "$status" -eq 64 ]
+  # Name the arm. Without this the `case` arm could be deleted and this test
+  # would stay green, because `export ""` fails on its own and the `|| exit 64`
+  # beside it would catch that instead.
+  grep -qF -- 'not an assignment' <<<"$output"
   grep -qF -- 'PAYLOAD=' <<<"$output" && return 1
   return 0
 }
 
 @test "run_with refuses a malformed assignment rather than running without it" {
   run run_with 'FOO-BAR=1' -- gaia_deliver_hook 'x' "$HOOK"
-  [ "$status" -ne 0 ]
+  [ "$status" -eq 64 ]
   grep -qF -- 'PAYLOAD=' <<<"$output" && return 1
   return 0
 }
@@ -140,7 +152,7 @@ SH
   # and applies no value, so without an explicit check the command would run
   # with its override silently absent, which is this suite's whole subject.
   run run_with STUBVAR -- gaia_deliver_hook 'x' "$HOOK"
-  [ "$status" -ne 0 ]
+  [ "$status" -eq 64 ]
   grep -qF -- 'not an assignment' <<<"$output"
   grep -qF -- 'PAYLOAD=' <<<"$output" && return 1
   return 0
--- a/.gaia/scripts/tests/concurrency-harness-helpers.bats
+++ b/.gaia/scripts/tests/concurrency-harness-helpers.bats
@@ -1,0 +1,125 @@
+#!/usr/bin/env bats
+# The delivery primitives in .gaia/tests/concurrency/lib/concurrency-harness.sh:
+# `gaia_deliver_hook` and the `run_with` runner it composes with.
+#
+# Why these are pinned here rather than in the concurrency meter itself: the
+# meter's scenario list and its `target` denominator are frozen
+# (.gaia/tests/concurrency/README.md, expected-status.txt), so a primitive's own
+# unit coverage cannot be added there without moving a number that moves only by
+# the maintainer's word. This sits beside bats-files-helper.bats, which pins the
+# other cross-suite bats primitive for the same reason.
+#
+# The failure mode being pinned is specific and it is NOT "the meter goes red".
+# The meter cannot see this class at all. Under bats' `run` errexit is off, so a
+# `run_with` that silently accepted a missing `--` would eat the command as
+# environment assignments, invoke no hook, and still report status 0 with empty
+# output; a scenario asserting `[ "$status" -eq 0 ]` plus an absence then passes
+# against a hook that never ran. That is the vacuous green the whole primitive
+# exists to prevent, so it is asserted directly here.
+#
+# Assertion style (`.claude/rules/bats-assertions.md`): POSIX `[ ]` and explicit
+# `return 1`, never a bare `[[ ]]`, so these fail correctly under macOS's bash
+# 3.2 as well as CI's bash 5.
+
+setup() {
+  THIS_DIR="$( cd "$( dirname "$BATS_TEST_FILENAME" )" && pwd )"
+  REPO_ROOT="$( cd "$THIS_DIR/../../.." && pwd )"
+  . "$REPO_ROOT/.gaia/tests/concurrency/lib/concurrency-harness.sh"
+
+  # A stand-in hook: echoes what it was handed on stdin, plus one environment
+  # variable, so a test can tell "delivered" from "never ran".
+  HOOK="$BATS_TEST_TMPDIR/hook.sh"
+  cat > "$HOOK" <<'SH'
+printf 'PAYLOAD=[%s] STUB=[%s]\n' "$(cat)" "${STUBVAR:-unset}"
+SH
+}
+
+# --- gaia_deliver_hook -----------------------------------------------------
+
+@test "gaia_deliver_hook delivers the payload on stdin byte-exact" {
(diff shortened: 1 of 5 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.XMRIR5Gu16

PR #1281, opened 2026-08-09
debt(cli): suppress git auto-maintenance for the whole CLI test run (#1235)

### LINKED ISSUE #1235: The git-sandbox maintenance suppression #1216 added lives in one suite, while seventeen others build the same unsuppressed sandbox

stevensacks:
Closes #1235

Nineteen files under `.gaia/cli/src` build a git sandbox and commit into it, and only `wiki/commit-classify.test.ts` gated git's background auto-maintenance. A detached maintenance run repacks the object store while the next `git commit` is still running, which is what corrupted HEAD in #1216, or outlives the test into the `afterEach` that deletes the repository out from under it.

## The design question, settled before authoring

The issue asked for one fork to be settled first and the code does not settle it: a shared fixture module adopted opt-in across the population, or a run-wide environment. **Settled by the maintainer: the run-wide environment.** A `setupFiles` entry in `.gaia/cli/vitest.config.ts` puts the four maintenance keys in `GIT_CONFIG_COUNT` / `GIT_CONFIG_KEY_n` / `GIT_CONFIG_VALUE_n`, which reaches every git subprocess in the run regardless of `cwd`, so a sandbox added later inherits the suppression without opting in.

The measurement that drove it: **the issue's own enumeration of which files need the fix came out short**, missing `wiki/sync-await.test.ts`. The opt-in arm's stated risk ("a suite added later can still miss it") was already realised at filing time, by the author who had just read the code. Under the env arm that hand-written list stops being load-bearing. Both corrections are recorded on the issue rather than only here.

The suppression is deliberately **not** in `execGaiaGit` (`util/git-env.ts`), which is production code the shipped CLI runs against adopters' real repositories.

## The consequence the issue flagged, and the shape of the fix

An environment entry outranks repo-local config, the same way `git -c` does. `commit-classify.test.ts`'s control repository deliberately leaves the gates off to prove git still spawns maintenance, so under the env it would inherit the suppression and report a fixture behaving perfectly. It now clears `GIT_CONFIG_COUNT` for that whole test, so what the test measures is that file's own config list and nothing else. Whole-test scope rather than per-call is deliberate: a narrower opt-out would leave the env active for the sandbox arm and make its assertion vacuous.

The gate vocabulary both guards need (`MAINTENANCE_GATE_DEFAULTS`, `MAINTENANCE_GATES`, `autoMaintenanceSpawns`, `detachedSpawns`) moves to `util/git-maintenance.ts` so the two cannot drift on what counts as ungated or as a spawn.

## Verification

TDD: red observed before the setup file existed (`expected [ Array(1) ] to deeply equal []`), green after. The predicted break in `commit-classify` was reproduced (`expected 0 to be greater than 0`) before being fixed, not assumed.

**Seven mutants, each restored with `cp` and verified byte-identical**, never `git checkout -- <path>`:

| mutant | result |
|---|---|
| setup file's `GIT_CONFIG_COUNT` set to `'0'` | RED, subject arm |
| setup file's two gate values set to git's defaults | RED, subject arm |
| control opt-out removed from `git-maintenance-env.test.ts` | RED, control arm |
| control opt-out removed from `commit-classify.test.ts` | RED, maintenance test |
| shared `maintenance.auto` default flipped to `false` | RED, **both** consumers |
| control repo set to detach (`maintenance.autoDetach=true`) | RED, foreground assertion |
| shared `gc.auto` default flipped to `0` | **GREEN, survived in both consumers** |

**The surviving mutant is recorded rather than papered over**, and it is a version-specific coverage limit rather than a defect. Measured on git 2.55 against `GIT_TRACE2_EVENT`: `gc.auto=0` suppresses on its own when `maintenance.auto` is unset (0 spawns), but an **explicit** `maintenance.auto=true` spawns a run alongside it (1 spawn). The control writes both explicitly, so `maintenance.auto` is the entry that decides and mutating its sibling cannot be observed on this git. `gc.auto` is not dead weight, it is the only spelling on git predating the maintenance task set, so it is what leaves a control ungated there. The docblock says so, and names both the mutant that survives and the one that reds.

**Quality Gate**, run from `wiki/decisions/Quality Gate.md` rather than memory: typecheck (root and CLI), `lint` + `lint:cli`, app vitest 30 files / 156 passed, **CLI vitest 136 files / 2050 passed / 1 skipped**, `pw` 8 passed, dev smoke 200, `build` and `.gaia/cli bundle`. The CLI suite is the one that matters here, since it runs entirely under the new setup file.

**CI-gated suites, all green**: `.github/audit/tests/` 165, `.gaia/scripts/tests/` 1495, `.gaia/tests/hooks/` 1370, `.gaia/tests/lib/` 448, `.gaia/tests/statusline/` 36, `shell-lint.sh`, `distribution/run-all.sh` 17/17, and the rebuild-from-src check. Run through the bash-5 guard so local matches CI.

## Scope

**CHANGELOG: no entry.** `.gaia/cli/src` and `.gaia/cli/vitest.config.ts` are both in `.gaia/release-exclude` (lines 129 and 135) with zero manifest entries, no shipped behaviour changes, and `pnpm bundle` reproduces `gaia`, `gaia-maintainer` and `templates/` byte-identical to what is committed. Test-only.

**Bound stated rather than hidden:** this covers the `.gaia/cli` vitest project. Git sandboxes built by the bats suites under `.gaia/tests/` are outside it, and outside #1235, which is scoped to `.gaia/cli/src`.

### DIFF
--- a/.gaia/cli/src/util/git-maintenance-env.test.ts
+++ b/.gaia/cli/src/util/git-maintenance-env.test.ts
@@ -9,9 +9,13 @@
  *
  * Two arms, because a bare "no maintenance was spawned" assertion passes just
  * as well when the trace recorded nothing at all. The control arm clears
- * `GIT_CONFIG_COUNT` for one commit into the same repository, which is the only
- * variable that changes between the arms, so a spawn there proves both that the
- * repository is genuinely ungated and that the trace can see a spawn.
+ * `GIT_CONFIG_COUNT` and commits into an identically configured repository, so
+ * a spawn there proves both that such a repository is genuinely ungated and
+ * that the trace can see a spawn. What holds every other variable fixed is the
+ * shared `beforeEach`, not a shared repository: each arm gets its own
+ * `mkdtempSync` repo, built from the same list and deleted in `afterEach`. They
+ * are deliberately not hoisted into a `beforeAll`, which would let the control
+ * arm's own maintenance run touch the object store the subject arm measures.
  */
 import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
 import {mkdtempSync, rmSync} from 'node:fs';
--- a/.gaia/cli/src/util/git-maintenance-env.test.ts
+++ b/.gaia/cli/src/util/git-maintenance-env.test.ts
@@ -1,0 +1,90 @@
+/**
+ * The run-wide git auto-maintenance suppression, proved against git itself.
+ *
+ * `git-maintenance-env.ts` is a vitest `setupFiles` entry, so nothing imports
+ * it and no assertion about its exports would show that it reached a git
+ * subprocess. This asserts the effect instead: a repository carrying git's own
+ * defaults for both gates, committed into from inside this run, spawns no
+ * maintenance.
+ *
+ * Two arms, because a bare "no maintenance was spawned" assertion passes just
+ * as well when the trace recorded nothing at all. The control arm clears
+ * `GIT_CONFIG_COUNT` for one commit into the same repository, which is the only
+ * variable that changes between the arms, so a spawn there proves both that the
+ * repository is genuinely ungated and that the trace can see a spawn.
+ */
+import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
+import {mkdtempSync, rmSync} from 'node:fs';
+import {tmpdir} from 'node:os';
+import path from 'node:path';
+import {execGaiaGit} from './git-env.js';
+import {
+  autoMaintenanceSpawns,
+  detachedSpawns,
+  MAINTENANCE_GATE_DEFAULTS,
+} from './git-maintenance.js';
+
+/**
+ * Enough config to commit, with the gates left at git's own defaults.
+ *
+ * `MAINTENANCE_GATE_DEFAULTS` carries the gates, so what makes this repository
+ * ungated is stated once, in the module both maintenance guards read it from.
+ *
+ * `gc.autoDetach` and `maintenance.autoDetach` are not gates and are set here,
+ * which keeps the control arm's own maintenance run a foreground child so it
+ * cannot outlive the test into the `rmSync` in `afterEach`. Both spellings,
+ * because modern git resolves `maintenance.autoDetach` ahead of `gc.autoDetach`.
+ */
+const UNGATED_CONFIG: [string, string][] = [
+  ['user.email', 'test@example.com'],
+  ['user.name', 'Test'],
+  ['commit.gpgsign', 'false'],
+  ['gc.autoDetach', 'false'],
+  ['maintenance.autoDetach', 'false'],
+  ...MAINTENANCE_GATE_DEFAULTS,
+];
+
+describe('run-wide git maintenance suppression', () => {
+  let root: string;
+  let traceRoot: string;
+
+  beforeEach(() => {
+    root = mkdtempSync(path.join(tmpdir(), 'gaia-git-maint-'));
+    traceRoot = mkdtempSync(path.join(tmpdir(), 'gaia-git-maint-trace-'));
+    execGaiaGit(['init', '-q', '-b', 'main'], root);
+
+    for (const [key, value] of UNGATED_CONFIG) {
+      execGaiaGit(['config', key, value], root);
+    }
+  });
+
+  afterEach(() => {
+    vi.unstubAllEnvs();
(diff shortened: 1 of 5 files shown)


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.bAfL8hH2MN

PR #1282, opened 2026-08-09
debt(audit): check the roster for paths that dispatch no member (#1245)

### LINKED ISSUE #1245: Nothing checks the Code Audit Team roster for paths that dispatch no member, so an ownerless path merges unreviewed and is found only by hand

stevensacks:
Closes #1245

## What this is

The Code Audit Team roster maps globs to auditor members, and the merge gate requires a per-member clearance for every member a diff dispatches. Nothing ever asked the inverse: **which paths dispatch nobody.** Such a path resolves an empty member set, falls to the merge gate's legacy branch, is out-of-scope-allowlisted, and merges unreviewed — silently, because an empty set is equally the correct answer for a genuinely out-of-remit diff.

`verify-audit-roster.sh` gains invariant 7: every tracked file resolves an owner, or is declared in a new `unowned:` list beside `auditors:`.

**Measured before the fix: 511 of 1523 tracked files dispatched no member.**

## The DECISION OWED, settled by the maintainer

The issue is design-first on purpose and named four questions. Two were settled by the code (the universe is `git ls-files`; the host is the existing roster checker, which already has the output contract and the CI job). The maintainer chose the other two on one question, from measurements taken first:

- **Predicate: fail closed over every tracked file, with an adopter-tunable exemption list.** Rejected: a code/executable extension predicate, because that list is itself a hand enumeration and so reproduces the failure mode the issue exists to break, and it would leave `.gaia/manifest.json`, `.gaia/release-exclude` and `.github/CODEOWNERS` ownerless. Also rejected: report-only, because nothing then separates a reviewed decision from an unread list.
- **Pattern arm: apply the one known instance, do not build a detector for the class.**

### Question 4 is the one that decides whether this is worth building

The issue says so outright: *"a completeness check with a freely-appendable exemption list reproduces the current situation with more machinery."* Three answers, and none of them is documentation:

1. **Appending is reviewed.** The list lives in `.gaia/audit-ci.yml`, a path the roster already grants to a member, so adding an exemption dispatches an auditor. Adding an ownerless file dispatches nobody. That asymmetry is the whole point.
2. **A blanket entry cannot pass.** An entry reaching a path some member already owns fails as **overbroad** — which `**`, the one move that would neuter this in a single line, does by construction.
3. **The list cannot rot into scenery.** An entry no ownerless path needs fails as **redundant**; one matching nothing fails as **dead** (maintainer repo only, since an adopter clone legitimately lacks paths the shipped list names).

## The triage

Today's 511 were triaged, not blanket-exempted — shipping the check by exempting every existing gap is exactly the outcome question 4 warns about.

**Granted owners** (executed, or governs what ships or how the repo is built): `.playwright/**` and the root tooling that decides what runs → default member; `.gaia/scripts/**/*.mjs` → node member; `.gaia/*.yml`, `.gaia/*.json`, `.gaia/release-exclude`, `.claude/settings.json`, `.github/CODEOWNERS` → shell member. **Every row moves from ownerless to a member; none moves between members**, which the routing-parity fixture now records. The one exception is `.gaia/audit-ci.yml`, already granted explicitly, where `.gaia/*.yml` generalizes rather than adds.

**Declared unowned:** prose, assets, fixtures, generated bundles and test data.

## The pattern arm was falsified, and the correction is the more useful result

`.gaia/cli/*.config.*` **cannot be granted at all.** `*` matches any run within a segment, so it also spells `.gaia/cli/x.config.sh`, which the shell member claims through `.gaia/**/*.sh`, and claimant disjointness is decided **as glob languages** rather than over files that exist. The checker rejected it with that synthesized witness; the dialect has no way to subtract an extension.

The family is therefore enumerated across the five JS/TS extensions — and **the coverage invariant is what actually retires the hand-enumeration defect the issue documents.** The check reports ownerless *paths*, never missing *globs*, so it never recommends an enumeration: a sixth config extension added without a roster entry is now reported ownerless instead of merging unreviewed. Recorded on the issue by comment and inline in the roster.

## A finding against this PR's own claim, fixed rather than filed

The invariant claims to catch a newly added ownerless path. `audit-roster.yml` gated the job behind a paths filter over roster inputs, so **a PR adding one file anywhere else would have skipped the check entirely** — leaving the headline claim false in CI. The filter is gone (the checker is read-only and takes ~0.8s), and the now-unused `pull-requests: read` scope went with it.

## A round-1 fix leaked a maintainer path, and CI caught it

Applying the audit's round-1 findings made the coverage-invariant skip audible on stderr, and the comment explaining why cited `.gaia/tests/distribution/16-audit-remit-parity.sh` by path. That script **ships** (`manifest: owned`), so the citation is a maintainer-path reference on an adopter clone: 15 of 17 distribution scenarios went red at `build-staging` on a single `[maintainer-paths]` leak. The citation now sits inside a maintainer-only marker pair, so the scrub removes it for adopters and keeps it for maintainers.

The pair is at column 0 deliberately. The lockstep tests model the strip with a line-anchored match while the shipped stripper (`stripMarkerBlocks`) matches as a substring, so an indented pair survives the tests' model and fails them. The tests are the stricter side, so nothing can ship through them wrongly — every marker they accept the real stripper also strips — but the divergence costs a cycle to rediscover, so the file now records the constraint at the marker itself.

## The grant hollowed out five witnesses, and one of them was a real guard

Granting the root tooling to the default member made `Dockerfile` an **owned** path. Every test using it as the *in-scope-but-ownerless* witness therefore stopped testing what its name claims. Round 1 caught one such site; four more survived it.

The one that matters: `.gaia/tests/hooks/audit-digest-lib.bats` held the digest suite's **only** ownerless-fold guard, and it went hollow. Proven by mutation in both directions — deleting the fold from `audit-digest.sh` outright left that suite **fully green** at the previous HEAD, and red at the fork point. A change whose entire premise is *"a check that silently reports green is the defect"* cannot merge having quietly disabled one.

The witness moves to `.editorconfig`: no roster glob claims it, and the out-of-scope allowlist does not reach it. Each repaired guard is proven to fail again on the mutation it watches (delete the digest fold → the digest case reds; widen the allowlist to swallow `.editorconfig` → both the digest case and the golden-table row red). Same correction applied to the golden scope table, the two zero-match legacy-branch cases, one UAT-011 case, and the four prose sites that cited `Dockerfile` as the ownerless example.

Two further repairs from the same round: the `merge-audit-ci.ts` docblock's own premise — that `unowned` belongs to no managed set — was unenforced prose, so a test now pins it (adding `unowned` to `MANAGED_WHOLE_VALUE_KEYS` fails that test and only that test); and its claim about re-proposal cadence is corrected, since with the adopter value absent and baseline equal to latest the verdict is `noop`.

The roster wiki page gains the coverage invariant it was missing and loses two sentences this branch falsified: that the workflow runs on pull requests behind a paths filter, and that the checker verifies disjointness but never coverage.

## Verification

- **TDD**: 9 cases written and confirmed red before the implementation, plus a non-vacuity control — the invariant is silent on a non-git fixture root, which is what keeps this suite's other 70 fixtures green, and that test passes identically if the invariant never fires at all.
- **Guard proofs**: 5 mutants against the real roster (drop a grant / drop an exemption / add `**` / add a subsumed entry / add an entry matching nothing) each red for the right reason; every restore `cp`-based and verified byte-identical; control clean.
- **Quality Gate green**: `/simplify` (4 agents), typecheck root + CLI, `lint` + `lint:cli`, app vitest 30 files / 156, CLI vitest 136 files / 2050, `pw` 8 passed, dev smoke HTTP 200, `build`, `.gaia/cli bundle` reproducing byte-identically.
- **CI-gated suites**, re-run green after the final round: `.github/audit/tests` 165, `.gaia/scripts/tests` 1508, `.gaia/tests/hooks` 1372, `.gaia/tests/lib` 448, `.gaia/tests/statusline` 36, `shell-lint`, `distribution/run-all.sh` 17/17, `region-marker-conformance` 6/6.

## Accepted residuals

**The exemption list is a statement about one tree and cannot serve both.** Everything the maintainer-only members own is ownerless on an adopter clone, and the release scrub can strip a marker block but never add one, so no arrangement makes one list correct in both trees. An adopter running the checker by hand sees GAIA's machinery reported ownerless — accurate there — and the roster comment tells them to declare `.gaia/**` and `.claude/**`. Nothing runs it automatically on an adopter clone, so this is a hand-run finding with a printed remedy, not an inherited red check.

**`unowned:` is deliberately unmanaged by `/update-gaia`.** The list declares which paths in *this* tree are meant to dispatch no auditor, so what GAIA ships is a seed for its own tree and the entries carry no authority over an adopter's. It is declined on ownership, not on risk: **no verdict row in the merge table deletes an adopter value** — a key present in latest and absent from baseline writes nothing, a diverged one keeps the adopter's, and the command is read-only regardless. The consequence that is not what "unmanaged" first suggests is the other one: **an adopter who scaffolded before this key existed is never seeded at all**, because `.gaia/audit-ci.yml` is manifest class `shared` and merges key by key, so an unmanaged key never arrives. Their hand run reports every prose, asset and root-markdown path as ownerless in bulk rather than reporting one missed entry, which is the intended prompt to author the block.

*(Round 1 of the audit falsified the first two sentences this entry previously carried — that GAIA seeds it and stops, and that merging whole-value would delete adopter entries. Both were proven wrong executably against `retrigger_workflows`, the existing list-valued managed key. The code docblock and this entry are corrected; the claim never reached the CHANGELOG.)*

**Corrected in place: this is FIXED, not accepted.** The entry previously recorded that `unowned:` globs bypass the bounded-dialect check invariants 3 and 6 apply to roster globs, and declined it on cost. A later round of this PR's own audit falsified the cost estimate: `pair_records` already consumes three record streams and already holds the single copy of `glob_items()`, so the gate is a fourth stream rather than the record-plumbing change the decline assumed. It now ships. A dialect-violating exemption such as `docs**` fails `undecidable-unowned-glob` and exits 1, proven by mutating the real roster and restoring it byte-identically. The reversal is this change's own thesis applied to itself: an exemption that can silently compile to a `/`-crossing regex suppresses ownerless findings, which is a silent green inside the very check being added here.

**Corrected in place: this is FIXED, not accepted.** The entry previously recorded that `audit_scope_init`'s builtin-roster fallback gives the coverage section a second opinion its own docblock rules out. The coverage section now passes the config explicitly, so the fallback cannot reach it and the section classifies against the roster whose name it prints. No wrong green was ever reachable here, since such a roster fails `default-member-count` first; the repair was to the diagnostics, which misattributed their subject.

**Nothing enforces the "reviewed here, repaired nowhere" pairing this change creates.** Seven root tooling files (`.npmrc`, `.lintstagedrc.json`, `.prettierignore`, `Dockerfile`, `.env.example`, `.nvmrc`, `.node-version`) are granted to the roster's only `push_fixes: true` member and simultaneously added to `AUDIT_SELFHEAL_REFUSE_ERE`, so that member reviews them and may repair none of them. Both halves hold today and each was proven by probing the real hook: all seven return `deny`, while an `app/**` path still passes. The gap is that the pairing is a convention rather than an invariant, so a later grant of an eighth gate-deciding root file gains self-heal write access silently. `.lintstagedrc.json` is the stake, since `.husky/pre-commit` drives it and rewriting it narrows the Quality Gate floor. The repair is a roster-derived bats invariant, a new mechanism owing its own guard proof; declined here as scope growth on a change already `difficulty:hard` at its seventh audit round, not on merit.

**The `.claude/commands/**` exemption stays while a 629-byte agent README is granted, and the asymmetry is deliberately left standing.** 2582 lines of instruction prose, including a 1175-line `setup-gaia.md`, sit in `unowned:` and dispatch no member. The roster's own docblock invites the challenge: "Adding a path here is a claim that nothing about it is worth a specialist's review. Prefer granting it to a member." It is not repaired here because the remaining `unowned:` triage has produced a fresh finding in four separate audit rounds, each at a site no earlier round had examined, and answering that one round at a time is the drip a single systematic pass exists to replace. Recorded as evidence for that pass rather than sampled once more.

**`.gaia/cli/health/**/*.md` stays exempt on the same reasoning that granted the agent lenses an owner, and the inconsistency is left standing deliberately.** The exemption covers 1064 lines of instruction prose: a runbook, a taxonomy, and four `comprehensive/lenses/*.md`. The grant beside it routes `.claude/agents/*/**` to the prose member because those files decide what that member checks, and the rationale holds verbatim for the health lenses. The split runs through a single directory, with `gauge.sh` resolving to a member and the four lens files beside it resolving to nothing. Not repaired here for the same reason as the `.claude/commands/**` entry above: this is the fifth finding the `unowned:` triage has produced across five audit rounds, and the surface wants one systematic pass rather than another sample.

**The adopter remedy names two of six machinery trees, and the shortfall is now measured rather than estimated.** The comment tells an adopter to clear the hand-run finding by adding `.gaia/**` and `.claude/**`. Applying exactly that to the scrubbed roster leaves **50 ownerless paths**, under `.specify/`, `.github/audit/`, `.github/actions/` and `.husky/` — all GAIA machinery whose owners the release strip removes for precisely the same reason as the two the comment does name. The remedy is directionally right and incomplete. It sharpens the adopter-clone residual recorded above rather than opening a separate defect, and nothing here is a red check on an adopter clone: the workflow that runs the checker is release-excluded and is not a required check.

**The self-heal scope bullet's never-touch categories do not cover five of the seven newly granted root files.** `Dockerfile`, `.npmrc`, `.nvmrc`, `.node-version` and `.env.example` are declared-but-unrepairable in the same way the bullet already describes for `test/**`, and match none of its five enumerated categories. This is prose drift, not a live hole: the deterministic hook refuses all seven regardless, proven by direct probe in two separate audit rounds. Recorded rather than repaired because it is the same grant-falsified-prose class a systematic sweep already addressed once in this change, and re-running rounds against that class's tail is the drip the sweep exists to replace.

**Two Quality Gate findings declined, both recorded at their call sites** because both are optimizations a later reader will re-propose: matching ownership inside the awk pass (0.6s of the 0.8s) would duplicate the *precedence algorithm* rather than just the compiler, giving the coverage check a second opinion about ownership; and dropping the `P` record tag would let a path containing a tab be parsed as a live exemption regex, failing toward suppressing findings.

**One unreproduced observation, recorded rather than filed.** `distribution/run-all.sh` scenario 04 failed once inside a long sequential local run (a storybook addon throwing on `localStorage` in the staged scaffold, unrelated to anything here) and passed 2/2 standalone afterward, including a full green `run-all.sh`. One occurrence under self-inflicted load with no standalone reproduction is not a filing yet; if it recurs it earns one. It has not: the scenario has since passed in two further full `run-all.sh` runs, and CI's own harness run reached it cleanly too, failing only on the maintainer-path leak above.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.FsHQAJWhPA

PR #1290, opened 2026-08-10
debt(audit): compute the merge gate's preconditions once and ship the status upserter (#1284, #1285)

### LINKED ISSUE #1284: The audit workflow's precondition chain is copy-pasted across 27 guards, so one missed site fails the merge gate open
`.github/workflows/code-review-audit.yml:135`

## Failure mode

The job's precondition chain is copy-pasted across its `if:` expressions rather than computed once. Within the single `code-review-audit` job:

| Clause | Occurrences |
|---|---|
| `steps.gate.outputs.gated == 'false'` | 27 |
| `steps.source-changes.outputs.has_source == 'true'` | 20 |
| `steps.workflow-self-mod.outputs.self_modified == 'false'` | 18 |

Adding or tightening a precondition means editing up to 27 separate `if:` expressions. Miss one and that step's guard becomes a strict subset of its siblings', so it fires on a code path the new precondition was meant to exclude.

This is not cosmetic, because the guarded steps are the ones that write the `GAIA-Audit` commit status that `.claude/hooks/pr-merge-audit-check.sh` gates merges on. A missed site can therefore post a `success` status on a path that should have stood down: the merge gate fails **open**, which is the worst direction for a gate to fail. The divergence is also invisible to CI, since each guard covers a different code path and a given PR exercises only one of them.

## Suggested fix

Add one step that computes the shared chain into a single output (for example `steps.phase.outputs.reached_audit`), then reduce each downstream guard to that output plus the one clause that actually distinguishes it:

```yaml
if: steps.phase.outputs.reached_audit == 'true' && steps.audit.outcome == 'success'
```

One edit site for the shared preconditions, and each remaining `if:` states only its own distinguishing condition. No behavior change intended: the collapsed expressions must be equivalent to the current ones path-for-path, so the change wants an adversarial fixture per terminal path (gated, no-source, self-modified, trailer-match, chore-deps, stand-down, abort, complete) proving each still fires exactly when it does today.

Note `.github/audit/tests/ci-decision-step.bats` already exists and is the natural home for those assertions.

Handler: prompt

### LINKED ISSUE #1285: The status-comment upserter is a bash script written into RUNNER_TEMP, so it escapes shell-lint and bats while failing silently
`.github/workflows/code-review-audit.yml:1558`

## Failure mode

The `Prepare status-comment upserter` step writes a ~30-line bash script into `$RUNNER_TEMP/cra-status-upsert.sh` via heredoc. It is a real script living inside YAML, which puts it outside both guards its siblings in `.github/audit/` enjoy:

- `.gaia/tests/shell-lint.sh` never sees it, so no shellcheck oracle runs against it.
- It has no bats suite, while `.github/audit/tests/` carries eleven.

The breakage is silent by construction. Every error path in the script is deliberately non-fatal: it `exit 0`s when the comment lookup fails, and warns-and-continues when the `PATCH` fails. So a quoting or `gh api` regression does not red the job; it degrades to "the status comment silently stops updating" while CI stays green. Nothing would catch it except a human noticing a stale comment on a PR.

## Suggested fix

Move the script to `.github/audit/cra-status-upsert.sh` as a committed, executable file and delete the `Prepare status-comment upserter` step outright. `actions/checkout` already places it on disk, so the eight terminal `Status - *` steps can invoke `.github/audit/cra-status-upsert.sh` directly instead of `"$RUNNER_TEMP/cra-status-upsert.sh"`.

Removes roughly 40 lines from the workflow and picks up shell-lint plus the existing bats harness for free. Worth adding a small suite alongside it covering the two fail-soft branches (lookup failure, PATCH failure), since those are exactly the paths that currently fail invisibly.

Handler: prompt

stevensacks:
Drains two `tech-debt` issues, both in `.github/workflows/code-review-audit.yml`, the machinery that decides whether the `GAIA-Audit` commit status gets written.

## #1284 — the precondition chain was copy-pasted across 27 guards

The audit-phase preconditions were restated in full on every step that depends on them: `gated == 'false'` in 27 sites, `has_source == 'true'` in 20, `self_modified == 'false'` in 18. Tightening a precondition meant editing every site, and a missed one left that step guarding on a strict subset of its siblings, so it fired on a path the new precondition was meant to exclude.

That is not cosmetic. The guarded steps write the `GAIA-Audit` status that `.claude/hooks/pr-merge-audit-check.sh` gates merges on, so the failure direction is **the merge gate failing open**. CI cannot catch it either: each guard covers a different code path and one PR exercises only one of them.

A new `Resolve audit phase` step computes the chain once into `reached_audit`. The 18 full-chain sites now read that single output plus whatever clause actually distinguishes them. `self_modified == 'false'` now appears zero times outside the phase step; `gated == 'false'` drops 27 → 9 (the partial-chain and inverted guards, which are genuinely different conditions and were left alone).

## #1285 — the status-comment upserter escaped both guards

The upserter was a ~30-line bash script heredoc'd into `$RUNNER_TEMP`, which put it outside `.gaia/tests/shell-lint.sh` (never saw it) and outside `.github/audit/tests/` (no suite). Its every error path is deliberately non-fatal, so a quoting or `gh api` regression would not red the job; it would degrade silently to "the status comment stops updating" while CI stayed green.

It now ships as `.github/audit/cra-status-upsert.sh`, picked up automatically by shell-lint's `*.sh` discovery (clean at `-S style`) and covered by its own bats suite. The `Prepare status-comment upserter` step is gone and the 16 call sites invoke the committed path.

## How the no-behavior-change claim is backed

`.github/audit/tests/ci-guard-paths.bats` translates the GitHub Actions expression subset the workflow uses into evaluable bash, then, for each of the eight terminal paths, asserts the **exact** set of steps that fire. Exact rather than superset: a widened guard fires an extra step and a narrowed one drops a step, and only an exact-set assertion catches both.

The expected sets were captured against the **pre-refactor** workflow and are byte-identical after it.

The scenarios feed the guards a `reached_audit` value obtained by **executing** the phase step's own `run:` block, extracted from the workflow YAML. A hand-written copy of that block would have left the suite blind to the one thing it exists to prove: a structural test can pin the step's inputs and its two output literals, but not the condition choosing between them, so deleting a clause from the real step would have kept every scenario green.

Every new guard was mutation-tested rather than assumed:

| Mutation | Killed by |
|---|---|
| drop `gated == 'false'` from the audit step | `adversarial: gate label missing …` |
| drop `has_source == 'true'` | `adversarial: no source changes …` |
| drop `self_modified == 'false'` | `adversarial: workflow self-modification …` + `a skipped self-mod check is not read as 'false'` |
| revert one call site to `$RUNNER_TEMP` | `upsert: the workflow invokes the committed script, not a temp copy` |
| drop an input from the phase step | `audit phase: … exactly the supplied inputs` + the subordination test |
| drop `[ "$SELF_MODIFIED" = 'false' ]` from the phase step's `run:` body | `terminal path: workflow self-modification`, both `adversarial: … self-modification` tests, and both `audit phase:` precondition tests |

The harness itself is guarded: it fails loud (status 2) on an untranslatable expression rather than reading a truncated guard as "did not fire", and it carries self-checks proving it discriminates. That bug was real during development, and silent.

## Collateral, and why

- `ci-status-member-gate.bats`: `install_upsert_stub()` now writes its fixture at the repo-relative path the steps invoke. The three `cra-status-upsert:` fail-soft tests move to the script's own suite, following the one-suite-per-`.github/audit` script convention the sibling scripts already use; their rationale comments moved with them.
- `ci-decision-step.bats`: the subordination test asserted the *duplication* as its mechanism (`grep -c … >= 2`). It now asserts the derived precondition plus the phase step's inputs, which is the same invariant against the current shape.

## Verification

- `.github/audit/tests/` — 197 tests, 0 failures (bash 5 via `.gaia/scripts/bats5.sh`)
- `bash .gaia/tests/shell-lint.sh` — passed; now 198 tracked scripts, the new one among them
- `pnpm -C .gaia/cli test --run` — 2051 passed, incl. `audit-template-dogfood` (live == source template)
- three workflow copies byte-identical (`md5` match); build artifact regenerated via `bundle:adopter`, `.gaia/cli/gaia` unchanged
- `release manifest --check --json` — clean; new script answered `--ship` as `owned`

Quality Gate skips by its own rule: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.

Note this PR modifies `code-review-audit.yml`, so `self_modified` is `true` and CI's own audit stands down; the marker is earned by a local Code Audit Team spawn.

## Recorded, not fixed

Audit findings on paths this PR already changes. Every dispatched member has earned its marker, so folding any of these in would buy a full three-member re-dispatch (`.github/audit/**` is gate machinery, which rotates every digest) to re-earn clearance for a nit. None has a live failure mode, so they are recorded here per the digest economics in `wiki/concepts/PR Merge Workflow.md`.

From `code-audit-maintainer-shell`, all graded Suggestion:

- `ci-guard-paths.bats`: `fired_with()` consumes `reached_audit` in a word argument, so a `return 1` from the extractor becomes the empty string, which is itself the stand-down value. If `phase_step_body()` ever stopped extracting the step, the four adversarial stand-down fixtures would pass vacuously. Local vacuity only: the terminal-path and `audit phase:` tests still red the suite. `local ra; ra="$( … )" || return 1` is the fix.
- `ci-guard-paths.bats`: `assert_fired()` is unused; `assert_scenario()` is what every test calls. It bypasses `fired_with()`, so a future caller would evaluate the guards with `reached_audit` unset.
- `ci-guard-paths.bats`: `guarded_steps()` reads a name only from a `- name:` line, so a bare `- uses:` step carrying an `if:` would inherit the previous name and escape evaluation. All 32 steps are named today.
- `cra-status-upsert.sh`: `pr="$1"` under `set -u` is the one fatal path in a deliberately fail-soft script. Needs a caller bug, and every terminal step's call is pinned per-step by `cra-status-upsert.bats`.
- `cra-status-upsert.sh`: `head -n1` patches only the first marker comment, so a race past the lookup would strand a second one showing a stale verdict. `cancel-in-progress: true` makes this narrow, and this PR removed the systematic duplicate-creation path.

Carried from the earlier rounds:

- `cra-status-upsert.sh` resolves the repository two ways: the lookup and `PATCH` calls use `repos/${GITHUB_REPOSITORY}/…` while the create path uses `gh pr comment`, which resolves from the git remote. Identical in the runner today. `gh api --method POST repos/…/issues/<n>/comments` would unify them. Pre-existing shape, carried over deliberately to keep the extraction behavior-identical.
- `cra-status-upsert.sh`: `select(.body | contains(env.MARKER))` would error on a comment with a null body; `.body // ""` is the defensive form. Practically unreachable, and the error path is fail-soft.
- Transitional: a PR branched before this commit runs the new YAML against a head that lacks `.github/audit/cra-status-upsert.sh`, so the upsert exits 127 (fail-soft, no job failure). #1290 is the only open PR, so the blast radius is zero and it self-resolves on rebase.

Closes #1284
Closes #1285


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.SFbisXc2bi

Issue #726, opened 2026-07-14
tech-debt: post-audit-status.sh POSTs to an unpushed sha, so the local GAIA-Audit success status can never land on the common path


stevensacks:
**Location:** `.claude/hooks/post-audit-status.sh:130` (target sha), driven by the pre-push call sites at `.claude/agents/code-audit-frontend.md:818`, `.claude/agents/code-audit-maintainer-shell.md:117`, `.claude/agents/code-audit-maintainer-node.md:99`

Filed as an out-of-scope finding by `code-audit-frontend` during the audit of PR #722. **Pre-existing** (identical on `origin/main`); #722 changes the marker *key*, not the status POST's target sha or the stamp/mark/status/push ordering. Verified STANDS at Important by an adversarial refuter, and reproduced live during the audit run.

## Failure mode

The Code Audit Team's marker-write sequence is ordered **stamp -> mark -> status -> push**. Step 2b calls `post-audit-status.sh`, which resolves `head_sha=$(git rev-parse HEAD)` (line 130) and POSTs to `repos/<repo>/statuses/<head_sha>` (line 170). The stamp commit is not pushed until step 3.

On the **empty-commit placement path** -- which `audit-stamp-trailer.sh` takes whenever HEAD is *already pushed*, i.e. the normal case for a PR under review -- the stamp creates a new local commit GitHub has never seen. The POST therefore targets a SHA absent from the remote, the API returns 422, and line 179 emits `status: declined: post failed`. Nothing re-posts after the push, so the `GAIA-Audit` status never lands.

The amend path has the same defect: `audit-stamp-trailer.sh` only amends **un-pushed** HEADs, so the amended SHA is also absent from the remote.

**This is not a cosmetic degradation.** `/setup-gaia` registers `GAIA-Audit` as a **required status check** (see `.github/audit/tests/required-check-registration.bats:64-70`). A status that never lands leaves the required check unfulfilled, so branch protection blocks the merge for both the github.com button and a plain `gh pr merge`. On a mixed diff where a specialized member is dispatched, CI posts only `pending` by design, so the local POST is the **sole** producer of the success status.

### Reproduced live (audit of #722, HEAD a381b2d3)

With the marker written and all three member markers satisfied:

1. `post-audit-status.sh <marker>` -> `status: declined: post failed`
2. `git push` (stamp commit a381b2d3 lands on origin)
3. Re-ran the **identical** command, no other change -> `status: posted GAIA-Audit success a381b2d3`

[...]

PR #1295, opened 2026-08-10
debt(audit): one GAIA-Audit status writer, not five copies (#1286)

### LINKED ISSUE #1286: Five copies of the fail-closed GAIA-Audit status writer can diverge independently, and no PR exercises enough paths to catch it
`.github/workflows/code-review-audit.yml:747`

## Failure mode

Lines 747 to 1397 are roughly 650 lines, about 35% of the file, spent on four near-duplicate status-writing steps plus the stand-down step that posts the same status:

| Line | Step |
|---|---|
| 747 | `Write GAIA-Audit commit status` |
| 886 | `Write GAIA-Audit commit status (clean, no push)` |
| 1041 | `Stand down (local-mode, no override)` |
| 1125 | `Write GAIA-Audit commit status (out-of-scope skip)` |
| 1271 | `Write GAIA-Audit commit status (chore-deps skip)` |

Each block independently recomputes the frontend member's content digest via `.gaia/scripts/audit-member-digest.sh`, resolves the tree, applies its own copy of the fail-closed rule ("never post a status on an unrecomputable digest"), and POSTs. That fail-closed rule is the most safety-critical logic in the file, and it exists in five copies.

The concrete failure: a correction to the digest or fail-closed logic gets applied to one copy and silently leaves the others stale. Because each block guards a different terminal path, the stale copies execute only on paths a given PR may never exercise, so the divergence survives CI indefinitely and surfaces as an inconsistent merge gate, the same status meaning different things depending on which path produced it.

## Suggested fix

Extract one `.github/audit/write-audit-status.sh` taking the axes the call sites actually vary on:

```
write-audit-status.sh --sha <sha> --state <success|pending> --description <text>
```

and reduce the five blocks to thin callers.

**Sequence this behind the two cheaper items and write the tests first.** The call sites are not trivially uniform: they differ in which SHA they stamp (`steps.push-fixes.outputs.audit_sha` versus `github.event.pull_request.head.sha`, and the comments at 886 explain the HTTP 422 that follows from getting that wrong), in state (`success` versus a `pending` stand-down), and in guard. Unifying them means first settling the writer's contract, which is a real design decision rather than a mechanical extraction.

`.github/audit/tests/post-audit-status.bats` and `ci-clean-no-push-status.bats` already exist. Extend those to pin current behavior on all five paths, get them green, and only then extract against them.

Handler: plan

stevensacks:
Closes #1286

## What was wrong

`code-review-audit.yml` finishes down five terminal paths, and each carried its own copy
of the same safety-critical rule: recompute the frontend member's content digest or post
nothing, never clobber a live `success` for that digest, only then stamp. Five copies of
the most safety-critical logic in the file meant a correction landed on one and left the
others stale, and because each copy guards a path a given PR may never exercise, the
divergence survived CI indefinitely and surfaced as one status meaning different things
depending on which path produced it.

542 lines of duplicated shell are now 117 lines of call-site rationale plus one shipped
script, `.github/audit/write-audit-status.sh`.

## The decision this issue owed

The issue routed here because unifying the writers "means first settling the writer's
contract, which is a real design decision rather than a mechanical extraction". Settled
from the five call sites:

```
write-audit-status.sh --sha <sha> ( --base <pr-base-sha> | --force-pending <description> )
                      [--require-marker] [--success-post-non-fatal]
```

**Two modes, not one flag set.** `--base` resolves the members CI cannot clear and can
stamp success; `--force-pending` never can, which is the local-mode stand-down where CI
dispatches nobody. The five call sites are two shapes, and pretending otherwise would mean
a flag that turns off half the script. The stand-down is included rather than left out
because the non-clobber predicate is the safety-critical duplication and it is that path's
entire body.

**Two modifiers, gated mode only.** `--require-marker` is the clean-no-push path's
proven-clean check. `--success-post-non-fatal` is its #726 fix.

## One divergence is preserved on purpose, not resolved

Three of the four success writers leave the `gh api` bare, so a rejected POST reds the job;
the clean-no-push writer wraps it. **Neither unification is behavior-preserving:** the two
skip paths' `success_stamped` output ordering depends on the bare POST failing the step
under `set -e`, and non-fatality is half the fix for #726, where an HTTP 422 on an unpushed
sha turned an otherwise clean audit red. Consolidating must not silently pick a winner, so
the divergence is carried as a flag, pinned per path by new tests, and surfaced in the
script's docblock. Whether the four *should* agree is a separate question and is filed.

## Why the behavior claim is evidence rather than assertion

`ci-status-member-gate.bats` extracts each step's real `run:` body from the YAML and
executes it in a git sandbox against a `gh` mock. **The same 60 behavioral tests pass
before and after the extraction without being touched**, which makes them an equivalence
oracle rather than a description of the new code.

Before extracting, the fail-closed preconditions were pinned on *every* path rather than
the paths that happened to have a test: version guards existed on two paths, a digest guard
on one, the sha guard nowhere. 13 characterization tests were added and every one was
mutation-proved against a block-scoped edit to the step it guards.

**Two of my own tests were hollow and the mutation pass caught both** (neither was caught
by review):

- asserting "no status posted" on the clean-no-push digest path passes either way, because
  a step that lost the guard finds no marker under the bogus digest and posts nothing
  anyway. It now asserts where the step stopped.
- every malformed call to the new script exits 2, so asserting the status alone could not
  tell a parser refusal from a later guard catching the same call. It now asserts which
  refusal fired.

The shared shape is worth carrying: **an outcome-only assertion on a path whose next branch
declines for its own reasons is hollow by construction.**

## Tests that had to change, and why none was weakened

Four grep-based structural tests pinned literals that moved into the script. Each was
re-pointed at **both** ends rather than relaxed, because a flag nobody reads and a flag
nobody passes fail the same way:

| Test | Now asserts |
|---|---|
| every status step resolves members over the full-PR base | the step passes `--base "${PR_BASE_SHA}"`, and the writer feeds it to the gate |
| the sha comes from the event payload | the step passes `--sha "${HEAD_SHA}"`, and the writer stamps the sha it was handed |
| clean-no-push marker + non-fatality (`ci-clean-no-push-status.bats`) | the step passes the flags, and the writer implements them |
| `chore-deps-audit-stamp.test.ts` | the step invokes the writer member-awarely, and the writer POSTs the required context |

**The pending-writer structural lock comes out stricter.** It used to loop the five step
bodies and pin the count at 5; a sixth writer could pass by carrying its own correct copy
of the guard. It now asserts the workflow POSTs pending **nowhere**, that exactly one
pending POST exists and is guarded, and that all five steps route through it.

## One ordering change, invisible where it applies

The member gate now runs before the digest recompute on all four gated paths, matching what
the two skip paths already did so they keep publishing `members_pending` when the digest
fails. The two stamp paths publish no outputs and carry no `id:`, so nothing can observe the
difference, and `gate-pending-members.sh` exits 0 on every branch, so it cannot red a step
that previously exited earlier.

## Distribution

Mirrored to the source template and the build artifact; all three copies byte-identical.
The new script is recorded `owned` in the manifest by the release CLI, so adopters receive
it — the shipped template invokes it, so withholding it would break their rendered workflow.

One repair worth naming: the script's docblock originally named release-excluded test paths,
and the `maintainer-paths` leak-check failed the bundle on exactly that. Caught by the
distribution harness, not by review. The distribution rule records this class as one that
already regressed once.

## Verification

| Check | Result |
|---|---|
| `bats .github/audit/tests/` | 220 ok / 0 not ok (198 on `main`) |
| `bats .gaia/scripts/tests/` | 1508 / 0 |
| `bats .gaia/tests/hooks/` | 1378 / 0 |
| `bats .gaia/tests/lib/` | 448 / 0 |
| `bats .gaia/tests/statusline/` | 36 / 0 |
| `bash .gaia/tests/shell-lint.sh` | pass |
| `shellcheck write-audit-status.sh` | clean |
| `bash .gaia/tests/distribution/run-all.sh` | 17 / 17 |
| `bats .gaia/scripts/tests/region-marker-conformance.bats` | 6 / 6 |
| `pnpm -C .gaia/cli test --run` | 136 files, 2051 passed, 1 skipped |
| `pnpm typecheck` / `lint` / `test --run` / `build` | all pass |
| `pnpm pw` | 8 passed, 1 skipped |
| dev smoke | HTTP 200 |
| three workflow copies byte-identical | yes |

## Simplify pass

Four review agents. Two findings applied: the four fail-closed exits now go through one
`decline()` helper, and both modifiers are refused in stand-down mode rather than silently
accepted. One declined, with the reason recorded in the commit: collapsing the two
modifiers into one caller-named flag would make the script know a caller's identity instead
of its own behavior, and would bury the fatality divergence the docblock deliberately
surfaces. One is out of this diff and is filed.

## Findings filed rather than waived

Both are filed rather than recorded here, and both are `surface:adopter`.

**#1296 — the four success writers give three different answers to a rejected status POST.**
Waive-eligible (non-security, on a path this PR changes), but DRAIN rule 4's override fires:
the class is observed to have spread, since the clean-no-push writer was repaired for exactly
this (#726) and its three siblings were not. This PR preserves the divergence rather than
picking a winner, so the filing is what makes it drainable.

**#1297 — the `.gaia/VERSION` read-and-normalize idiom is copy-pasted across five shipped
files with no shared helper.** Four of the five sites are paths this PR does not change, so
the touched-file waive never reaches it. Same silent, fail-quiet shape as #1286 on a
different primitive: five implementations of one equality that the audit trailer and the
commit-status description both compare for a match.

## Audit round 1

Three members dispatched, all three cleared: `code-audit-github-workflows`,
`code-audit-maintainer-node`, `code-audit-maintainer-shell`. **Zero Critical, zero
Important.** No member disputed the two-mode contract or the deliberately preserved
success-POST fatality divergence.

Five Suggestions, **all five applied in one commit** rather than accepted-and-noted. Four of
them live in `.github/audit/**`, which is gate machinery, so they rotate every member's
digest *together*; the fifth is in a file whose member re-dispatches anyway. The marginal
audit cost of each additional fix was therefore zero, which inverts the usual
accept-and-note economics.

- **The writer reached its siblings by CWD.** It resolved a repo root for the digest
  recompute, then reached `gate-pending-members.sh`, `audit-member-digest.sh`,
  `audit-success-present.sh`, the clean marker, and `.gaia/VERSION` relative to CWD. Inline
  in a `run:` body those are the same path; standalone they are not, so a subdirectory run
  would compute one tree's digest and read another tree's marker. Now resolved once and
  anchored, in the same spelling `gate-pending-members.sh` and `resolve-audit-base.sh`
  already use. An empty root is left unguarded on purpose: it preserves today's split, where
  gated mode reds the step and stand-down mode declines.
- **The argument-contract suite did not actually neutralize `gh`.** It prepended an *empty*
  directory to PATH under a comment claiming "no `gh` on PATH". An empty directory shadows
  nothing, so on a runner a refusal path that ever grew a POST would have issued a real
  commit-status write from the test suite. Now a stub that exits 127. Verified both ways:
  the empty directory leaves a real `gh` resolving, the stub does not.
- **A docblock credited two step outputs to the wrong mode.** `success_live` / `read_failed`
  come from the shared non-clobber read, not from the stand-down, and only the two skip
  paths consume them.
- **Two guards pinned the writer's source text rather than its behavior.** The digest grep
  pinned a `)"; then` tail, so a behavior-preserving reformat would red the suite with a
  message reading as a lost fail-closed guard. That tail asserted the call sits in a
  fail-closed `if`, which is a *behavioral* claim, and is already asserted behaviorally by a
  test that executes the step with the digest engine removed and requires it to stop before
  the marker branch. The tail is dropped; the behavioral assertion was confirmed present
  first. The gate-call grep became a regex so it pins adjacency rather than quoting.
- **A CLI test resolved the repo root by counting `..` levels.** Replaced with
  `resolveRepoRootFromImportMeta`, the helper its sibling in the same directory already uses.

**Every re-pointed guard was mutation-proved**, restoring with `cp` rather than
`git checkout --`: a tree-keyed marker reds the marker assertion, a dropped `--ref "${sha}"`
reds it, and a writer resolving its own base instead of the handed one reds four tests. The
ordering guard was also checked against the hollowness trap this branch documented earlier:
`audit-member-digest.sh` appears on exactly one line of the writer, so `head -1` cannot pick
a comment and satisfy the ordering vacuously.

**No workflow YAML change and no bundle regeneration.** The five call sites are untouched;
only the writer's internals moved. The three byte-identical workflow copies stay identical.

### Deliberate non-findings, recorded so they are not re-litigated

- Moving gate logic out of `.github/workflows/**` means a future PR touching only
  `write-audit-status.sh` gets `has_source=false` and `self_modified=false`, so the
  out-of-scope step executes that PR's own modified script. Pre-existing (two siblings
  already sat in that position), mitigated on both sides, and not adversary-reachable: a
  fork `pull_request` gets a read-only token and no `statuses: write`.
- `expect(writer).toContain('gate-pending-members.sh')` is satisfiable by a comment
  occurrence, so it alone cannot detect removal of the real invocation. Already covered
  behaviorally by `ci-status-member-gate.bats`, which executes the call site.

## Audit round 2

Same three members, all three cleared again on the repair commit. **Zero Critical, zero
Important.** Each round-1 repair was independently re-verified rather than taken on trust:
the shell member re-ran the behavioral assertion the loosened grep now leans on and confirmed
it is genuinely discriminating, and confirmed the widened negative marker greps got
*stronger* rather than weaker.

Two Suggestions, both landing on round 1's own repairs. **Both accepted-and-noted and filed
as #1298 rather than fixed in-round**, per the stopping rule written before the round ran:

- `chore-deps-audit-stamp.test.ts:86` — `expect(writer).toContain('gate-pending-members.sh')`
  is satisfiable by a comment, and is already non-discriminating today: the writer names that
  script in two comments besides the real call.
- `ci-clean-no-push-status.bats:88` and the ordering guard at `:111` — round 1 loosened this
  pin to a bare substring when the path moved behind the `$repo_root` anchor, so a future
  comment naming the script would satisfy it and feed the ordering guard a comment's line
  number. Not hollow today (the writer names it exactly once, at the call).

**Why filed rather than fixed.** Both are graded Suggestion by their own finders, both are
explicitly no-live-failure, neither gates a marker, and the behavioral claim under both is
separately pinned by a test that executes the step with the digest engine removed and
requires it to stop before the marker branch. Fixing them buys a third full three-member
round for no proven reduction in risk. Where the round landed, stated rather than left as a
count: the **oracle**, not the parser, comparison, payload, or design. Two consecutive rounds
have now found only oracle-strength nits and the artifact under test has been stable since
round 1.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.5ETeAAVi4m

Issue #1030, opened 2026-07-22
Two required checks can outlive the self-heal poller window, leaving their contexts absent on the new HEAD


stevensacks:
**Location:** `.github/workflows/tests.yml:18` (the `tests` job's `timeout-minutes: 60`), and `.github/workflows/chromatic.yml` (neither job declares `timeout-minutes` at all)

## Failure mode

Two required checks can outlive the self-heal re-trigger poller's window, which leaves their contexts **absent** on a self-heal HEAD rather than red. That is the same permanent merge wedge #1019 fixed for the other three required checks, reached by a different route.

The poller in `.github/workflows/code-review-audit.yml` waits `for _ in $(seq 1 150)` with `sleep 10`, roughly 25 minutes. On expiry it logs `did not complete within 25 min; not stamping` and returns 0, so nothing is stamped onto the self-heal HEAD and nothing fails loudly.

Of the five entries now in `retrigger_workflows`:

| Entry | Job | `timeout-minutes` | Under the ~25 min poller window? |
|---|---|---|---|
| `CLI Tests` | `Vitest (.gaia/cli)` | 15 | yes |
| `Audit CI Tests` | `Audit CI Tests` | 15 | yes |
| `Distribution Audit (PR)` | `Distribution Audit` | 15 | yes |
| `Tests` | `Vitest and Playwright` | 60 | **no** |
| `Chromatic` | `Run Chromatic` | none (6-hour runner default) | **no** |

A Playwright run past 25 minutes, or a hung Chromatic upload, reproduces the wedge. The dispatch lane skips `tests.yml`'s scoping step, so a re-triggered run executes the full Vitest plus Playwright suite rather than the incremental subset a pull request would.

`.gaia/scripts/tests/retrigger-reachability.bats` asserts three links of this invariant (the trigger, the job `if:`, and membership in the knob) and not this fourth one, so nothing catches it.

Latent while the repo runs `default_mode: local`, which never exercises the CI self-heal lane. Reachable the moment any author resolves to `ci`, which a fork pull request does unconditionally.

## Suggested fix

Add `timeout-minutes` comfortably below the poller window to `tests.yml`'s `tests` job and both of `chromatic.yml`'s jobs, then extend `.gaia/scripts/tests/retrigger-reachability.bats` with a fourth assertion that every job in a `retrigger_workflows` workflow declares a `timeout-minutes` under the poller ceiling.

[...]

PR #1299, opened 2026-08-10
ci: bound job runtimes, default-deny workflow tokens, de-duplicate the bundle gate

stevensacks:
Five mechanical fixes from a workflow audit of this repo. `code-review-audit.yml` is deliberately untouched (byte-identical triplicate, handled separately).

## 1. `tests.yml`: drop the container action

`martinbeentjes/npm-get-version-action` is a **container** action, so GitHub builds its Docker image during job setup, *before* step-level `if:` gating applies. Measured at **15s of a 24s run** in which every subsequent step was filtered out and nothing else ran.

Replaced with an inline read. Same step id (`package-version`) and same output name (`current-version`), so the two `npm_package_version:` consumers (Vitest, Playwright) are untouched.

Deviation from the literal one-liner in the brief, and why: `echo "current-version=$(jq -r ...)"` makes `echo` the status-bearing command, so a jq that cannot read `package.json` — or a runner image without jq — writes an empty version and passes **green**, with both suites then running under an empty `npm_package_version`. The value is captured first and `jq -e` makes a null/absent `.version` an error. Probed: missing / malformed / absent-version now exit 2 / 5 / 1 instead of 0; happy path unchanged.

**Expected saving:** ~15s per `tests.yml` run, paid on every run including the ones that filter out all work.

## 2. `.github/dependabot.yml`

43 hand-maintained full-SHA pins across 13 workflows, no bump automation, and five workflow headers asserting pin parity in prose only.

- **Grouped** into one PR: any `.github/workflows/**` edit arms Audit CI Tests (~13 min), so ungrouped bumps would pay that per action.
- **Prefix `ci`** + `include: scope` → `ci(deps): bump ...`. Verified against the real `gaia ci-check-subject` binary. Not `chore(deps)`: that is the signal `tests.yml` reads as "a human already ran the quality gate locally", and Dependabot runs no local gate, so it must not borrow that skip.
- **Not covered, knowingly:** pins in `.gaia/cli/src/automation/templates/workflows/*.tmpl` are outside `.github/workflows/`, so Dependabot cannot see them; they stay hand-maintained.
- **Expect one recurring red:** a bump touching `code-review-audit.yml` desyncs it from its two template copies until mirrored. That red is the drift guard working; documented in-file.

**Distribution answer: withheld** (category 9). It opens scheduled PRs against a `.github/` tree that is 11-of-13 maintainer-only and absent on an adopter clone; GAIA's adopter CI installs on demand via `/setup-gaia` for that reason, and the pins in the two shipped workflows reach adopters as an `/update-gaia` merge.

## 3. Caps and default-deny tokens

Job caps added, sized against measured runtime with a wide margin: `pr-title.yml` 5m (12s, ~25x), `verify-required-checks.yml` 5m (13s, ~23x), `shell-lint.yml` 10m (58s, ~10x — wider because the one hang-capable step is a github.com download), `release.yml` 30m. The release job is the sharpest: it capped only its distribution-gate step, so a hang anywhere else held a runner for the 6-hour default **while holding `contents: write`**, the widest token in the directory.

Top-level `permissions: {}` added to `tests.yml`, `audit-roster.yml`, `pr-title.yml`, `shell-lint.yml`, `verify-required-checks.yml`. All five are single-job and every job already declares its own scopes, so nothing drops to zero permissions today; this only stops a later-added job inheriting the repo default.

After this PR: **13/13 workflows declare top-level permissions and 13/13 jobs carry a runtime cap.**

## 4. De-duplicate the bundle-freshness gate

The "Verify committed CLI binaries and templates are rebuilt from src" step was two verbatim copies in `cli-tests.yml` and `release.yml`, kept in lockstep by a comment asking a human to remember — on a required check, so what one copy missed merged. Extracted to `.gaia/scripts/verify-cli-bundle-fresh.sh`; both callers invoke it.

Behaviour preserved; the fixed `/tmp` paths become a `mktemp -d` with an EXIT trap, so the script no longer leaks scratch files or `rm -rf`s a fixed path when a maintainer runs it by hand.

**Distribution answer: withheld** (category 4). It rebuilds from `.gaia/cli/src`, which adopters do not receive, so on their clone it could only fail or mislead.

## 5. Stale issue citation

`audit-ci-tests.yml`'s timeout comment cited #1030 for machine-checking the caps. #1030 is closed and never scoped that work. Replaced with a plain statement of the limitation; no replacement issue opened.

---

## Audit roster gap this surfaced

`.github/dependabot.yml` initially dispatched **no** auditor — the exact class fixed by #1245/#1282, reintroduced by adding the file. `verify-audit-roster.sh` then caught a second lockstep requirement, the builtin fallback roster in `.claude/hooks/lib/audit-scope.sh`. Granting it to `code-audit-github-workflows` was the wrong owner and CI caught it: that member's roster entry, agent definition, and the fallback roster all **ship**, so the grant put a release-excluded path into a shipped script and `release runtime-deps` failed every distribution scenario at build-staging. It now belongs to `code-audit-maintainer-shell`, which is release-excluded end to end and already owns the same shape of maintainer-only `.github/` governance file (`.github/CODEOWNERS`). No adopter-visible change. Roster clean.

## Verification

- All 9 edited/new YAML files parse.
- `bash .gaia/tests/shell-lint.sh` passes with the new script (200 tracked scripts).
- bats, full Audit CI Tests matrix, run locally: `.gaia/scripts/tests/` 1508, `.gaia/tests/hooks/` 1378, `.gaia/tests/lib/`, `.github/audit/tests/`, `.gaia/tests/forensics/`, `.gaia/tests/statusline/` — all green.
- `bash .gaia/scripts/verify-audit-roster.sh` clean.
- `release manifest --check --json`: 0 missing, 0 region drift. Both new files confirmed excluded via `release exclude-regex`, with control paths still shipping.
- CHANGELOG `## [Unreleased]` entries landed.

Note: #1273 (StrictMode timing canary in `tests.yml`, ~2 failures in 11 runs) is a known flake and not chased here.

---

## Audit outcome

Two rounds. `code-audit-maintainer-shell` cleared on round 1; `code-audit-github-workflows` refused on round 1 with one Important finding, which is fixed in 40beb2a1, and cleared on round 2.

**Important finding, fixed.** Extracting the bundle-freshness gate into a script moved it *out* of `cli-tests.yml`'s `code:` paths filter. The gate used to be inline on a filtered path, so editing it re-ran it on its own PR; as a script it was unfiltered, so a PR touching only that script would resolve `code=false`, skip the step, and report the required check `Vitest (.gaia/cli)` green having run nothing. `release.yml` is no independent second opinion now that it runs the same script, and `shell-lint.yml` sees only what shellcheck sees — dropping the template `diff -rq` or inverting a `cmp` would pass both lanes. Filter entry added with rationale.

**Suggestion applied** (free: same member's digest was already rotating): `release.yml` moves to top-level `permissions: {}` with `contents: write` on the publish job. Effective permissions are unchanged — a workflow-level block already sets unlisted scopes to `none` — so this buys nothing today and removes the inherited grant a future second job would pick up, in the one file whose token can push.

### Accepted residuals, not fixed here

Recorded rather than folded in: each would rotate an already-marked member's digest and buy a full re-dispatch to re-earn a clearance, for changes with no live consequence.

1. **`audit-ci-tests.yml` keeps top-level `contents: read` + `pull-requests: read` with no job-level block.** Same inheritance shape the six siblings in this PR just fixed, and this PR does edit the file (comment only). Left out because it was not in this change's scope and effective permissions are identical either way; `distribution-audit-pr.yml` and `distribution.yml` share the shape and are untouched here. Worth a follow-up that does all three together.
2. **Nothing machine-checks that a step gated on a `dorny/paths-filter` output has its inputs listed in that filter.** This PR's Important finding is the third instance of that class (`shell-lint.yml`'s own comments record two earlier ones, each closed by hand after the fact). A guard is more machinery than this PR should carry.
3. **`verify-cli-bundle-fresh.sh` snapshots the working tree while its docblock and error strings say "committed".** Local-invocation-only: both callers run on a fresh `actions/checkout` where tree and HEAD agree. Pre-existing behaviour, faithfully preserved from the inline copies rather than introduced here.
4. **The script is committed `100644`.** Both callers invoke it as `bash <path>`; two sibling scripts share the same shape.
5. **A weekly Dependabot PR cannot produce the `GAIA-Audit` status from CI.** Dependabot-triggered runs get a read-only token and no secrets. With `default_mode: local` a maintainer already runs the gate locally on every PR, so the merge is no more blocked than usual; the delta is weekly red-job noise. Raised independently by both members.

Not raised as a defect, recorded as context: the SHA pins in `.gaia/cli/templates/workflows/*.tmpl` have no live counterpart and so no drift guard, meaning adopter-template pins now freeze while the maintainer's own move weekly. Pre-existing, disclosed in the new config's own comment, and owned by a different member.


stevensacks:
@/var/folders/pn/m58c653j1js2r48m5n358ync0000gn/T/tmp.SbQ7i1aPf7

PR #1608, opened 2026-08-27
test(capability-oracle): the wide predicate's coverage is decided by bash, not remembered (#1605)

### LINKED ISSUE #1605: capability-oracle-anchors: the wide predicate is not wider than the narrow one, and its stated limit is false
`.gaia/scripts/tests/capability-oracle-anchors.bats:105`

## Failure mode

`position_test_constants` is the WIDE predicate that exists to hold the narrow `anchor_names` derivation to account: the suite asserts the two sets are equal, so a spelling only the narrow one misses becomes loud instead of silently shrinking every anchor-driven arm in the file at once.

It is not wide enough to do that job, and it is narrow in a way its own header denies. The awk body strips a leading `readonly` (line 105) or `declare -<flags>` (line 106) and then matches names as `_GAIA_CAPCHECK_[A-Za-z_]*` (line 107), which is the **same** name shape the narrow predicate uses at line 80. Four spellings are therefore invisible to **both** predicates:

- `export _GAIA_CAPCHECK_X='(^|...'`
- a bare `declare` with no flag: `declare _GAIA_CAPCHECK_X='(^|...'` (line 106 requires `-[a-zA-Z]+`)
- stacked keywords: `readonly declare _GAIA_CAPCHECK_X='(^|...'`
- any digit-carrying name: `_GAIA_CAPCHECK_X2='(^|...'`

For each of those, wide and narrow agree, the set-equality arm stays green, and every anchor-driven arm in the file silently drives one fewer anchor while its name goes on claiming every anchor. That is exactly the partial-respelling state the arm was added to make loud.

Verified by running the awk body verbatim over a fixture carrying plain, `readonly`, `declare -gr`, `export`, bare `declare`, stacked-keyword and digit-suffixed spellings. The wide predicate emitted only the plain, `readonly` and `declare -gr` forms; it dropped the other four. Nothing is broken today: the library carries zero `export`/`readonly`/`declare` lines and no anchor name carries a digit, so this is latent, not live.

The function's own honest-limit paragraph is false about its own predicate. It frames what falls outside as constants "assigned some other way, through an `eval` or a nameref", which is "outside what any line-wise reader sees". An `export` line is plainly something a line-wise reader sees, and so is a bare `declare`.

## The class question, which is why this is filed rather than patched

This is the **third** instance of one class in this one file, and the third instance lives inside the second instance's own repair:

- **Round 1** of PR #1604's audit refused: the scanner roster was hand-written and narrower than its name, with a comment claiming an arm caught an omission that did not exist.
- **Round 2** found `anchor_names` was the one derivation with no short-read guard.
- **Round 3** found that the guard written to close round 2 reads narrower than its own stated limit.

All three times the deterministic battery was green (14/14 under bash 5, shell-lint clean, whole-tree invariants pass). A suite written to catch guards whose coverage is narrower than their names has now carried that defect three times, at three different derivations.

Adding `sub(/^export[ \t]+/, "", line)` would be a fourth per-site memory that closes one spelling of four and leaves the general problem intact. The question worth deciding first is whether every derivation in this file should have its coverage claim **derived and enforced** (a rule the suite checks) rather than stated in prose beside it and remembered per site.

## Suggested fix

Settle the class question first, then repair accordingly. If the answer is per-site after all, the minimum is: widen the keyword strip to one alternation covering `export`, bare `declare`, `declare -<flags>` and stacked keywords; admit digits in the wide predicate's name shape so it is genuinely wider than the narrow one; and rewrite the honest-limit paragraph so the stated limit matches what the predicate actually reads.

stevensacks:
Closes #1605

## What was wrong

`.gaia/scripts/tests/capability-oracle-anchors.bats` drives every arm off `anchor_names`, so a spelling that derivation misses does not shrink one arm, it shrinks all of them at once and each still reports green. `position_test_constants` is the wide predicate that exists to make that loud, by a set-equality arm against the narrow one.

It was not wider. It stripped `readonly` and `declare -<flags>` and then matched the **same** digit-excluding name shape the narrow predicate uses, so on `export`, a bare `declare`, stacked keywords and any digit-carrying name, wide and narrow agreed vacuously. Its honest-limit paragraph framed what fell outside as `eval` or a nameref, "outside what any line-wise reader sees", which is false about an `export` line.

## Why this is not the one-line repair the issue offered

The issue's minimum fix would have been the fourth per-site memory in one file for one class: a guard whose coverage is narrower than its stated limit. A second in-file predicate cannot close it, because whatever checks the checker is then making an unchecked claim of its own, one level out. The regress terminates only at an authority outside this file's vocabulary, and for "does this line define a shell constant" that authority is bash, already this suite's interpreter.

## What nine review rounds established, which changed the answer twice

The first version kept a regex predicate and checked it against bash over a **generated space** of spellings. Three rounds each falsified that space, every one against a green suite: no leading-indent axis; then the indent axis driving spaces only, where the obvious repair was a silent no-op because the transport was tab-delimited and a tab is IFS whitespace; then three undriven branches at once plus a header sentence claiming an axis per branch that those branches falsified.

The common cause was none of those branches. **Consulting an authority over an enumerated set of inputs is not the same as being one.** The generated space stood between the predicate and bash, and that space was the coverage claim the authority had been brought in to retire.

Later rounds found the same class at the file's other derivation, then found that its repair had reintroduced the class one axis over, then found a rot-prone cardinal in a comment. Each is recorded below.

## What ships

- **Both wide derivations are bash.** `position_test_constants` sources the library in the suite's own `$BASH` and reports the constants that then exist carrying an anchor value. `library_scanners` diffs `declare -F` across the source, so membership is what loading the library adds. `fn_body` is `declare -f`. None of them approximates bash, so none has branches to cover or an input space to enumerate.
- **Deleted**: the regex predicate, its frozen pre-widening control, the spelling generator, the `bash_defines` oracle, the differential arm and its control, `SPELL_VALUE`, and the function-name regex. 16 arms became 15.
- **Membership is bounded by the source, not by a name.** Asking bash and then keeping only `_gaia_capcheck_*` put the defect back on the name axis: nothing in the tree enforces that prefix, so a differently-named scanner was absent from the derived set and the hand-written roster alike and the membership arm read them as equal. Driven: such a scanner left the arm green before the fix and reds it after.
- **A new arm for the other direction**: a `local` line defines nothing at file scope, so neither predicate may claim it. Previously prose, now an arm.
- **The header rule stands, with a corollary**: where the authority can answer directly, let it, so the enumeration disappears rather than needing a guard of its own.

## Honest limits, stated rather than left to be found

- The equality arm's signal is a **disagreement**, so a respelling both predicates miss keeps them equal and green. Probing surfaced one such shape: an anchor moved inside a function *and indented with it*. At column zero, or in a function called at load, they diverge and the arm reds. The regex predicate missed it identically, and such an anchor would leave the detectors composing against it referring to nothing.
- The roster control's rows are **load-bearing and unguarded**, unlike the anchor control's, whose rows are evidence and cost nothing if dropped. Said rather than guarded, because no authority outside this file can enumerate the spellings bash accepts that a regex refuses.
- Neither comment states a cardinal. `.claude/rules/bats-assertions.md` prohibits counts in comments describing a set, and this branch demonstrated the rot on itself: a row landed two commits before its cardinal caught up, green throughout, and a human counting rows is what found it.

## Verification

- 15/15 under bash 5 in ~5.3s (was 16/16 at ~10.9s).
- Every guard proved able to fail, by mutation, in-tree, with an unmutated control reding nothing: wide predicate collapsing to narrow; dropping its `(^|` value test; reading text again and stripping `local`; every anchor control row becoming a no-op; the double-quote escaping broken; `fn_body` reverted to the text reader; discovery reverted to the regex; the name filter reinstated; the keyword-anchor filter dropped.
- Dropping a control row from the **anchor** list survives by design and is documented as such.
- Driven end to end, not argued: a digit-named scanner and a non-prefixed scanner each left the roster arm green before their respective fixes and red after.
- `shell-lint.sh` clean including the `*.bats` shellcheck pass at severity=warning; `whole-tree-invariants.sh` exit 0.
- Nine audit rounds; the last returned no findings.
- No CHANGELOG entry: `.gaia/scripts/tests/` is release-excluded, so nothing here reaches an adopter.

### DIFF
--- a/.gaia/scripts/tests/capability-oracle-anchors.bats
+++ b/.gaia/scripts/tests/capability-oracle-anchors.bats
@@ -146,11 +146,11 @@ anchor_names() {
 # is nobody's reading of a line.
 #
 # Two things bash decides here for free, which the hand-widened regex got
-# wrong in both directions. `local` at file scope is an error defining nothing, so a
-# `local` line contributes no name here and cannot be claimed; a strip list
-# growing `local` as a keyword did claim one. And a line that does not parse
-# aborts the load, so it surfaces as a short set against anchor_names rather
-# than as a name nobody can use. Neither needed a rule.
+# wrong in both directions. `local` at file scope is an error defining
+# nothing, so a `local` line contributes no name here and cannot be claimed;
+# a strip list growing `local` as a keyword did claim one. And a line that
+# does not parse aborts the load, so it surfaces as a short set against
+# anchor_names rather than as a name nobody can use. Neither needed a rule.
 #
 # The nested interpreter is `$BASH`, the one running this suite, rather than
 # whatever `bash` resolves to on PATH. The library refuses to load under bash
@@ -526,12 +526,12 @@ EOF
   # scanner appended, composing against an anchor the real library already
   # carries so the copy stays loadable, and the same derivation runs over it.
   #
-  # Three spellings, and the last two are what keep this derivation on bash.
+  # Four spellings, and the last three are what keep this derivation on bash.
   # Every function in the library today is written the one way the regexes this
   # derivation used to use could read, so reverting them breaks nothing
   # measurable against the real library and the conversion would be free to
-  # rot. Each of the two added spellings is a definition bash accepts that one
-  # of those readers refused, so each pins one of them:
+  # rot. Each of the three added spellings is a definition bash accepts that
+  # one of those readers refused, so each pins one of them:
   #
   #   `_v2` carries a digit, which the discovery regex's `[a-z_]*` refused.
   #   `function NAME {` carries no parens, which BOTH the discovery regex and
--- a/.gaia/scripts/tests/capability-oracle-anchors.bats
+++ b/.gaia/scripts/tests/capability-oracle-anchors.bats
@@ -526,12 +526,12 @@ EOF
   # scanner appended, composing against an anchor the real library already
   # carries so the copy stays loadable, and the same derivation runs over it.
   #
-  # Four spellings, and the last three are what keep this derivation on bash.
-  # Every function in the library today is written the one way the regexes this
-  # derivation used to use could read, so reverting them breaks nothing
-  # measurable against the real library and the conversion would be free to
-  # rot. Each of the three added spellings is a definition bash accepts that
-  # one of those readers refused, so each pins one of them:
+  # The rows below, and every one after the first is what keeps this derivation
+  # on bash. Every function in the library today is written the one way the
+  # regexes this derivation used to use could read, so reverting them breaks
+  # nothing measurable against the real library and the conversion would be
+  # free to rot. Each added spelling is a definition bash accepts that one of
+  # those readers refused, so each pins one of them:
   #
   #   `_v2` carries a digit, which the discovery regex's `[a-z_]*` refused.
   #   `function NAME {` carries no parens, which BOTH the discovery regex and
--- a/.gaia/scripts/tests/capability-oracle-anchors.bats
+++ b/.gaia/scripts/tests/capability-oracle-anchors.bats
@@ -136,6 +136,7 @@ position_test_constants() {
   awk '
     {
       line = $0
+      sub(/^[ \t]+/, "", line)
       while (sub(/^(export|readonly|declare|typeset)[ \t]+/, "", line)) {
         while (sub(/^[-+][a-zA-Z]+[ \t]+/, "", line)) { }
       }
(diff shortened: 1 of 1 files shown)


stevensacks:
## Round history so far (2 rounds, both clean passes)

One member is dispatched for this diff: `code-audit-maintainer-shell`. Both rounds cleared with an earned marker and exactly one Suggestion each, and both Suggestions were the same defect one axis further out: **the generated input space is itself a coverage claim.**

**Round 1** — the space had no leading-whitespace axis, so an indented top-level assignment was outside the region bash decides coverage over. Reproduced end to end: a headerless indented anchor appended to the library left the suite 16/16 green, including the arm whose entire subject is an anchor with no header. Repaired by adding the indent axis and a leading-whitespace strip ahead of the keyword loop.

**One clause of round 1's suggested fix was declined.** It also asked for `anchor_names` to receive the matching widening. It did not, because every spelling the wide predicate learns and the narrow one still cannot is a spelling that reds the equality arm the moment the library carries one, and that redness is the signal. Round 2 was told about the decline and upheld it: keeping the pair divergent fails closed, teaching both sides the same spelling fails open on exactly the axis this suite guards.

**Round 2** — the new indent axis drove spaces only, so the tab half of the predicate's own `[ \t]` class was undriven; narrowing that class to `[ ]` left all sixteen arms green. The trap it found is why this was not a one-line axis addition: the obvious repair is a **silent no-op**, because the space travelled over a tab-delimited transport and a tab is IFS whitespace, so a tab-indented spelling arrives with its tab eaten and reads as added coverage while driving nothing. Repaired by moving the transport to the unit separator first, then adding the axis.

## Verification at head `7ede665f`

- Suite 16/16 under bash 5, ~9.8s over a 180-element cross product.
- The space-only narrowing that was green before the round-2 repair now reds the oracle arm; the new transport carries a leading tab where the tab-delimited one ate it.
- Round 1's own reproduction now reds the equality arm instead of passing.
- Mutations still red the oracle arm: indent strip deleted, indent strip moved after the keyword loop, `local` added to the strip list, `export` dropped from it, digits dropped from the name shape, quote requirement dropped.
- `shell-lint.sh` clean, `whole-tree-invariants.sh` passes, no red CI rows.

## Open question for the next round

Both rounds landed in the enumerated input space, which is the residual this change named in advance and accepted. But round 2 showed the residual is livelier than the header's prose allows: an axis can be **present and still undriven**, which reads from outside exactly like the false claim the change set out to remove. Whether that wants a further repair or a sentence is unsettled.

PR #1745, opened 2026-09-03
debt(audit): give GAIA's executable prose a Code Audit Team owner (#1294)

### LINKED ISSUE #1294: The 41 unowned: exemptions have never been read as a population, and a wrong one fails silently green
`.gaia/audit-ci.yml`, the `unowned:` block (42 entries, header at `:256`, entries at lines 268-323).

> **Coordinates re-verified at `75f6143`.** Every line number and count below was re-derived against that commit. The dedup key above deliberately still reads `line=239`, its value when filed: the key is an identity matched on `path=`+`line=` across time, not a citation, so correcting it would break dedup and let a future audit re-file this as a duplicate. Do not "fix" it into agreement with the prose.

## Failure mode

The roster answers "who owns this path". The `unowned:` list answers "this path is meant to have no owner", and the coverage invariant in `.gaia/scripts/verify-audit-roster.sh` reports green **because** of it: an exemption converts an ownerless path from a reported gap into a declared intent.

`verify-audit-roster.sh` already checks each entry's *structural* health — it fails an entry that is dead (matches no tracked path), overbroad (reaches an owned path), redundant (no path needs it), or undecidable (outside the classifier dialect). **Nothing checks whether the judgment is right**: whether a path declared unownable should in fact have had an auditor. So a wrong entry fails silently green, and it fails in the one direction the roster cannot report, because the exemption is what suppresses the report.

**This is not hypothetical and the rate is the point.** PR #1282 (`#1245`) ran eight audit rounds, and five of them (2, 3, 6, 7, 8) each produced a fresh triage finding at a site no earlier round had examined. Round 7's is qualitatively different from the other four: an audit member with **no knowledge of the open question** independently graded `.claude/commands/**`, one of the exact paths already named as unexamined. A population that answers a prediction is not a sample, and drip-feeding it one finding per round has now been tried five times.

## The deliverable is the sweep

**A systematic re-read of all 42 `unowned:` entries, with a recorded disposition for each.** Fixing only the three seed paths below would repair three entries and leave the population unmeasured, which is the failure this issue exists to end.

Per entry, one of: keep (with the rationale stated where it is not already), re-own to a named member, narrow the glob, or delete it as spent.

**Seed, the three paths already named as unexamined:**

| Path | Line |
|---|---|
| `.claude/commands/**` | `.gaia/audit-ci.yml:284` |
| `.claude/instructions/**` | `.gaia/audit-ci.yml:285` |
| `.gaia/cli/health/**/*.md` | `.gaia/audit-ci.yml:291` |

All three are prose a language model executes rather than prose a human reads, which is why they surfaced.

## Dependency, named rather than blocking

`#1278`'s leg-2 prior question governs part of this triage: **is "a command a language model runs from prose" in scope for an auditor at all?** If yes, the three seed paths and several siblings are candidates to re-own; if no, they are correctly exempt and the sweep records that with a rationale instead. This does not block the sweep — most of the 42 entries do not turn on it — but the disposition for the seed three does, so sequence accordingly or record them as pending on `#1278`.

## Surface

Labelled `surface:maintainer` on the evidence: **every observed instance sits inside the `gaia:maintainer-only` marker span** (`.gaia/audit-ci.yml:278-324`), which the bundle-time scrub strips, so no adopter has those 32 entries.

**Named condition that reclassifies this to `surface:adopter`:** `.gaia/audit-ci.yml` is in `.gaia/manifest.json` and is not release-excluded, and **10 of the 42 entries sit outside the marker span and ship** (`*.md`, `LICENSE`, `.editorconfig`, `.gitignore`, `wiki/**`, `docs/**`, `public/**`, `.github/FUNDING.yml`, `.github/ISSUE_TEMPLATE/**`, `.github/pull_request_template.md`, at lines 268-277). Those ten are in the sweep's population and their correctness has never been checked either. If the sweep finds a wrong entry among them, an adopter inherits a wrong exemption and this becomes adopter-facing.

## Suggested fix

Work the 42 entries in file order, recording each disposition. Land the corrected list plus whatever member-glob changes the re-owned paths need, and state the sweep's result as a count (entries read, kept, changed) so the next reader can tell a completed sweep from a partial one.

Handler: plan

---

<sub>**Body revision history.** Filed against `23c7301`, where the block held 41 entries at lines 239-298 with a marker span of 253-298. Two commits have touched the file since — `b2946155` (#1365) and `eb2f0afe` (#1476) — shifting the block +29 lines and adding one entry, `.gaia/tests/**/*.tsx`, inside the marker span (so inside went 31 → 32 while outside stayed 10). The finding itself never changed; only its coordinates did. Body corrected in place rather than by comment so a reader never meets the stale figures first.</sub>

stevensacks:
Reads the roster's `unowned:` list as a population for the first time and records a disposition for every entry, which is what #1294 asked for. Fixing only the three seed paths would have repaired three entries and left the population unmeasured.

The sweep found one shape running through the list: **prose a language model executes** was sitting in an exemption that asserts it is meant to have no reviewer, while the structurally identical `.claude/skills/**/*.md` has been owned by the prose member since that member existed.

## Changed

Re-owned to `code-audit-maintainer-prose`:

| Was | Now | Files |
|---|---|---|
| `.claude/commands/**` | prose member | 13 |
| `.claude/instructions/**` | prose member | 2 |
| `.claude/agents/worthiness-evaluator.md` | prose member | 1 |
| `.gaia/cli/health/**/*.md` | prose member | 7 |
| `.github/forensics/**/*.md` | narrowed to `README.md` + `tests/**`; the two prompts → prose | 2 of 14 |
| `.specify/extensions/gaia/**/*.md` | narrowed to README + `test/**`; `commands/` `rules/` `templates/` → prose | 13 of 18 |
| `.specify/presets/**` | narrowed to `**/*.yml`; Markdown → prose | 3 of 4 |

Every one of these sits inside the `gaia:maintainer-only` marker span, so the release scrub strips them and **an adopter clone's roster and exemption list are unchanged**.

Narrowing rather than deleting is deliberate: a new directory under one of these roots is now reported ownerless instead of inheriting an exemption nobody decided to give it.

## Considered and rejected

`.gaia/cli/templates/**` (36 files) looked like a strong re-own to the node member, whose remit already names render templates, and it is not inert: `release/scrub.ts` reads `templates/workflows/` for workflow YAML that ships. It was rejected. `bundle:adopter` runs `rm -rf templates && cp -r src/scaffold/templates templates`, so the directory is a build artifact re-created verbatim from `src/`, which the node member already owns via `.gaia/cli/src/**/*.tmpl`. A finding raised on the copy cannot be repaired there. The routing-parity fixture caught this: the row for `.gaia/cli/templates/workflows/code-review-audit.yml.tmpl` was already deliberately ownerless, and the re-own would have silently reversed that decision.

## Kept, with the reasoning recorded

The ten shipping entries are unchanged: root `*.md`, `LICENSE`, `.editorconfig`, `.gitignore`, `wiki/**`, `docs/**`, `public/**`, `.github/FUNDING.yml`, `.github/ISSUE_TEMPLATE/**`, `.github/pull_request_template.md`. Two checks worth recording:

- `public/**` — the file's own counter-example note is still accurate. `public/mockServiceWorker.js` is present, so the entry stays unowned here and deliberately in scope for the merge gate's allowlist.
- Root `*.md` — **one unresolved tension, recorded rather than fixed.** `CLAUDE.md` is executable prose by the same test everything above was re-owned on. It is not re-owned because the entry ships and the prose member is maintainer-only: carving `CLAUDE.md` out of the shipping glob would orphan it on every adopter clone, and claiming it while the glob still reaches it fails as `OVERBROAD`. Left alone deliberately; worth its own issue if it should move.

The maintainer-only remainder stays exempt as fixture corpora, generated output, and prose nothing executes: the CLI bundles and `templates/`, `test-fixtures/**`, `.gaia/scripts/**` and `.gaia/tests/**` fixture data, `.gaia/templates/**`, `.serena/**`, and the spec-kit extension's `*.yml` and `*.tmpl`. Two are exempt only because no auditor exists for the language: `.claude/skills/release-notes/eval/**` and `.gaia/tests/observability/serena/usage_scan.py` are Python, and the roster has no Python member. The literal pin on the latter is kept so a second Python tool reports ownerless rather than inheriting the exemption.

The distinction the sweep settled on is now recorded in the block itself, so the next entry gets decided rather than guessed: **not "is this Markdown" but "does an agent read it as instructions or as input data".**

## Audit round 1

`code-audit-maintainer-shell` withheld its marker on one Important finding, plus two Suggestions. All three were mine and all three are fixed:

- **The member's own definition was left behind by its widened remit.** `.claude/agents/code-audit-maintainer-prose.md` still narrated a two-surface scope that this diff falsified. That is not cosmetic: `write-audit-remits.sh` regenerates the glob block but not the prose above it, so a dispatched prose member reading that paragraph could self-skip a slash command, write no marker, and strand the merge with nothing able to clear it. The opening paragraph now defers to the remit block as the whole of its scope and says explicitly not to narrow scope from the prose, and the surfaces that need a stated review posture get one. The frontmatter `description` carried the same falsified claim and is corrected in the same pass.
- **`.github/forensics/*.md` contradicted this PR's own decision rule.** It also captured `README.md`, a human-facing script table nothing executes, while the identical `.specify/extensions/gaia/README.md` was deliberately kept exempt twenty lines above. Same file in the same role, decided two ways in one roster. The roster now names the two executed prompts as literals and `README.md` joins the spec-kit one in `unowned:`; removing it again reds `ownerless-path`, so the exemption is load-bearing.

## Verification

- `verify-audit-roster.sh` green. Its `builtin-fallback-lockstep` invariant caught the fallback roster in `.claude/hooks/lib/audit-scope.sh` needing the same globs; both now agree.
- `write-audit-remits.sh` regenerated the prose member's remit region and is idempotent on re-run.
- **Adversarial fixtures, both run and both observed red before being restored:**
  - Dropping the `.claude/commands/**` grant reds with `ownerless-path` naming all 13 files, proving the grant is load-bearing rather than decorative.
  - Routing the same glob to the shell member instead reds the new routing-parity arm on 13 rows with `expected=code-audit-maintainer-prose`, proving the arm pins the member rather than merely accepting "not ownerless".
- `audit-scope-routing-parity.bats` gains one direction-guarded arm (`before = "-"`), matching the suite's existing style, plus its header entry. No count added, per that suite's own note.
- `whole-tree-invariants.sh`, `shell-lint.sh`, `lint-stale-cardinals.sh`, `lint-shipped-issue-refs.sh` all green; roster, remits, resolver and audit-scope bats suites green.
- Quality Gate skipped by its own rule: no staged `.ts|tsx|js|jsx|mjs|cjs|css` or gate-affecting config.

Closes #1294


stevensacks:
## Recorded dispositions (round 5)

Round 5 was a clean pass. Two advisory items, neither blocking, both dispositioned rather than silently dropped:

**Filed as #1751 (cross-remit).** `wiki/decisions/Code Audit Team.md:174` restates the roster's globs as a second copy that nothing checks, and it has drifted both ways: the prose row still lists only `.claude/skills/**/*.md` (this PR widened that member to twelve globs), and the shell row omits `.gaia/tests/vendor/**` and `.github/dependabot.yml` (pre-existing). `wiki/` is in no member's remit, so no member could have repaired it. Filed rather than fixed here because the real decision is whether that table should restate the globs at all, or be generated from the roster the way each member's remit region already is.

**Not filed, recorded here.** `audit-scope-routing-parity.bats:118` (header at `:44`) describes the moved set as `.claude/commands/**` / `.claude/instructions/**`, while the roster grants the `**/*.md`-qualified form, and the arm matches without an `.md` anchor. Left as-is deliberately: both directories are Markdown-only, the before/after fixture is generated once and never regenerated, so no row can reach the divergence — and if one ever did, the arm would red spuriously rather than pass wrongly. That is the safe direction, and a tracked issue for an inert imprecision that fails closed is backlog noise. Worth tightening opportunistically the next time that suite is edited.

Two other findings surfaced during this PR's rounds were pre-existing and are tracked separately: #1748 (stale `record_count` pin, red on main for every branch) and #1750 (a spelled-out cardinal the cardinal lint cannot read).

Issue #1816, opened 2026-09-06
The Release Workflow page's shell-lint guard enumeration names one guard the harness does not run and omits six it does


stevensacks:
`wiki/concepts/Release Workflow.md:151`, the `.github/workflows/shell-lint.yml` bullet.

## Failure mode

That bullet says the harness "also folds in custom pattern lints beyond shellcheck, each for a class shellcheck cannot model:" and then names them, one clause each, with a description of the class. It reads as the enumeration of what `Shell Lint` actually runs, and it is the only prose in the tree that describes those guards in one place, so a reader asking "what does this gate enforce" reads it and stops there.

It has drifted in both directions at once:

- Six guards `.gaia/tests/shell-lint.sh` folds in are absent from it: `lint-collapsed-signal-trap.sh`, `lint-stale-cardinals.sh`, `lint-guard-rule-shell-coverage.sh`, `lint-hook-wiki-inventory.sh`, `lint-wiki-cached-version.sh`, and `lint-hook-advisory-classification.sh`. A reader enumerating the gate from this page concludes the shell tree is unguarded against a collapsed `trap` arm, that a wiki page's cached version is unchecked, and so on.
- One guard the page names, `lint-shipped-issue-refs.sh`, is *not* folded into the harness at all: it runs as a `whole-tree-invariants.sh` member. So the page is wrong about where that one runs as well as about which ones exist.

The derivation both ways is one command against the harness the page describes:

```bash
grep -oE 'lint-[a-z-]+\.sh' .gaia/tests/shell-lint.sh | sort -u
grep -oE 'lint-[a-z-]+\.sh' 'wiki/concepts/Release Workflow.md' | sort -u
```

Nothing compares them, which is why the drift accumulated silently across at least four separate changes that each added a guard and left the page alone.

## Suggested fix

Apply the standing remedy this repository adopted in gaia-react/gaia#1809 for exactly this shape: **delete the restatement and point at whatever owns it**, rather than correcting the list. A corrected list is still a hand-kept list and starts decaying again from a fresher number, and this one has already demonstrated that.

[...]

Issue #1814, opened 2026-09-06
lint-sigpipe-readers scans no workflow YAML, the surface whose run: bodies are shell by another name


stevensacks:
`.gaia/scripts/lint-sigpipe-readers.sh:79`, the `workflow YAML` row of that gate's deliberate-exclusion list.

## Failure mode

`lint-sigpipe-readers.sh` (gaia-react/gaia#1810) flags a short-circuiting reader standing downstream of a pipe in a file that arms `pipefail`, the shape whose pipeline status inverts on a match. Its scan surface is the shared library's `shell` set: tracked `*.sh` and nothing else. Workflow and composite-action YAML are outside it, and their `run:` bodies are shell by another name, so an instance of the class written there is unreached by the gate that exists to catch it.

The exclusion is not arbitrary. A `run:` body inherits `pipefail` from the step's **resolved shell** rather than from any text in the body: GitHub's default `run:` shell is `bash -e`, an explicit `shell: bash` resolves to `bash --noprofile --norc -eo pipefail`, and a `defaults.run.shell` at workflow or job level moves both. The gate's armed test reads the body's own `set` lines, so it cannot answer the question for this surface at all, and running it there unchanged would grade every block by whether its author happened to write a redundant `set -o pipefail`.

That is what makes this an arming gap rather than a scope decision: the surface is governed by the same rule, and nothing checks it.

## Why it is worth tracking rather than waving through

The sibling gates that scan this surface already exist (`lint-collapsed-signal-trap.sh`, `lint-grep-ere-escapes.sh`, `lint-errexit-status-read.sh`, `lint-workflow-run-interpolation.sh` all ask the shared library for the `workflows` set), so the discovery half is free and only the pipefail oracle is missing.

The tree today is close to clean, which is exactly the condition under which an arming gap goes unnoticed:

- Most `run:` blocks carrying the shape arm no `pipefail` at all. Their only `set -o pipefail` sits **inside a command substitution** (`changed=$(set -o pipefail; git diff ... | tr ...)`), which does not reach the block around it, so those readers are not instances.
- One block is genuinely armed: `.github/workflows/audit-ci-tests.yml`, the `needs-cli-workspace` step, whose `run:` opens with `set -euo pipefail` and then pipes `bats-shards.sh files "$SHARD"` into `grep -qF`.

[...]

Issue #761, opened 2026-07-14
tech-debt: cra-specialist/cra-refuter still SIGPIPE-misclassify a large real result as NO-OP under pipefail


stevensacks:
## `cra-specialist` and `cra-refuter` still SIGPIPE-misclassify a large real result as NO-OP under `pipefail`

**Location:** `.gaia/scripts/audit-noop-detect.sh:243` (`cra-specialist` branch) and `.gaia/scripts/audit-noop-detect.sh:253` (`cra-refuter` branch).

### Failure mode

The script runs `set -uo pipefail` (line 74). Both branches classify a captured agent return with:

```bash
elif printf '%s' "$content" | grep -Eq '`[^`]+:[0-9]+`'; then   # cra-specialist, line 243
...
if printf '%s' "$content" | grep -Eq '\b(REFUTED|DOWNGRADE|STANDS)\b'; then   # cra-refuter, line 253
```

`grep -q` exits as soon as it finds its first match and closes the read end of the pipe. When `$content` exceeds the pipe buffer (~64KB) and the match lands early in the text, the upstream `printf` subshell receives SIGPIPE and exits 141. Under `pipefail` the whole pipeline's exit status collapses to 141, so the `if`/`elif` does **not** take the `real` branch even though a genuine match exists. Control falls through to `noop`, misclassifying a large, real specialist/refuter return as a no-op.

This is the identical hazard already removed from the success-present guard in PR #748 (closing #745), and just removed from this same file's new `audit-team-member` shape in PR #757 after `code-audit-maintainer-shell` reproduced it end-to-end with a >64KB fixture (see `audit-team-member`'s here-string fix and its `audit-noop-detect.bats` regression test). These two branches are the last remaining `printf | grep -q` pipes in the file and share the same class of bug.

**Blast radius:** the merge gate itself stays fail-closed either way (a misclassified no-op just triggers an unneeded hardened retry via the caller's no-op-detection-and-retry flow), so this is not a security or correctness break in the gate, but it wastes a retry dispatch and, on a second consecutive false no-op, surfaces a misleading "no-op'd twice" signal to the operator for a specialist/refuter that actually did real work and returned a large result.

### Suggested fix

Convert both pipes to here-strings, matching the #748/#757 remedy already applied elsewhere in this file:

```bash
elif grep -Eq '`[^`]+:[0-9]+`' <<<"$content"; then
...
[...]

Issue #745, opened 2026-07-14
tech-debt: pipefail + SIGPIPE in audit-success-present.sh can report 'no success live' on a commit that has one


stevensacks:
## `pipefail` + SIGPIPE on the guard's match pipeline can report "no success live" on a commit that has one

**Location:** `.github/audit/audit-success-present.sh:113`

```bash
if printf '%s\n' "$descriptions" | grep -qF -- "$tree"; then
```

Found by `code-audit-maintainer-shell` during the pre-merge audit of #740 (the fix for #739). Filed rather than fixed there: it is **latent and unreachable today**, and #740 had already converged.

### Failure mode

`grep -q` exits on its **first match**, which closes the pipe and SIGPIPEs `printf` (exit 141) whenever the input is large enough that `printf` has not already finished writing. The script runs under `set -euo pipefail`, so the pipeline's exit status becomes **141**, the `if` reads false, and execution falls through to the terminal `exit 1`.

Exit 1 is the guard's **"definitively NOT live"** answer. It is the one answer that tells all four callers to POST `pending` over whatever status is currently there. So this shape makes the guard report "no `GAIA-Audit` success is live" on a commit that **does** have one, and it does so *because the match succeeded*.

That is the #734 clobber, arriving at the **source** rather than at a caller: every one of the four call sites would be dutifully testing `-ne 1` against an answer that is already a lie, and no call-site lock in `ci-status-member-gate.bats` can see it.

Reproduced: a match on line 1 followed by ~1 MB of filler yields `rc=141` and takes the `exit 1` path.

### Why it is unreachable today (and only today)

The guard reads the **combined**-status endpoint (`/commits/<sha>/status`), which returns at most **one** status per context. So `$descriptions` is a single line of ≤140 chars (GitHub caps a status description there), far under the 64 KB pipe buffer: `printf` completes its write before `grep` can close the pipe, and SIGPIPE never fires.

**The reason this is worth recording rather than shrugging off:** the script's own `Notes` block names the `/statuses` endpoint (every status for the sha, chronologically, **unbounded**) as the alternative it deliberately rejected. So the precise edit that arms this bug is one a future reader has already been told about, in the same file. A "let me just read the full status history instead" change flips this from unreachable to live, silently, and the symptom (a clobbered merge gate on some PRs and not others) is exactly the hard-to-diagnose shape #734 was.

### Suggested fix

Remove the pipeline, and with it the `pipefail` exposure. Either:

```bash
[...]

PR #1817, opened 2026-09-06
debt(shell): gate the SIGPIPE-reader class and convert every site

### LINKED ISSUE #1810: A short-circuiting reader downstream in a pipeline under pipefail inverts the pipeline's truth value on a match, and nothing catches it
`.claude/hooks/wiki-session-stop.sh:48`

## Failure mode

A short-circuiting reader downstream in a pipeline, under `set -o pipefail`, inverts the pipeline's truth value on a match.

`grep -q` exits at its first match and closes the read end of the pipe. When the upstream writes more than the pipe buffer (~64KB) and the match lands early, the upstream takes SIGPIPE and exits 141. Under `pipefail` the pipeline's status collapses to 141, so an `if`/`&&` reading that status takes the **false** branch **because** a match was found.

The cited line is the one live instance in the tree where all three conditions plausibly hold together:

```bash
if git log "$start_sha..HEAD" --name-only --pretty=format: 2>/dev/null | grep -q '^wiki/'; then
```

The file runs `set -euo pipefail` (`:15`). The upstream is every changed path across a whole session's commit range, which can exceed the buffer, and `^wiki/` can match early when an early commit in the range touched a wiki page. The bad outcome is that the "update `wiki/hot.md`" reminder silently does not fire for exactly the sessions that changed the most wiki files. It fails quiet, so nothing surfaces it.

## Why this is worth a gate rather than a patch

This is the fourth known occurrence of the class, and the first three were each patched in place with no gate left behind:

- gaia-react/gaia#745 - `audit-success-present.sh`, fixed by gaia-react/gaia#748
- gaia-react/gaia#757 - `audit-noop-detect.sh`, the `audit-team-member` shape
- gaia-react/gaia#761 - `audit-noop-detect.sh`, the `cra-specialist` and `cra-refuter` shapes

Three of the four are **guard machinery**, where the failure mode is that a guard reports clean over a live defect. The fourth is the classification guard added in gaia-react/gaia#1809: its first draft used `grep -vE ... | grep -qE ...` and reported clean over the very defect it was written for, on its first run against the live tree. It was caught only because its red state was deliberately observed rather than assumed; no review round and no deterministic check saw it. Its header records the class at `.gaia/scripts/lint-hook-advisory-classification.sh`.

Nothing in the tree catches it:

- **shellcheck at the configured floor does not.** A fixture carrying both the `grep -v | grep -q` and `printf | grep -q` shapes under `set -uo pipefail` returns exit 0 at `--severity=style`, which is the floor `.gaia/tests/shell-lint.sh` sets for `*.sh`.
- **No lint scans for it.** Of the fourteen `.gaia/scripts/lint-*.sh` files, the only one whose text contains `SIGPIPE` is the classification guard's own header comment describing its own bug.
- The `code-audit-maintainer-shell` agent has caught it once (gaia-react/gaia#757), but it is model-dispatched and advisory, so nothing enforces it.

## Scope of the repair

Measured with `git grep` over tracked `*.sh` whose file sets `pipefail`, excluding comment lines: **69 sites across 29 files**. Of those, 62 are `printf`/`echo` of a variable and 2 read a file or the repository.

The 62 are not benign as a class: all three prior occurrences were exactly `printf '%s' "$var" | grep -q` where the variable held a large agent return. Whether a given variable can exceed the buffer is a data-flow property, so no static check can decide it.

That is fine, because the **remedy is unconditional and free**. `grep -q <<<"$var"` is a drop-in for `printf '%s' "$var" | grep -q` with identical semantics and no pipeline at all, and a single `awk` pass replaces the `grep -v | grep -q` shape. So a lint can flag every site with no discrimination and every fix is mechanical.

That is the same tradeoff `.gaia/scripts/lint-hook-array-guard.sh` already makes for a class it likewise cannot fully discriminate: it flags the shape, and its header directs a reader whose flagged site is genuinely safe to apply the fix anyway, "so the gate stays zero-exception rather than carrying an inline suppression."

## Suggested fix

Add `.gaia/scripts/lint-sigpipe-readers.sh`, folded into `.gaia/tests/shell-lint.sh` beside the other guards shellcheck cannot model, with a sibling bats suite under `.gaia/scripts/tests/`:

- Flag a short-circuiting reader (`grep -q`, `rg -q`, and the `-q` bearing flag clusters) appearing downstream of a `|` in a file that sets `pipefail`, in a non-comment line.
- Fail closed at discovery: exit non-zero rather than clean when the tracked-file listing comes back empty.
- No pragma. The remedy is always available, which is what makes a zero-exception gate the right shape here.
- Convert the 69 existing sites in the same change, so the gate lands green.

Two shapes deliberately outside it: `*.bats`, because bats-core does not enable `pipefail`, and a pipeline whose status is discarded rather than read as a truth value.

### LINKED ISSUE #1820: The sigpipe gate's closure tests never discriminate the basename index, so a path-keyed regression stays green
`.gaia/scripts/tests/lint-sigpipe-readers.bats:321`, and the sibling closure tests beside it.

## Failure mode

`lint-sigpipe-readers.sh` decides which files run under `pipefail` with a source closure: the seeds are the self-arming files, and an edge resolves by **basename** against a `#file <basename> <path>` index the scan loop emits. The basename index is the mechanism, because a load names its target through a variable far more often than not (`. "$lib_dir/guard-awk-lib.sh"`) and the literal token on the line is the basename rather than the path.

Every closure fixture in the suite writes its library at the fixture repo **root**, where `${f##*/}` is byte-identical to the path. The index key is therefore never discriminated: keying by path and keying by basename produce the same answer for every fixture the suite drives.

The real tree keeps its libraries in subdirectories, `.claude/hooks/lib/` above all, which the gate's own header names as the family three of the four historical occurrences of the class lived in. So the untested half is exactly the half that covers them.

A regression that re-keyed the index by path would silently drop every nested library out of the closure, and this suite, which declares itself the blocking runner, would stay green over it.

## Verified

The `code-audit-maintainer-shell` member reproduced it during the round-3 audit of gaia-react/gaia#1817: it mutated the index-key line to emit the full path instead of `${f##*/}` and ran the whole suite against the mutant, and all four closure tests still passed. Against a purpose-built fixture with the library at `hooks/lib/deep.sh` sourced from an armed root caller, the real gate reports `hooks/lib/deep.sh:2` and exits 1 while the mutant reports clean and exits 0. The suite cannot tell the two apart.

The gate itself is correct today: the same round verified the nested-library path works against the live tree. What is missing is the fixture that would red if it stopped working.

## Suggested fix

Relocate the library in one closure test into a subdirectory, so the basename and the path differ:

```bash
fixture_file lib/probe-lib.sh '#!/usr/bin/env bash
probe() { printf "%s" "$1" | grep -q needle; }'
fixture_file caller.sh '#!/usr/bin/env bash
set -uo pipefail
. "$(dirname "$0")/lib/probe-lib.sh"'
```

then assert `lib/probe-lib.sh:2`. One relocation is what makes the index key load-bearing under test. Re-run the auditor's mutant afterwards and confirm it now reds.

### LINKED ISSUE #1821: lint-sigpipe-readers misses a reader inside a downstream brace group or subshell, and its exclusion list does not name the shape
`.gaia/scripts/lint-sigpipe-readers.sh:190`, the skip run in `segment_reader`.

## Failure mode

`segment_reader` takes the segment head as the command word, stepping past only the modifiers it names: the empty token, `!`, `&`, `command`, `env`, and a `VAR=value` assignment. A downstream **compound command** puts `{` or `(` in that position instead, so the segment grades as some other command and the reader inside it is never seen:

```bash
some_command | { grep -q needle; }
some_command | ( grep -q needle )
```

Both carry the exact class the gate exists to end, and both pass it.

The cost is larger than the miss, because the gate's header enumerates its exclusions carefully, `*.bats`, the husky hooks, workflow YAML, and whether the pipeline's status is read at all, and presents that list as the whole of what the gate does not reach. A reader takes a clean run as covering every spelling of a downstream quiet reader. This one is not on the list and not covered.

## Verified

The `code-audit-maintainer-shell` member reproduced it during the round-3 audit of gaia-react/gaia#1817, on a fixture repo carrying an armed script with all three spellings on consecutive lines: the gate reported only the bare form and exited 1, saying nothing about the brace group or the subshell.

Nothing is live today. A repository-wide sweep for a downstream brace or paren over tracked `*.sh` returns only jq program text and `|| { ...; exit 1; }` error arms, neither of which is a reader.

## Suggested fix

Either is defensible, and the choice is the point:

- **Extend the skip run** past a leading `{` or `(` token, so the reader inside a compound command grades as the command word. Cheap, and it makes the gate's boundary match its header. Add a fixture for each spelling.
- **Name the compound-command wrapper in the header exclusion list**, so the stated boundary matches what the gate reaches.

Prefer the first: the remedy for a flagged site is unchanged (remove the pipeline), so widening costs nothing at the sites and closes a spelling a future author could reasonably write.

### LINKED ISSUE #1822: lint-sigpipe-readers calls its class short-circuiting readers while its flag set stops at the -q family, missing grep -m1
`.gaia/scripts/lint-sigpipe-readers.sh:181`, `is_qflag`, against the header claim on line 7 of the same file.

## Failure mode

The gate opens by naming its subject a **short-circuiting reader**, and the CHANGELOG entry repeats that framing. `is_qflag` admits a narrower set than the name: a `-q`-bearing flag cluster, `--quiet`, and `--silent`.

`grep -m1` and `--max-count` short-circuit identically. The reader stops at the count, closes the read end of the pipe, the upstream takes SIGPIPE and exits 141, and `pipefail` promotes that to the pipeline's status. So

```bash
if some_command | grep -m1 RE >/dev/null; then
```

inverts on a match for exactly the reason the `-q` shape does, and the gate reports it clean.

The mismatch between the name and the flag set is what makes this worth recording rather than merely noting: an author who reads the header, writes a `-m1` reader into a file that runs under `pipefail`, and sees a green gate has been told by the gate that the shape is fine.

## Verified

The `code-audit-maintainer-shell` member read `is_qflag` against the header claim during the round-3 audit of gaia-react/gaia#1817, and swept the tree for the spelling. Nothing is live: the only piped `-m`/`--max-count` reader in tracked `*.sh` is `.gaia/scripts/check-updates.sh:251`, which arms no `pipefail` and is sourced by nothing that does, so it is outside the closure. The three `.bats` hits are outside the scan surface by design.

## Suggested fix

Either, and the choice is between widening the gate and narrowing its claim:

- **Add the count-limited forms to `is_qflag`**: a cluster carrying `m` followed by a count (`-m1`, `-nm1`), the separated form (`-m 1`), and `--max-count`. This keeps the header's framing honest and closes a live spelling of the class.
- **Narrow line 7** to say the gate flags the `-q` family specifically rather than short-circuiting readers as a class, and say why the count forms are out.

Prefer the first: the class really is short-circuiting readers, the remedy at a flagged site is identical, and `check-updates.sh:251` stays clean because it is outside the closure, so widening lands green.

stevensacks:
Closes #1810

## What

A short-circuiting reader (`grep -q`, `rg -q`, and every `-q`-bearing flag cluster) standing downstream of a `|` **inverts** the truth value of its pipeline under `set -o pipefail`. The quiet reader exits at its first match and closes the read end of the pipe; once the upstream has written past the pipe buffer (~64KB) it takes SIGPIPE and exits 141, `pipefail` promotes that to the pipeline's status, and the `if` or `&&` reading it takes the FALSE branch *because* a match was found.

The cited instance is `.claude/hooks/wiki-session-stop.sh:48`, which piped a whole session's changed-path listing into `grep -q '^wiki/'`. The "update `wiki/hot.md`" reminder therefore did not fire for exactly the sessions that changed the most wiki files, and it failed quiet.

This is the fourth known occurrence, three of them in guard machinery where the failure mode is a guard reporting clean over a live defect (#745/#748, #757, #761, plus the first draft of `lint-hook-advisory-classification.sh`, which reported clean over the very defect it was written for). Each of the first three was patched in place with no gate left behind.

## The gate

`.gaia/scripts/lint-sigpipe-readers.sh` flags the shape in any tracked `*.sh` that arms `pipefail`, on a non-comment line, downstream of a `|` (same line, or carried across a continuation).

- **No discrimination, no pragma.** Whether a given upstream can outrun the buffer is a data-flow property no static check can decide. That is affordable because the remedy is unconditional and free: each repair removes the pipeline outright, so the class cannot recur on a repaired line. Same trade `lint-hook-array-guard.sh` already makes.
- **Buffered until END.** `pipefail` can be armed on a line *below* a pipeline, so candidates are held and released only if the file arms it anywhere.
- **Fail-closed at discovery.** The surface comes from `gaia_guard_scan_files ... shell`; an empty union exits 1, a discovery that never ran exits 3, a missing library exits 2.
- Folded into `.gaia/tests/shell-lint.sh` (advisory) with `.gaia/scripts/tests/lint-sigpipe-readers.bats` as the blocking runner in the `Audit CI Tests` scripts shard.

## The conversion

All 69 sites across 29 files, which is the count the issue measured and which re-derived exactly:

- 61 mechanical: `printf '%s\n' "$v" | grep -q RE` → `grep -q RE <<<"$v"`.
- 2 read the repository (`git log | grep -q`): captured into a variable first.
- 4 two-stage `grep -v` feeding a quiet grep, and 2 inverted readers (`grep -qv`, `grep -qvF`): folded into single `awk` passes. Here-strings alone would be wrong for these — `<<<` appends a newline that an inverted match counts as a non-matching line and succeeds on unconditionally, so those carry an explicit `length($0) > 0` guard.

## Guard-must-fail

The suite drives the gate against fixture trees one shape at a time, including the exact historical `wiki-session-stop` line, every flag-cluster spelling, both continuation forms, the pipefail-armed-below case, and the repair the gate itself advertises (a gate that reds on its own advice cannot ship). Red was observed on the live tree before the conversion: 69 findings, matching the issue's measurement.

## Verification

- `bash .gaia/scripts/lint-sigpipe-readers.sh` → clean (69 → 0)
- `.gaia/scripts/tests/lint-sigpipe-readers.bats` → 30/30
- the five other guards whose sites were converted, plus their suites → 297/297
- hook suites for the four changed hooks → 74/74; scripts suites for the six changed scripts → 246/246; spec-session-lock → 34/34; spec-kit lib → 14/14; forensics harness → pass
- `bash .gaia/tests/whole-tree-invariants.sh` → all pass (shellcheck at the `*.sh` floor and the bash-3.2 parse both cover the new file)
- Quality Gate: skipped by its own rule, no staged file is source or gate-affecting config

## Out of scope, filed

- **#1814** — the gate does not scan workflow `run:` bodies. A step's pipefail state comes from its *resolved shell*, not from body text, so the armed test cannot answer there. Measured: nearly every workflow instance sits in a block arming no pipefail at all (its only `set -o pipefail` is inside a command substitution, which does not reach the block); the one armed instance already carries a comment arguing it unreachable on a measured size margin.
- **#1816** — `wiki/concepts/Release Workflow.md` enumerates these guards by hand, names one the harness does not run, and omits six it does. Deliberately not appended to: adding a seventh name while six are missing leaves the list just as wrong and harder to notice.

## Audit rounds

Three rounds of `code-audit-maintainer-shell`; rounds 1 and 2 refused, round 3 earned the marker.

**Round 1** (two Important, two Suggestions), all fixed in `8570b674`:

- **`pipefail` is a process option, not a file attribute.** The first cut asked whether each scanned file carried its own `set -o pipefail`, so every *sourced library* read as clean, and `.claude/hooks/lib/` is where three of the four historical occurrences lived. Armedness is now a transitive source closure: seeds are self-arming files, edges are `.`/`source` loads resolved by basename against the tracked set. Two limits are stated in the header rather than hidden — a shared basename draws an edge to both files (costs a correct edit, never a missed defect), and a load built entirely from variables draws no edge. The edge reader recognizes a load anywhere a command may start, because the bracketed load an errexit-arming file writes puts it after a `;` and an `&&`. That closure exposed four more sites in `audit-dispositions.sh`, sourced by both merge-gate hooks; converted.
- The modifier skip list now steps past a `VAR=value` prefix, so a locale-prefixed quiet reader grades on `grep`. That spelling was live in the same library, so the two Importants shared one instance.
- The armedness pattern accepts the split-flag spelling; the bats exclusion says "by default" rather than claiming the class cannot fire there.
- Cross-remit: the CHANGELOG's "every existing site is converted" was falsified by the library sites. Corrected.

**Round 2** (one Important, three Suggestions), all fixed in `4668d96e`:

- The armedness pattern admitted only *flag tokens* ahead of the `-o` carrying `pipefail`, so a long-form option preceding it broke the run and graded the whole file unarmed. The suggested remedy admitted a bare word anywhere in the run; this takes the narrower form, admitting an option name only directly after a flag token, because `set` stops parsing options at its first non-option word and a line setting positional parameters arms nothing. Fixtures pin both directions and each reds under the other pattern.
- A comment describing a `|| true` guard on greps, left from a draft the awk lines replaced. A `basename` fork and a reopened append per scanned file, both removed.

**Mutation-tested rather than assumed.** Each new mechanism was driven into its failing state: removing the assignment arm, disabling source-edge emission, narrowing the edge reader to line-start, capping the closure loop, capping it after the first merge (reds on exactly the transitivity test), narrowing the armedness pattern, and widening it to the auditor's suggested form (reds on exactly the positional-parameters test).

## Round 3, and why it did not merge

Round 3 returned a **clean verdict and an earned marker**. It carried three Suggestions with no live failure mode, and they were first filed as follow-up debt with the intent to accept-and-note and merge.

Then the merge turned out to be blocked for an unrelated reason: main advanced during the audit rounds and the branch went `CONFLICTING` on `CHANGELOG.md`. The catch-up merge is forced, and it touches `.gaia/release-exclude` and `.gaia/tests/`, both inside the shell member's remit, so it rotates the content digest and retires the earned marker whatever else the commit carries.

That changed the calculus rather than the verdict. With a re-dispatch owed regardless, deferring the three suggestions would have bought a second one for nothing, so they ride the merge commit instead:

- **#1820** — every closure fixture wrote its library at the fixture repo root, where the basename and the path are the same string, so the basename index the closure resolves edges through was never discriminated. The auditor demonstrated it with a path-keyed mutant that survived all four tests. One fixture now nests its library, the shape the tree's own libraries take, and that mutant reds on exactly that test.
- **#1821** — a reader inside a downstream brace group or subshell put the wrapper in the segment head, so the segment graded as some other command. The skip run steps past both, and the command word tolerates a paren fused to it, since a subshell may open with no space.
- **#1822** — the flag test stopped at the quiet family while the header called the class "short-circuiting readers". A count-limited reader closes the pipe at the Nth match exactly as a quiet one closes it at the first, so `-m`, `--max-count` and their spellings join it, and the header now says so.

Each is mutation-tested: removing the count arm, the brace and paren tokens, the fused-paren strip, and re-keying the index by path each turn the suite red, the last on the relocated fixture alone. The suite is now 45 tests; `whole-tree-invariants.sh` passes on the merged tree.

**No accepted residuals remain.** This session is at the three-round audit cap, so the fresh clearance the merge commit now owes is left to the session that resumes.

## The resuming session's three rounds

A second session re-entered at step 1 and earned the clearance the merge commit owed. `origin/main` had not moved again; the whole-tree invariants, the gate itself, and its 45-test suite were green before the first dispatch, and the full `gh pr checks` output carried no red in any round.

Every round returned a **clean verdict and an earned marker**, and no round found a defect in any logic this branch authored. What the three rounds found was one prose class, in shrinking amounts: the branch had widened the gate to count-limited readers, and four sites still named the quiet family as the whole trigger.

- **Round 1** (two Suggestions), plus one site the tree-wide sweep found that the member did not, all fixed in `ad4be599`: the `.gaia/release-exclude` withhold rationale, the `shell-lint.sh` fold-in banner, and the bats suite's header sentence. The banner is the one an operator reads on a red, so a count-limited hit was being announced under a class it does not belong to. Each site now points at the gate's own header instead of restating its flag list, per the standing rule that a stale enumeration is deleted rather than completed.
- **Round 2** (one Suggestion), fixed in `d3c0295d`: one surviving header-level site described the suite's coverage as the flag cluster alone, while test 6 exercises the count family beside it. It defers to the header too, so a third family adds no obligation there.
- **Round 3**: clean, carrying one accepted residual.

### Accepted residual

- `.gaia/scripts/tests/lint-sigpipe-readers.bats:20` — round 2's own repair substituted a longer clause and left the line at 87 columns in a header block otherwise wrapped at 80. Cosmetic, in a file no line-length gate covers, and the member confirmed nothing reds on it. Accepted rather than fixed: the repair rotates the digest and buys a fourth round this session cannot dispatch, which would hand the pull request to a third session over a line wrap. The next edit to that paragraph re-wraps it.

### Third session: absorbing the sharder fix from main

The shard that blocked this pull request is fixed on main. `shard scripts-1`
was cancelled at 13m16s here with no failing assertion, reddening the
`Audit CI Tests` aggregator on a cancelled dependency; a catch-up
`git merge origin/main` absorbs the fix, and the same job now passes at 8m6s
with the aggregator green.

That merge touches three files in the shell member's remit, so it rotated the
member's digest and owed a fresh clearance. All three rounds this session
returned a clean verdict and an earned marker, and no round found a defect in
any logic this branch authored. The findings were on the sharder change this
branch absorbed, and this branch is what surfaced them: it adds a suite to the
directory that change describes.

- **Round 1** (two Suggestions), fixed in `aaac31ce`: the cost-outlier
  comment described its group by counting it, and this branch's new suite
  falsified both numbers. Dropped rather than bumped, per the counts rule, so
  the argument now rests on the correlation figures and the outliers' share of
  the group's wall clock, neither of which a new suite falsifies. The same
  substitution lands on the decision page. Separately, `scripts_shard_of`
  spelled the group's shard ids literally where the sharder builds
  `SCRIPTS_IDS` as the one place a scripts shard is added; it asks for the
  exchange group now.
- **Round 2** (two Suggestions), fixed in `59198c77`: the rationale added by
  round 1 claimed a fourth scripts shard would strand S16 and A6, when A6's
  anchoring-on arm names its expected shard ids outright and is unaffected. It
  scopes the claim to the arms it really covers. The loop also walked an
  unquoted command substitution where its two siblings in the same file read
  their id list line by line; it matches them now.
- **Round 3**: clean, carrying one accepted residual.

Verification before the first dispatch, all re-run **after** the merge:
`bash .gaia/tests/whole-tree-invariants.sh` all pass,
`bash .gaia/scripts/lint-sigpipe-readers.sh` clean, and the 45-test
`.gaia/scripts/tests/lint-sigpipe-readers.bats` suite 45/45. One interaction
was worth confirming rather than assuming: the merged sharder adds a
`while read ... break` loop and a `grep -qF` pipe, both close to the shape this
branch's own linter hunts. Neither is in the class, and the member verified it
independently: process substitution is not a pipeline, and the linter excludes
`*.bats` because bats arms no `pipefail`.

### Accepted residual, third session

- `.gaia/tests/lib/bats-shards.bats:573` — round 2's own repair wrote that
  A6's anchoring-on arm "names its two expected ids outright". The pair is
  accurate today, but nothing recounts it, so a third id assertion added to
  that arm would leave the sentence claiming two with nothing going red. It is
  the same decay class the round before it flagged, one spelling further on,
  and `lint-stale-cardinals.sh` cannot see it because a possessive determiner
  is one of the shapes its header leaves to a reader. Accepted rather than
  fixed for the reason the residual above gives: dropping the word rotates the
  digest and buys a fourth round this session cannot dispatch, which would
  hand the pull request to another session over one word.

Closes #1820
Closes #1821
Closes #1822


stevensacks:
Parked on #1823, not on anything in this diff.

All three Code Audit Team rounds are spent and clean: `code-audit-maintainer-shell` holds an earned marker for the current digest (`3f94322a…`, tree `dee3cb8a`), and `GAIA-Audit` is green on that tree. Five of the six required checks pass.

The sixth, `Audit CI Tests`, fails because `shard scripts-1` exceeds its 13-minute job cap. It was cancelled twice at 13m00s and 13m16s with no failing assertion, having simply run out of wall clock; on the previous run of this same branch it finished at 10m28s against the same cap while its two siblings ran 3m24s and 3m22s. A re-run is not a remedy, and this diff cannot be the cause: it is three comment lines.

The cap is unbudgeable (shards 13 + aggregator 2 is the ceiling the self-heal poller window allows, with zero headroom), so the fix is the partition, filed as #1823. Nothing in it can land here either: the sharder and the workflow are both in the shell member's remit, so any repair rotates the digest and buys a fourth audit round, which `block-fourth-audit-round.sh` denies.

Sequencing agreed with the maintainer: land #1823 on `main` first, since it reds every code-touching pull request rather than only this one. This branch then merges `main`, its shard goes green, and the merge proceeds on the clearance it already holds. Do not push anything else here in the meantime, the marker stays valid only while the tree does.

PR #1833, opened 2026-09-07
debt(tests): resolve env-indirected run: paths before the membership test

### LINKED ISSUE #1784: literal_inputs() drops a run: path indirected through a shell variable, so case 5 grades the step as having no inputs
`.gaia/scripts/tests/workflow-filter-coverage.bats:323`

`literal_inputs()` keeps a `run:` token only when it contains a `/` **and** is a member of `git ls-files`. A path indirected through a shell variable is captured by the token regex as `SOME_DIR/script.sh`, which is untracked, so it is dropped from the step's input set and case 5 grades that step as having no such input.

## Failure mode

Add a `dorny/paths-filter` gate to a job that indirects its script paths through a variable, and case 5 passes that step as having no inputs beyond its own workflow file. The step is then unarmed against a filter that does not reach the scripts it runs, and the guard that exists to catch exactly that reports green.

`forensics-triage.yml` is full of that shape (`$FORENSICS_DIR/parse-issue-body.sh` and around a dozen siblings, all tracked under `.github/forensics/`) and escapes today only because that job runs no paths-filter step at all, so every one of its steps is filtered out before the input check happens.

## Blast radius today is zero, and that was measured

Case 5 grades a step only when the step's `if:` names a `dorny/paths-filter` output in the same job. Under that narrowing exactly one untracked slash-bearing token exists in any gated `run:` body: `.gaia/cli`, from `pnpm -C .gaia/cli typecheck`. That is a directory rather than a file, and the `code:` filter reaches it via `.gaia/cli/**` regardless.

## Reproduced

By `code-audit-github-workflows` during PR #1781, in a scratch copy: repointing a gated step's `run:` at a nonexistent `.gaia/scripts/no-such-unfiltered-thing.sh` left the suite 16/16 green, while repointing it at `.gaia/scripts/check-main-root-derivation.sh` (tracked, and not reached by `code:`) correctly reds case 5. So the drop is the membership test, not the reachability logic.

## The membership test is load-bearing; do not simply loosen it

Its own comment says `Under-reaching by construction`. Without `token in tracked`, case 5 would drown in git refs (`origin/main`), URL fragments, absolute system paths (`/usr/local/bin/shellcheck`) and directory prefixes.

## Suggested fix

Expand simple `$VAR/` prefixes from the step's own `env:` block before the membership test, so a variable-indirected path resolves to the tracked file it names. That keeps the membership test's real job intact.

## Not covered, and correctly so

A PR that adds both a script and the step that runs it is fine: the guard runs against that PR's HEAD, where the new script is already tracked. That is the path PR #1781 itself took.

stevensacks:
## What

`literal_inputs()` in `.gaia/scripts/tests/workflow-filter-coverage.bats` keeps a `run:` token only when it holds a `/` and is a member of `git ls-files`. A path written through a shell variable never reaches that test as a path: `$D/x.sh` tokenizes as the untracked `D/x.sh`, and `${D}/x.sh` splits at the brace into `D` and `/x.sh`. Both drop, and the gated step grades as reading nothing but its own workflow file.

A step whose scripts are indirected that way is then unarmed against a filter that does not reach them, and the guard that exists to catch exactly that reports green.

## How

Substitute `$NAME` / `${NAME}` for the value the workflow states outright, in a workflow-, job-, or step-level `env:` block, before the membership test runs. The membership test's real job is untouched, which the issue was explicit about: it exists to keep git refs, URL fragments, absolute system paths, and bare directory prefixes out of the input set, and it still does.

Only a name whose value is a **literal string** resolves. A `${{ }}` expression, a name assigned in the shell, and a glob expansion all stay invisible, which is this guard's deliberate under-reaching bias. A scope that redeclares a name drops the outer value with it rather than falling back to it, so the extractor never names a path the step provably never reads: that is the one over-reach direction the change could have introduced, and the second new test pins it.

The header's out-of-scope note (item 2, paths built at runtime) and the "names literally" definition are both narrowed to match what the extractor now does.

## Verification

- `.gaia/scripts/tests/workflow-filter-coverage.bats` 18/18 green under bash 5, including cases 3-7 over the live tree.
- Both new tests were driven into their failing state before being relied on: the expansion test reds against the pre-change extractor, and the shadowing test reds against a build with the inner-scope drop removed.
- `bash .gaia/tests/shell-lint.sh` clean; `bash .gaia/tests/whole-tree-invariants.sh` passes.

## Blast radius

No verdict on the live tree changes. No gated step indirects a path through `env:` today, which is what makes this arming rather than a repair. `forensics-triage.yml` is full of the shape (a job-level `FORENSICS_DIR` in front of a dozen-odd tracked scripts) and escapes only because that job runs no paths-filter step at all.

The suite is release-excluded and absent from `.gaia/manifest.json`, so nothing adopter-facing changes and no CHANGELOG entry is owed.

## Audit rounds

Three rounds, each clean with a marker written. Rounds 1 and 2 each returned one Suggestion on the header's out-of-scope item 2, both applied:

- **Round 1**: item 2 claimed a reference which is not a whole leading segment stays invisible, but the substitution has no positional anchor. Repaired in `63dd4458`, which also put both off-leading-segment shapes into the expansion fixture so the claim re-checks itself.
- **Round 2**: the round-1 rewrite ended item 2's list with an `and`, closing a set that omits an `env:` value holding another `$NAME`. Repaired in `c3b015f5` by stating the rule that generates the set rather than naming the fourth case.
- **Round 3**: no finding on this change. Its two Suggestions are on a file this PR does not touch, recorded below.

The disposition for round 3 was pre-committed in a comment on this PR before the round ran.

Both sit on `.gaia/tests/hooks/local-janitor-wiki-doc.bats`, which reached this branch through the catch-up merge of `origin/main` and is absent from this PR's own diff against its fork point. Each is latent at that fork point, reads the same whether or not this change lands, and sits on a gate-machinery path, so both are waived rather than filed. No pointer to either is written into shipped content.

- `.gaia/tests/hooks/local-janitor-wiki-doc.bats:147` — SC2314: an absence assertion written as a bare `!`-negated `grep`. It gates today only because it is the test's final command; a line appended after it would make `set -e` exempt the negation and the assertion would stop failing, silently. Dedup key: `holistic/unclassified` at `.gaia/tests/hooks/local-janitor-wiki-doc.bats:147`.
- `.gaia/tests/hooks/local-janitor-wiki-doc.bats:54` — SC2020: `tr` with two-character sets reads as a likely word-replacement mistake. The character-set semantics are the intent here and the adjacent comment says so, so there is no live failure mode; recorded because the oracle is authoritative on whether a hit is real. Dedup key: `holistic/unclassified` at `.gaia/tests/hooks/local-janitor-wiki-doc.bats:54`.

Closes #1784

### DIFF
--- a/.gaia/scripts/tests/workflow-filter-coverage.bats
+++ b/.gaia/scripts/tests/workflow-filter-coverage.bats
@@ -31,10 +31,11 @@
 #   the globs of the filter output that gates it,
 #
 # where "names literally" means a git-tracked file appearing as a whitespace-
-# delimited token in the step's `run:` body, plus the `action.yml` of a local
-# composite action the step `uses:`. Each gated step's own workflow file counts as
-# an input of itself, which is the self-coverage half: a gate must re-run on a
-# change to its own definition.
+# delimited token in the step's `run:` body, after any `$VAR` whose value is a
+# literal string in an `env:` block in scope has been substituted for that value,
+# plus the `action.yml` of a local composite action the step `uses:`. Each gated
+# step's own workflow file counts as an input of itself, which is the
+# self-coverage half: a gate must re-run on a change to its own definition.
 #
 # The following are deliberately out of scope, because each needs a mechanism
 # this floor does not have and a decision this suite should not make on its own
@@ -43,8 +44,12 @@
 #   1. Transitive inputs. A step invoking `run-all.sh` gets that script checked,
 #      not the files the scenarios inside it inspect. Reaching those needs either
 #      a declared-inputs convention on every gated step or a runtime witness.
-#   2. Paths built at runtime. A path assembled from a variable or a glob
-#      expansion is invisible to a token scan, and always will be.
+#   2. Paths built at runtime. A `$VAR/` prefix is resolved when the workflow
+#      states the value outright, in an `env:` block whose value is a literal
+#      string; that much is right there in the file. Everything else stays
+#      invisible to a token scan and always will be: a value that is a `${{ }}`
+#      expression, a name assigned in the shell, a glob expansion, a path
+#      assembled from a variable that is not a whole leading segment.
 #   3. Composite-action bodies. A local action's own `run:` steps are not
 #      descended into; only its `action.yml` is checked.
 #   4. A filter propagated across jobs. The gate scan reads
@@ -179,6 +184,12 @@ GATE = re.compile(r'steps\.([A-Za-z0-9_-]+)\.outputs\.([A-Za-z0-9_-]+)')
 # A shell token that could be a repo-relative path. Deliberately permissive: the
 # tracked-file membership test below is what decides, not this.
 TOKEN = re.compile(r'[A-Za-z0-9_./+-]+')
+# `$NAME` or `${NAME}` in a `run:` body. `$` is outside TOKEN's character class,
+# so an unexpanded reference never survives the scan intact: `$D/x.sh` tokenizes
+# as the untracked `D/x.sh` and `${D}/x.sh` splits at the brace into `D` and
+# `/x.sh`. Either way the path the step really reads is dropped and the step
+# grades as reading nothing but its own workflow file.
+ENV_REF = re.compile(r'\$(?:\{([A-Za-z_][A-Za-z0-9_]*)\}|([A-Za-z_][A-Za-z0-9_]*))')
 
 
 def die(msg):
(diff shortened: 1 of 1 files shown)


stevensacks:
## Pre-committed disposition for round 3

Written before the round runs, per `wiki/concepts/PR Merge Workflow.md` (`#### When rounds stop: pre-commit a disposition for every branch`). Rounds 1 and 2 each came back clean with one Suggestion on the header prose; round 2's finding was on round 1's own repair, which is the signal that page names.

- **Clean** → merge.
- **Only Suggestions on the header prose** → accept-and-note here and merge at the cap. No further round is bought to widen a comment list.
- **A reproduced defect in the extractor logic this change authored** → stop, report, and hand off. Do not merge.

### Round history

- **Round 1** (`6ebc6973` base): clean, marker written. One Suggestion: item 2 claimed a reference which is not a whole leading segment stays invisible, but `ENV_REF.sub` has no positional anchor. Repaired in `63dd4458`, which also put both off-leading-segment shapes into the fixture so the claim re-checks itself. The member independently proved each new fixture case load-bearing.
- **Round 2** (`8f866bc9` anchor): clean, marker written. One Suggestion: the round-1 rewrite ended item 2's list with an `and`, closing a set that omits an `env:` value holding another `$NAME` (the substitution runs once and does not recurse). Repaired in `c3b015f5` by stating the rule that generates the set rather than naming the fourth case, so a fifth case is covered by the rule instead of buying another round.

Issue #1656, opened 2026-08-30
Command substitution is refused under worktree isolation, so the rest of every audit member's prescribed gate handshake is unrunnable too


stevensacks:
`.claude/agents/code-audit-maintainer-shell.md:282`

## Failure mode

#1654 records that the prescribed heredoc sidecar write is refused under worktree isolation. Fixing that closes one command. The refusal is not specific to heredocs: **command substitution is refused as well**, and every remaining step of the prescribed gate handshake is built on it.

Observed in a worktree-isolated session, running the constructs directly:

| construct | result |
|---|---|
| `cat <<'FINDINGS' ... FINDINGS` | refused |
| `out="$(printf '%s' 'hello')"` | refused |
| `printf '%s' '[]' \| bash .gaia/scripts/audit-write-findings.sh ... --findings -` | runs |

The refusal text is `This session is isolated in the worktree <path>, but this command is too complex to verify that it stays inside the worktree.`

So beyond the sidecar write #1654 covers, each of these prescribed commands is unfollowable as written in every member spec that carries it:

- the marker write, `marker="$(bash .gaia/scripts/audit-write-clearance.sh ...)"` (`code-audit-maintainer-shell.md:282` and `:308`, with counterparts in the other four specs)
- the `AUDIT_ROOT` / `KEY_BASE` / `BASE_SHA` derivations under "Remit and self-skip", which are command substitutions and which every later command consumes
- the status push, `( cd "$AUDIT_ROOT" && .claude/hooks/post-audit-status.sh "$marker" )` (`:355`), a subshell over a variable the member can no longer capture

## Why the capture matters, not just the call

`$marker` is consumed at `:355`, and the specs state themselves that shell state does not persist between Bash calls. Under worktree isolation the member can neither capture the writer's printed path nor carry it to the next command, so the handshake cannot be repaired by dropping the assignment the way the sidecar write can (there, nothing downstream reads `$findings_sidecar`).

That is the design decision this issue does not settle: either the handshake is restructured to be capture-free end to end, or the isolation guard is widened to admit these constructs, and the surrounding code does not pick between them.

## Related, and probably the same root cause

[...]

PR #1917, opened 2026-09-09
debt(agents): derive AUDIT_ROOT physically so the fence runs in a worktree

### LINKED ISSUE #1871: The audit members' AUDIT_ROOT derivation fence cannot run under worktree isolation, so every worktree round improvises it
`.claude/agents/code-audit-frontend.md:44` (the same two-line fence at `code-audit-github-workflows.md:25`, `code-audit-maintainer-node.md:35`, `code-audit-maintainer-prose.md:37`, `code-audit-maintainer-shell.md:45`)

## Failure mode

Every Code Audit Team member definition opens with the same derivation fence:

```bash
AUDIT_ROOT="${AUDIT_ROOT:-$(git rev-parse --show-toplevel)}"
AUDIT_ROOT="$(git -C "$AUDIT_ROOT" rev-parse --show-toplevel)" || exit 1
```

and the paragraph under it states the guarantee that every later Bash call re-runs those two lines before using `$AUDIT_ROOT`, because an empty value sends `git -C ""` against whatever tree the session sits in "silently and regardless of shell".

**Under worktree isolation the second line cannot run at all.** The runtime's worktree confinement refuses a `git` invocation whose target it cannot statically prove stays inside the worktree, and a shell variable in the `-C` operand defeats that proof unconditionally. Observed refusal text, from a `code-audit-maintainer-shell` round dispatched into `.claude/worktrees/`:

> names git in a form too complex to verify that it stays inside the worktree

The member reported that both the compound form and a simplified variable-free variant were refused, and that it substituted a sequence of separate plain commands with the root spelled literally, then hand-checked `merge-base` before passing the literal on to the clearance writers. That substitution is improvised per round, is not what any definition prescribes, and nothing checks it.

So the guarantee is unachievable in the mode where the hazard it names is real: inside a linked worktree the ambient tree and the audited tree genuinely differ, which is exactly when resolving `AUDIT_ROOT` ambiently certifies a tree other than the one reviewed.

## Scope, and what limits it

- All **five** member definitions carry the identical construct, so no member is exempt.
- Two of them ship: `.claude/agents/code-audit-frontend.md` and `.claude/agents/code-audit-github-workflows.md` are `owned` in the manifest. An adopter running a worktree-isolated audit inherits the same refusal, which is why this is `surface:adopter` even though the round that surfaced it ran a release-excluded member.
- The refusal is **loud**, not silent, which is the one thing keeping this below Critical: the denial is visible to the member rather than resolving ambiently behind its back. Two backstops also sit downstream, the dispatch prompt's mandatory expected-tree self-check and the clearance writer's digest refusal, and no wrong clearance has been observed.

What is not limited is the cost and the drift: every worktree-isolated round pays an unscripted workaround in the path that decides which tree gets certified.

## Adjacent carve-out that does not cover this

`code-audit-maintainer-shell.md:229` already documents a *different* runtime confinement, `Write`/`Edit` into the `.gaia/local` symlink, and rules explicitly that "the refusal is the runtime's own worktree confinement rather than a GAIA guard, so there is nothing to widen and it is not a finding." That carve-out is about writes into a symlinked directory and says nothing about the `git -C` derivation fence, so a member meeting this refusal has no instruction covering it. Whatever the disposition, the two cases should end up documented together rather than one being precedent a reader stretches over the other.

## Suggested fix

The orchestrator already interpolates a literal absolute working root into the dispatch prompt, so the supplied-root case never needs the variable-operand `git` call at all. Sketch, to be decided rather than adopted verbatim:

1. When `AUDIT_ROOT` arrives already set and absolute, skip the `git -C "$AUDIT_ROOT" rev-parse` normalization instead of re-deriving through it, keeping the `git rev-parse --show-toplevel` fallback only for the undispatched case where no root was supplied.
2. Decide explicitly what is lost by skipping it. That call also canonicalizes a symlinked path, and the shell-cwd and worktree guards elsewhere lean on physical-path comparison, so dropping it is not free and the replacement has to state what it still guarantees.
3. Land the same change in all five definitions together, and extend the `:229` carve-out to name this refusal too.

## Provenance

Surfaced by `code-audit-maintainer-shell` during the PR #1868 audit (round 3), flagged by the member as an operational note rather than filed as a finding against a file that PR touched. The orchestrator hit the same class of refusal repeatedly in the same session on unrelated commands (a `bash` helper invocation inside a command substitution, `gh ... --jq` compounds, `sed -f`), so the scanner's reach is broader than the audit fence; this issue is scoped to the fence, which is the instance sitting in shipped instructions.

Related but distinct: #1799 records the merge workflow prescribing one spelling that misbehaves under worktree isolation. Same family, different construct and different file.

stevensacks:
Closes #1871

## What was wrong

Every Code Audit Team member definition opened with the same two-line `AUDIT_ROOT` derivation, byte-identical in all five:

```bash
AUDIT_ROOT="${AUDIT_ROOT:-$(git rev-parse --show-toplevel)}"
AUDIT_ROOT="$(git -C "$AUDIT_ROOT" rev-parse --show-toplevel)" || exit 1
```

Under worktree isolation the runtime refuses a `git` invocation whose target it cannot statically prove stays inside the worktree ("names git in a form too complex to verify that it stays inside the worktree"), and a shell variable in the `-C` operand defeats that proof unconditionally. So line 2 could not run in the one mode where the ambient tree and the audited tree genuinely differ, which is exactly when resolving `AUDIT_ROOT` ambiently certifies a tree other than the one reviewed. Every worktree-isolated round improvised a substitute no definition prescribes and nothing checks.

## What changed

Line 2 becomes the tree's own physical-resolution idiom:

```bash
AUDIT_ROOT="$(cd "$AUDIT_ROOT" 2>/dev/null && pwd -P)" || exit 1
```

This is the third arm the issue's settle comment selected, not the issue body's original sketch. The body proposed skipping the normalization on the supplied-root path; the settle amendment rejected that because `.gaia/tests/hooks/audit-root-resolution.bats` stage 8 pins the canonicalization as a deliberate, non-vacuously-controlled invariant, and preserving it more cheaply beats deleting it.

- **Not a `git` invocation**, so the git-specific refusal does not apply.
- **Keeps the canonicalization.** Stage 8 and its non-vacuity control stay green and unmodified. On macOS `mktemp -d` yields `/var/…`, a symlink to `/private/var/…`, so a supplied root that is not normalized fails stage 8's comparison; the new form normalizes it.
- **Keeps the fail-closed behaviour** via `|| exit 1`, and improves on it: `cd ""` fails on bash 5 where `git -C ""` used to succeed against whatever tree the session sat in.
- **Established in-tree idiom.** `.gaia/scripts/main-root-lib.sh`'s `_gaia_physical_dir` is `( cd "$1" 2>/dev/null && pwd -P )`; there is no `realpath` or `readlink -f` call site anywhere in this repository, because neither is guaranteed present on macOS.
- **What it gives up** is the subdirectory-to-checkout-root lift `--show-toplevel` also performed. That is already refused fail-closed downstream: `.gaia/scripts/audit-write-clearance.sh` exits 2 with `--root '<X>' is not a checkout root`, so a member that lost the lift cannot mint a marker.

All five definitions move together and stay byte-identical in the fence, plus one new byte-identical paragraph in each recording what the replacement still guarantees and what it hands downstream.

`code-audit-maintainer-shell.md`'s adjacent worktree-confinement carve-out (the `Write`/`Edit`-into-`.gaia/local` case, ruled not a finding) now names this second instance too: same mechanism, different trigger, opposite escape hatch, so neither reads as precedent a reader stretches over the other. That was an explicit requirement of the issue.

A `## [Unreleased]` / `### Fixed` CHANGELOG entry lands with it: `code-audit-frontend.md` and `code-audit-github-workflows.md` are `owned` in the manifest and ship, so an adopter running a worktree-isolated audit inherits the fix.

## Verification

- `.gaia/tests/hooks/audit-root-resolution.bats`: **26/26**, before and after. Stage 8 (supplied root resolves to WT, physically), stage 8's fallback test, and stage 8's non-vacuity source-mutation control are all untouched and green, which is the fastest check that this arm is the right one.
- Agent-definition suites all green: `audit-guard-structural`, `check-scope-digest-adoption`, `debt-origin-contract`, `region-marker-conformance`, `verify-audit-roster`, `write-audit-remits`, `audit-rules-changed-lib`, `block-selfheal-paths`, `pr-merge-audit-check`, and the seven `doc-*` prose suites (397 assertions, 0 failures).
- `bash .gaia/tests/whole-tree-invariants.sh`: all pass.
- The new fence form checked directly on **bash 3.2.57** and **bash 5.3.15**: `/var/folders/…` resolves to `/private/var/folders/…` on both, and a non-existent path takes the `|| exit 1` arm on both.
- Quality Gate: nothing to check. The staged set is `.claude/**` markdown plus `CHANGELOG.md`, no source or gate-affecting config, so the gate's own skip rule applies.

## Not claimed here

The worktree confinement's acceptance of `cd "$AUDIT_ROOT"` with a variable operand was **not** reproduced under a live worktree dispatch; the observed refusal is git-specific and this arm leaves `git` out of the operand entirely, which is an inference rather than a measurement. The settle's stop condition applies: if `cd "$AUDIT_ROOT"` turns out to be refused too, stop and say so rather than improvising a fourth form.

Separately, and deliberately out of scope: each definition still carries 13 to 19 later `git -C "$AUDIT_ROOT"` consumers, which the same confinement may refuse. #1871 scopes itself to the fence, and the broader handshake surface was already filed and declined as #1656 (closed not-planned), so nothing new is filed for it here.


stevensacks:
Closing unmerged. The settled fix was implemented and then measured under a real dispatched worktree agent: the replacement fence is refused with character-for-character the same wording as the shipped one, because the construct the confinement objects to is line 1's `$(git rev-parse --show-toplevel)` inside the `${AUDIT_ROOT:-...}` fallback, not line 2's `-C` operand. A variable in the `-C` operand runs fine, which falsifies the premise the whole arm rested on. This is the stop condition the issue's own settle comment wrote down ("if `cd "$AUDIT_ROOT"` is also refused, STOP and say so; do not improvise a fourth form"), so nothing here is merged and #1871 goes back to the backlog with the measurements attached. Full detail: https://github.com/gaia-react/gaia/issues/1871#issuecomment-5604073166. The branch stays on the remote if any of it is wanted.

Issue #1957, opened 2026-09-10
audit-ci-tests.yml's code: filter arms none of 6 wiki pages a bats suite actually reads


stevensacks:
`.github/workflows/audit-ci-tests.yml:114`

## Failure mode

The `shards` job's `code:` filter (`.github/workflows/audit-ci-tests.yml:114`) lists specific tracked `wiki/` paths as literal entries, one per page a bats suite is known to read. Several tracked `wiki/` pages that a bats suite genuinely reads are absent from that list. A pull request that changes only such a page resolves `steps.filter.outputs.code` to `false`, every bats step in the job skips, and the job reports the declared-required context green having run that suite's assertions zero times. This is the same silent-green class the `code:` filter's own inline comments are written against (see the `.claude/rules/**`, `.husky/**`, and other narrated entries in the same filter); this is just an instance the filter has not yet closed.

## Candidate set, recounted from the tree at HEAD (`39a6a06e`)

Recount recipe (repeat this rather than trusting the list below, which rots):

1. **ARMED set**: the `wiki/` literals inside the one `dorny/paths-filter` step's `code:` key in `.github/workflows/audit-ci-tests.yml` (currently 8: `wiki/.state.json`, `wiki/concepts/PR Merge Workflow.md`, `wiki/concepts/Audit Disposition and Debt Fix.md`, `wiki/concepts/GAIA Audit.md`, `wiki/concepts/Claude Hooks.md`, `wiki/concepts/Policy-Memory Loop.md`, `wiki/concepts/Code Review Audit Agent.md`, `wiki/concepts/Task Orchestration.md`). Separately, the same step's `sandbox:` key arms `wiki/concepts/OS Sandbox.md`, which covers that one page through a different filter rather than leaving it uncovered.
2. **MENTIONED set**: `git ls-files 'wiki/**'` (191 tracked pages today) matched, both by full repo-relative path and by bare basename, fixed-string, against every tracked `.bats` file plus the `helpers/`, `lib/`, and `fixtures/` directories that sit beside one.
3. **Candidates** = mentioned minus armed, with **`.gaia/tests/hooks/fixtures/audit-routing-before.tsv` set aside**: it is a whole-tree path inventory that names 206 `wiki/`-prefixed lines (well over a hundred distinct pages), and it is not a real read of any of them, it is a flat listing. Counting it in would swamp the real candidates with noise from one fixture.

## The count, both ways

[...]

PR #1958, opened 2026-09-10
ci(audit-ci-tests): narrow code:-gated arming for wiki-prose-only pull requests

stevensacks:
## Summary

Narrows which `Audit CI Tests` shard legs a wiki-prose-only pull request arms, through two composing levers. Both default to arming the full matrix on anything they cannot resolve.

- **Lever one, the change type.** The `code:` filter's `wiki/.state.json` entry is qualified by change type (`deleted`), so the content-only write every wiki sync makes no longer arms the `code:`-gated legs. It rests on `check-wiki-state-collision.sh` being content-blind, which a behavioral oracle pins.
- **Lever two, the per-leg gate.** `.gaia/tests/leg-arming.sh` derives from the tree at run time which exchange groups hold a suite naming a changed wiki page. Its answer is an *additional* conjunct on each per-leg `code:`-gated step. The `steps.filter.outputs.code == 'true'` conjunct stays, so every narrowed step remains inside `workflow-filter-coverage.bats`'s gated set.

The work lands phase by phase, and this PR updates as each phase commits.

Related, out of scope: #1957 (pre-existing `wiki/` pages a suite reads that the `code:` filter never arms).

## Test plan

- [x] Phase 1: W4/W8 adversarial fixtures derive their lines; `read_wf codefilter` unwrap and refusal cases (`audit-ci-shards.bats` green under bash 5, `shell-lint.sh` clean)
- [x] Phase 2: lever one plus W13/W14/W15
- [x] Phase 3: arming script
- [x] Phase 4: W16/W17/W18
- [x] Phase 5: workflow wiring plus W19
- [x] Phase 6: decision page, coverage-suite comment, CHANGELOG
- [x] Phase 7: every new guard driven red on the final tree
- [x] `workflow_dispatch` run, read per step (run 34466993910: every leg armed and ran)
- [x] Canary wiki-sync-shaped PR, read per step: #1959, run 34477803633. Narrowed legs skipped and `lib` ran; evidence in the comments.

## Audit rounds

- **Round 1** cleared both members with Suggestions only. The fixes, in `a1792754`:
  - The arming step's `CHANGED_FILES_JSON` is bounded (withheld above 1,000 changed files), so an oversized list reaches the script empty and arms every leg instead of exceeding the kernel's per-string env limit.
  - A W19 arm pins the bound.
  - Four comment and prose corrections.
- **Round 2** cleared both members with Suggestions only. The fixes, in `2c6c7a7d`:
  - W19 now pins the bound it names: an anchored whole value, `<=`, and an integer of at most four digits. It has adversarial cases for a flipped operator, a trailing fallback, and a five-digit bound.
  - The workflow comment now says the file count stands in for the byte limit, with its honest fail-closed limit.
  - A provenance note is dropped.

  Two findings did not reproduce: that the bound leaves the `push` lane unbounded, and that the script header overclaims. The `shards` job's own `if:` admits only `pull_request` and `workflow_dispatch`, so the arming step never runs on push.
- **Round 3** cleared both members with Suggestions only. Every one lands on wording or the W19 regex that round 2 wrote. Under the disposition fixed before the round, they are accepted and noted below rather than looped, and the W19 hardening goes to a tech-debt follow-up.

## Accepted residuals

- **SC2016 info notices in `.gaia/tests/lib/audit-ci-shards.bats`.** Each flagged line is a literal needle, a doctored workflow line, or probe text, where the single-quoted `$` is intended. They are info severity, below the `.bats` warning floor, so `shell-lint.sh` passes. A file-wide disable would hide a future real SC2016, so the notices stay visible.
- **The count bound is a proxy for bytes.** A pull request with 1,000 or fewer changed files could still exceed the per-string env limit. The action splits a rename into an added entry and a deleted entry, so at the bound the budget is about 65 bytes per entry, not the roughly 131 the comment states. The workflows member measured the 1,000 longest tracked paths, all renamed, at 111,143 bytes against the 131,072 limit. Crossing the limit fails the arming step and reds the leg (fail-closed, never a false green). Accepted as noted: a round-3 comment precision point on prose round 2 wrote.
- **W19's four-digit ceiling is a loose retune window, not a tight bound.** A bound of `9999` still passes W19 and would leave about 13 bytes per entry. Widening that far fails closed, never green. The docblock's "narrow enough to rule out an effectively unbounded value" overstates this. Accepted as noted: a round-3 comment precision point on prose round 2 wrote.
- **W19's anchors are per line.** `grep` anchors `^` and `$` per line, and the step-field reader returns a block scalar raw. So a multi-line `CHANGED_FILES_JSON` whose first line is the bounded expression still passes W19. It fails safe: GitHub renders such a value as a non-JSON string, which the script's malformed-list rule arms on every leg, and an oversized one fails the step. The fix, rejecting a newline or matching the whole string, goes to the tech-debt follow-up.
- **The five-digit adversarial transform keys on the current bound literal.** A retune the assertion tolerates would make that case's transform a no-op, and `assert_doctored` reds it loudly. The fix, deriving the widening from the live value, goes to the same follow-up.

## Canary: post-merge, by maintainer decision

The planned pre-merge canary cannot run. `audit-ci-tests.yml` triggers `pull_request` only for `branches: [main]`, so a canary based on this branch would trigger nothing. A canary based on `main` would carry this branch's whole non-wiki diff and arm every leg. Widening the trigger is out of bounds.

The merge proceeds on three sets of evidence:
- the offline red/green table;
- this PR's own live runs, where every leg armed and ran;
- `workflow_dispatch` run 34466993910, where every leg armed through the dispatch disjunct.

Right after merge, a wiki-sync-shaped canary PR against `main` is read per step, checking that:
- the narrowed legs skip;
- `lib` runs;
- sandbox is unchanged;
- the aggregator succeeds.

The evidence is posted here. If any leg skips work it should run, the change is reverted immediately.


stevensacks:
## Post-merge canary: narrowing verified, no revert

The canary is #1959, branched from `main` at `666c904b`. It makes three wiki-sync-shaped edits:
- a content-only `wiki/.state.json` bump (same keys, valid JSON);
- a frontmatter bump on `wiki/concepts/GAIA Audit.md`;
- one `wiki/log.md` line.

`Audit CI Tests` run [34477803633](https://github.com/gaia-react/gaia/actions/runs/34477803633) on `e039c8f1`, read per step through the Actions jobs API:

| Job | Arming step | Run a bats shard | Installs (bats, YAML parser/zsh) | CLI-workspace install | INV-7 meter | Job |
|---|---|---|---|---|---|---|
| `hooks-1`…`hooks-4` | success | skipped | skipped | skipped | skipped | success |
| `scripts-1`…`scripts-3` | success | skipped | skipped | skipped | skipped | success |
| `audit`, `misc` | success | skipped | skipped | skipped | skipped | success |
| `concurrency` | success | skipped | skipped | skipped | skipped | success |
| `lib` | success | **ran**. Attempt 1 failed on an unrelated teardown flake (see below); attempt 2 succeeded | success (pinned bats, YAML parser/zsh) | skipped (its own probe: `lib` holds no js-yaml suite) | skipped | attempt 2 success |
| `sandbox` | skipped (by design) | skipped | own `Install bats` success | skipped | skipped | success (its `Run .gaia/tests/sandbox bats suites` step: success) |

- Hop-1 jobs: `Hook Capabilities (live tree)` success, `Verb Arming Adoption` success.
- `Audit CI Tests` aggregator: attempt 1 failed only because of `lib`'s flake; attempt 2 succeeded.
- **The attempt-1 flake is unrelated to narrowing.** On `lib`, 795 of 796 tests passed. The one failure was `plan-resume-point.bats` test `009a`, whose own assertions passed; its teardown then failed at line 32: `rm -rf "$REPO"`, with `rm: cannot remove '.../.git/objects': Directory not empty`, a late writer into its fixture repo. Re-running the failed job on the same commit went green. Tracked as #1964.

**Reading.**
- The `wiki/.state.json` rewrite armed nothing (lever one).
- `GAIA Audit.md` narrowed to `lib` alone, which matches the measured table (lever two).
- No leg skipped work it should have run: `lib` holds every suite naming the changed page.
- Sandbox is unchanged.

The canary PR is closed unmerged and its branch deleted.

Follow-ups filed on milestone 2.0.0: #1960 (awk reader folded continuation), #1961 (W14 pagination premise), #1962 (plan template shared-checkout git ban), #1963 (W19 hardening), #1964 (the teardown flake above).

PR #1959, opened 2026-09-10
docs(wiki): canary probe for Audit CI Tests per-leg narrowing

stevensacks:
Post-merge canary for #1958. **Do not merge**; it is closed once read.

A wiki-sync-shaped change:
- a content-only `wiki/.state.json` rewrite, which lever one keeps from arming `code:`;
- a frontmatter bump on `wiki/concepts/GAIA Audit.md`, whose namers attribute to `lib` only;
- one `wiki/log.md` line.

Expected per step: `lib` runs its suites, while every other non-sandbox leg skips its shard run and installs. Sandbox, the hop-1 jobs and the aggregator are unchanged.

### DIFF
--- a/wiki/.state.json
+++ b/wiki/.state.json
@@ -1,7 +1,7 @@
 {
   "version": 1,
-  "last_evaluated_sha": "486c3142d9b33c0ff0d99136cb1b5f88b786e821",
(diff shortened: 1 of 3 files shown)


stevensacks:
Canary read per step; evidence posted on #1958. Closed unmerged by design.

